PR #71, opened 2019-11-08
Bugs for NCE Form, Reports & select lists + Liquibase changes

muiruri:


CalebSLane:
These changes are required to be merged into develop for "develop" branch to be in a functional state (otherwise it is missing necessary liquibase changes)

PR #188, opened 2020-12-02
Properly localize OE reports and updates on workplan reports

Pkom17:


CalebSLane (approved):
Don't see any issues with this. Good work Pascal!

rossumg (approved):

PR #219, opened 2021-02-25
Change labels on Patient Report #176387139

Pkom17:
configured URAP Number as this number only applies to Cote d'Ivoire.
hide URAP Number label and field on Patient Report for other implementations.
Ordinance Number changed to Order Number


CalebSLane (changes requested):
Is there something meant to be committed in this? Right now it has 0 files changed

PR #249, opened 2021-03-31
Work on various bug on sampleAdd form, reports and analyzers plugins

Pkom17:


CalebSLane (changes requested):
For the change in SampleOrganizationDAOImpl, it may be better to add 'lazy="false"' to organization in SampleOrganization.hbm.xml since SampleOrganization and Organization are such tightly coupled concepts and since that strategy is more commonly used than TypedQuery throughout the code-base.

PR #276, opened 2021-06-24
add innitial ci build pipeleine

mozzy11:
This is the inntial ci build pipeline ,this will build every Pr and can be triggered to run manually


CalebSLane (approved):
These changes should not have any impact on the operation of OpenELIS

PR #277, opened 2021-06-24
fix OE maven build

mozzy11:


mozzy11:
hi @CalebSLane ,
ive just reproduced the same scenario on [my repository](https://github.com/mozzy11/OpenELIS2-CI/actions) , 
a pull request with this change failed ,but after merge , then the build passed

PR #281, opened 2021-07-28
Publish OpenELIS docker image to Github Container Registry

### LINKED ISSUE #280: Publish built OpenELIS image on Github Container Registry
We should have a `Dockerfile` that can be built using the following type of workflow, given a specific context:
https://github.com/B-TECHBW/openhim-mediator-fhir-converter/blob/handlebars/.github/workflows/publish.yml 

This script provides published images that can be used in `docker-compose` files directly, like this: https://github.com/B-TECHBW/openhim-mediator-fhir-converter/pkgs/container/openhim-mediator-fhir-converter

The `Dockerfile` can be a multi-stage build that calls the various scripts that are usually executed during an install.

@mozzy11 @CalebSLane @caseyi thoughts?

pmanko:
Closes #280


Dockerfile
```diff
@@ -1,3 +1,74 @@
...
+# Build Stage
+#
+FROM ubuntu:focal as build
+
+ENV KEYSTORE_PW="pass:kspass"
+ENV TRUSTSTORE_PW="tspass"
+ENV DEFAULT_PW="oepass"
+ENV INSTALLER_CREATION_DIR="OEInstaller"
+ENV STAGING_DIR="OEInstaller_stagingDir"
+ENV OE_BRANCH="master"
```
pmanko, line 11: likely don't need this variable

Dockerfile
```diff
@@ -1,3 +1,74 @@
...
+ADD ./src /build/src
+ADD ./install /build/install
+ADD ./dev /build/dev
+
+WORKDIR /build
+
+##
+# Checkout Dependencies
+#
+RUN git clone https://github.com/I-TECH-UW/dataexport.git --branch develop
```
pmanko, line 53: We should probably have an env var for which branch / tag to use for each of these dependencies

CalebSLane: Could this be read from the .gitmodules file? or can we just copy the files in from the submodule and ensure that the project is cloned with the --recurse-submodules option? Trying to cut down on having 2 locations that could point to different branches/tag

Dockerfile
```diff
@@ -1,3 +1,43 @@
+##
+# Build Stage
+#
+FROM ubuntu:focal as build
+
+ENV DEFAULT_PW="adminADMIN!"
```
mozzy11, line 6: @pmanko , since this variable is required at  at build time of the image , 
i think  we better pass it as `ARG` , instead of 'ENV' ,so that it can easily be passed to the builder action at build time

Dockerfile
```diff
@@ -1,3 +1,43 @@
...
+##
+# Build DataExport
+#
+WORKDIR /build/dataexport
+RUN mvn clean install -DskipTests
+
+WORKDIR /build
+
+# OE Default Password
+RUN ${DEFAULT_PW} | ./install/createDefaultPassword.sh
```
mozzy11, line 34: @pmanko 
ive tried this locally but it doesnt seem to be able to pass the [password value ](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop/install/createDefaultPassword.sh#L21)

mozzy11: Given theres already a [default password](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop/src/main/resources/adminPassword.txt#L1) set , we could just leave out this script .
Otherwise ,on building this image ,it sets password as null

install/createDefaultPassword.sh
```diff
@@ -18,24 +34,40 @@ echo "Please enter a default password that can be used to login into the default
 echo "This password will be stored in the created project war as a hash. "
 echo "This is technically secure, but it is recommended to change the password on installation"
 
-htpasswd -cBC 12 ${resourcesDir}/adminPassword.txt admin 
+echo "Output location: ${resourcesDir}"
```
pmanko, line 37: @CalebSLane @mozzy11 the logic in this file can probably be simplified, but I wanted to add a way to easily run this as a non-interactive command while keeping the same exact behavior as before

Dockerfile
```diff
@@ -1,12 +1,59 @@
...
+
+##
+# Checkout Dependencies
+#
+RUN git submodule update --init --recursive
+
+ENV DEFAULT_PW="adminADMIN!"
+
+# OE Default Password
+RUN ./install/createDefaultPassword.sh -c -p ${DEFAULT_PW}
```
pmanko, line 30: @CalebSLane Is this the right place for the default password creation, during the build step? the password is saved into the resource directory and used in the build, correct?

pmanko: cc @mozzy11

pmanko (commented):
@CalebSLane @mozzy11 let me know what you think of the Dockerfile and createDefaultPassword.sh updates

Dockerfile
mozzy11: this should be the right place , before compiling the source code

what if we just run the non interactive command here directly without using the `createDefaultPassword.sh `
`htpasswd -b -c ./src/main/resources/adminPassword.txt admin ${DEFAULT_PW}`

install/createDefaultPassword.sh
pmanko: @CalebSLane If we wanted to use the built docker image in a native installation of OpenELIS, should we run this script externally and then mount the password file into the docker image? How would we be able to use this approach in general with pre-built docker images?

pmanko: @mozzy11 in a docker install, we might still want to provide user-input options. We would want the user to be able to supply the admin password in some way. Currently, I don't think we are able to use a pre-built image and supply a user password, since the password is set in the build stage. @CalebSLane is that correct?

Dockerfile
```diff
@@ -1,12 +1,59 @@
...
+ADD ./dev /build/dev
+
+WORKDIR /build
+
+##
+# Checkout Dependencies
+#
+RUN git submodule update --init --recursive
+
+ENV DEFAULT_PW="adminADMIN!"
```
mozzy11, line 27: @pmanko , i think one remaining minor issue ,
we would better pass this as ARG , so that it can easily be passed to the builder action.
ENVs are more suitable for  variables needed at runtime , rather than at Build time

install/createDefaultPassword.sh
mozzy11: @pmanko , 
may be for the case when a user wants to build the docker image form source code , we could run the script outside the docker build

mozzy11: No i dont think its possible to set the default password after the image is build.
but i dont see any big issue with setting a specific default password as this can simply be changed on Login.
OpenMRS has a fixed default password .

mozzy11: May be the other way i think we can set a default password after the docker build , 
we can just mount an sql script file into the db-container to directly insert the user and passowrd into the db container

pmanko: good points - maybe we can get @CalebSLane 's thoughts on it. Let's not block this PR in the meantime, since we can just reset the password on login.

mozzy11 (approved):

install/createDefaultPassword.sh
mozzy11: sure, makes sense

PR #292, opened 2021-11-17
Add support to fetch Facilities form Gofr

mozzy11:


CalebSLane (approved):
looks good. We'll need to find some way to bring the Gofr configuration either into the setup script or ExternalConnection configuration, bu those don't need to happen immediately

PR #306, opened 2022-02-01
Updates to documentation

caseyi:
Updated the OE 1 to 2 upgrade documentation to add a backup note.


CalebSLane (approved):
Good job, but why is one line the same deleted and added

PR #309, opened 2022-02-11
Fix Next/Previous button and partial fix for search on the Analyser Results Page

mozzy11:
PT story :https://www.pivotaltracker.com/story/show/181175738


mozzy11:
For the Search , this is a partial fix.
ie when you serach for a Lab number of another page ,the page is loaded but number is not highlighted.

CalebSLane (approved):

PR #313, opened 2022-02-16
Add Roles By Lab Unit

mozzy11:
STORY : https://www.pivotaltracker.com/story/show/180909980
This addresses Task 1 , to add User roles Based on Lab Units


mozzy11:
image

mozzy11:
image

CalebSLane (changes requested):
DB ID's should be generated by using the corresponding sequence, and not hardcoded in the code or in liquibase. Can these ids be grabbed by name from the db instead please?

CalebSLane (approved):

PR #314, opened 2022-02-22
test:Implement Reporting Range

gitcliff:
Implement "Reporting Range"

link:https://www.pivotaltracker.com/n/projects/18030/stories/179167495


gitcliff:
Screenshot from 2022-03-09 09-46-07
Screenshot from 2022-03-09 09-19-03
Screenshot from 2022-03-08 17-51-25
Screenshot from 2022-03-08 17-50-32

caseyi:
@CalebSLane Please review this PR

CalebSLane (changes requested):
can the added columns be implemented in liquibase (under src/main/resources/liquibase/2.6.x.x/...) instead of in the initDB scripts? Otherwise existing installations will miss out on these columns as those are only run on install the first time.

In ResultLimitBean.java, setValidRange has the same functionality of setReportingRange. Is this intentional? I see that TestModifyEntryController.java calls setValidRange twice in a row. I'm assuming that one is supposed to be setValidReportingRange?

CalebSLane (approved):

PR #320, opened 2022-03-18
Filter Tests  by Lab Unit roles

mozzy11:


CalebSLane (approved):
@mozzy11 awesome work! 

One suggestion I have is that the filter logic in BaseController seems misplaced, as much as it's convenient to place it there. I think it would be better in a Service somewhere. I'll merge it in anyways so Casey can do his testing on it, and will consider it further.

PR #321, opened 2022-03-22
Test Validation Screen

gitcliff:
Test Validation Screen


gitcliff:
Screenshot from 2022-03-22 12-41-57
Screenshot from 2022-03-22 12-41-48

CalebSLane (changes requested):
@gitcliff as it is now, this checkbox is checking everything on the page. I believe that it is only supposed to check the objects that belong to the same order.

CalebSLane:
image
 The second order (and the "select all" checkbox should not be selected)

CalebSLane (approved):

PR #323, opened 2022-03-25
update Global admin role to acces Test Catalogue and some ui update

mozzy11:


mozzy11:
image

mozzy11:
@CalebSLane ,
i simply copied the system modules that were assigned to the `Test Management`  Roles that we deactivated

mozzy11:
image

CalebSLane (approved):
@mozzy11 are the values in unifiedSystemUser.jsp line 190-198 (role_1, role,11,etc.) hardcoded ids that come from the db? It is best if we could avoid such things if so. 

Another non-essential edit is in formFooterButtons.jsp on line 178. Could the 'if' be reformatted so we don't have an if { blank } else { do something }

Otherwise looks good! I'll merge in so that Casey can still test

mozzy11:
Thanks @CalebSLane 
are the values in unifiedSystemUser.jsp line 190-198 (role_1, role,11,etc.) hardcoded ids that come from the db? It is best if we could avoid such things if so.

yeah , since some data and metadata is loaded via the sql dump file ,  we end up havind fixed ids for some objects.
.do you think its better to use the role name ir role_globalAdministrator , role_audit etc ??

PR #332, opened 2022-04-25
notification:Add Password Expiry Notification

gitcliff:


gitcliff:
Screenshot from 2022-04-25 15-47-52

src/main/webapp/pages/common/banner.jsp
```diff
@@ -92,18 +94,33 @@ function getCsrfToken() {
...
 			%>
+
+			<script type="text/javascript">
+				function showExpiaryNotificationMessage() {
+					var now = moment().format('DD-MMMM-YYYY, h:mm:ssA');
+				    var date_string = '<%=Encode.forJavaScript(saveDate)%>';
+                    var expiration = moment(date_string).format("YYYY-MM-DD");
+                    var current_date = moment().format("YYYY-MM-DD");
+                    var hours = moment(expiration).diff(current_date, 'hours');
+                       alert(now + ' Your password will expire on '  +  date_string  + 'You have ' + hours + ' hours left .Please set new password ');
```
mozzy11, line 111: @gitcliff 
can you instead get the message form the message files for translation ??

src/main/java/org/openelisglobal/common/services/NotificationService.java
```diff
@@ -0,0 +1,22 @@
...
+import org.openelisglobal.spring.util.SpringContext;
+import java.sql.Date;
+import org.openelisglobal.login.service.LoginUserService;
+
+@Service
+@Scope("prototype")
+@DependsOn({ "springContext" })
+public class NotificationService {
+
+    private  LoginUserService loginService = SpringContext.getBean(LoginUserService.class);
```
mozzy11, line 16: i think its better to just Autowire the LoginUserService here since this class is a bean .
its a better Option to let spring handle beans than mannually accessing them

src/main/webapp/pages/common/banner.jsp
```diff
@@ -92,18 +94,33 @@ function getCsrfToken() {
  		<div style="display: block">
 			<%
 				UserSessionData usd = null;
+				NotificationService ns = new NotificationService(); 
 				if (request.getSession().getAttribute(IActionConstants.USER_SESSION_DATA) != null) {
 					usd = (UserSessionData) request.getSession().getAttribute(IActionConstants.USER_SESSION_DATA);
+					String saveDate = ns.getLoginFromCombinedId(usd.getLoginName()).toString();
```
mozzy11, line 100: better to just pass this object via a session attribute or you you can add it to the existing  USER_SESSION_DATA attribute

src/main/webapp/pages/common/banner.jsp
```diff
@@ -92,18 +94,33 @@ function getCsrfToken() {
  		<div style="display: block">
 			<%
 				UserSessionData usd = null;
+				NotificationService ns = new NotificationService();
```
mozzy11, line 97: This works fine but not a good pattern .
the reason we use MVC pattern is for separation of concerns

src/main/java/org/openelisglobal/login/valueholder/UserSessionData.java
```diff
@@ -31,6 +33,7 @@ public class UserSessionData implements Serializable {
     private int systemUserId;
     private String loginName;
     private boolean isAdmin;
+    private  LoginUserService loginService  = SpringContext.getBean(LoginUserService.class);
```
mozzy11, line 36: @gitcliff , 
this is also Anti pattern. you dont want to mix the service classes into the dto objects
can you pass the expiry date string ,from where the other attributes are passed from ??

mozzy11:
Thanks @gitcliff .
Now what happens , the bell always dispays , but when theres a pending notification , a small round incon displays on the bell to show a user that theres a pending notification. Actually usually ,  theres even a counter to show the number of pending notifications

CalebSLane:
@mozzy11 is this ready to be merged? or are you waiting for changes from cliff?

mozzy11:
@not yet , let me look and see whats remaning

mozzy11:
Thanks @gitcliff .
We now have a Notification Framework in the new UI. feel free to reqork on this in the new UI

mozzy11:
created issue here https://github.com/I-TECH-UW/OpenELIS-Global-2/issues/1508

PR #333, opened 2022-04-26
update ci build

mozzy11:


mozzy11:
Updated the CI build to ensure PRs dont break the app.

PR #334, opened 2022-05-03
Update order entry page

mozzy11:
Refactor Order Entry page to dispay test per sample added
https://docs.google.com/document/d/1ixH5_MGXJI4TKj20g0ZQsB-393jbUy7N/edit

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,6 @@ jobs:
       uses: actions/checkout@v2
       with:
         repository: ${{github.repository}} 
-        ref: develop 
     - name: Build OpenELIS-Global2
       run: mvn clean install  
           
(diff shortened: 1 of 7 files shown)


mozzy11:
image

CalebSLane:
@caseyi do we still want the select option to self erase once a selection is made, or remember the option that is currently selected? The previous option was done to allow multiple selections to be made, but since we spawn a new select for every sample it is no longer required

CalebSLane:
additionally, when removing a sample, it clears out the content of the div, but there's nothing to delete that div wholesale. 
sampleSelects
is this the desired functionality?

CalebSLane (changes requested):
seeing 2 instances of form:hidden path="sampleXML". Is this needed?

mozzy11:
> seeing 2 instances of form:hidden path="sampleXML". Is this needed?
I moved the `form:hidden path="sampleXML`from the `sampleAdd.jsp file` to avoid making it repeated several times and put it directly in the` samplePatientEntry.jsp` . That means `sampleEdit.jsp` no longer inherits it ,so i have to add it there too

mozzy11:
@CalebSLane ,
Pushed commit for
- Select sample should remeber selected Oprion
- Remove sample should clear out the entire sample Block

image

PR #336, opened 2022-05-05
fix importing electronic orders

mozzy11:


mozzy11:
@CalebSLane ,
this is a followup commit to fix importing electroning orders succesfully

PR #339, opened 2022-05-13
change observation effective date to result.lastUpdated

mozzy11:


src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceImpl.java
```diff
@@ -1011,7 +1011,7 @@ private Observation transformResultToObservation(Result result) {
         observation.setSubject(this.createReferenceFor(ResourceType.Patient, patient.getFhirUuidAsString()));
 //        observation.setIssued(result.getOriginalLastupdated());
         observation.setIssued(result.getLastupdated());
-        observation.setEffective(new DateTimeType(result.getOriginalLastupdated()));
+        observation.setEffective(new DateTimeType(result.getLastupdated()));
```
mozzy11, line 1014: i had not tested the `result.getOriginalLastupdated() ` but it returns null ,so observation.EffectiveDate wasnt being set.
switched to `result.getLastupdated()`

PR #340, opened 2022-05-19
display:Thumbnails no longer show for logos or signature

gitcliff:
Screenshot from 2022-05-19 17-51-17


CalebSLane (changes requested):
The goal is not to replace those thumbnails with static images, but to make sure the update-able image stored in the db is displayed.;

table joins should be site_information.value = image.id.

I think this got broken due to a change in Spring pattern matching. Try dropping '.jpg' or '.png' from the src (ie ./dbImage/siteInformation/headerLeftImage)

mozzy11:
Fixed this here https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/345

PR #349, opened 2022-06-14
critical:Implement critical ranges 

gitcliff:
I have added the ability to detect and ;alert critical range results for validators and and result enterers when a value is a critical result. Will implement Result Entry Management for the admin in a separate PR
https://docs.google.com/document/d/1JlK3nPXwIg1U3CmRW0fTaVE9zj4A19Ew/edit#


gitcliff:
Screenshot from 2022-06-14 22-34-16
Screenshot from 2022-06-14 22-33-31
Screenshot from 2022-06-14 22-32-17
Screenshot from 2022-06-14 22-31-47
Screenshot from 2022-06-14 13-58-43

CalebSLane (commented):
@caseyi am I correct in thinking that 0-10 and 95+ were just example critical ranges? Do we expect a constant-ish number of critical ranges (0, 1, or 2) or do we expect there to be any number of critical ranges?

caseyi:
Yeah, that's correct. It could be any numeric value.

src/main/webapp/pages/accessionvalidationrange/accessionValidationRange.jsp
```diff
@@ -104,6 +104,13 @@ function /*void*/ enableDisableCheckboxes( matchedElement, groupingNumber ){
 	jQuery("selectAllAccept").checked = false;
 }
 
+function criticalResultCheck( index ){
+	var result = jQuery("#resultId_" + index).val();
+	if((result >= 0) && (result <= 10) || (result >= 95)){
```
CalebSLane, line 109: these numbers should not be hardcoded. They need to be pulled from the critical range.

src/main/webapp/pages/testConfiguration/testAdd.jsp
```diff
@@ -589,6 +589,29 @@
...
         });
     }
+
+    function criticalRangeCheck() {
+        var highCriticalRangeValue, lowCriticalRangeValue;
+        var lowCriticalRange = jQuery("#lowCriticalRange");
+        var highCriticalRange = jQuery("#highCriticalRange");
+        lowCriticalRange.removeClass("error");
+        lowCriticalRangeValue = lowCriticalRange.val();
+        if((lowCriticalRangeValue >= 0) && (lowCriticalRangeValue <= 10) || (lowCriticalRangeValue >= 95)){
```
CalebSLane, line 599: these numbers should not be hardcoded. The critical ranges should be defined in test add and test modify and be results should be checked against them on result entry/validation

src/main/resources/liquibase/2.6.x.x/critical_range.xml
```diff
@@ -0,0 +1,32 @@
...
+		<preConditions onFail="MARK_RAN">
+			<not>
+				<columnExists tableName="result_limits" columnName="high_critical_range" />
+			</not>
+		</preConditions>
+		<addColumn schemaName="clinlims" tableName="result_limits">
+			<column name="high_critical_range" type="double precision" defaultValueComputed="'Infinity'::double precision"/>
+		</addColumn>
+	</changeSet>
+
```
CalebSLane, line 31: @caseyi says that we should expect up to 2 critical ranges per test, so we will need up to 4 values to track this. highRangeHigh, highRangeLow, lowRangeHigh, lowRangeLow. If it is easier for you to make the backend handle a variable number of ranges, please feel free to do so, otherwise add the 2 other fields

CalebSLane: in the example given of 95+ and 0-10 it would be the following:
highRangeHigh=Infinity
highRangeLow=95
lowRangeHigh=10
lowRangeLow=0

CalebSLane (changes requested):

src/main/webapp/pages/testConfiguration/testAdd.jsp
gitcliff: @CalebSLane jsut for clarification on what you mean by `critical ranges should be defined in test add and test modify`, do you mean in the test add and test modify jsp?

src/main/resources/liquibase/2.6.x.x/critical_range.xml
gitcliff: @CalebSLane kindly review ,, thanks

src/main/java/org/openelisglobal/common/provider/query/TestResultLimitsProvider.java
```diff
@@ -110,8 +110,11 @@ public int compare(ResultLimit e1, ResultLimit e2) {
                 resultLimitObject.put("highValid", resultLimit.getHighValid());
                 resultLimitObject.put("highReportingRange", resultLimit.getHighReportingRange());
                 resultLimitObject.put("lowReportingRange", resultLimit.getLowReportingRange());
-                resultLimitObject.put("highCriticalRange", resultLimit.getHighCriticalRange());
-                resultLimitObject.put("lowCriticalRange", resultLimit.getLowCriticalRange());
+                resultLimitObject.put("highRangeHigh", resultLimit.getHighRangeHigh());
```
CalebSLane, line 113: I’d like variable names to continue to mention critical ranges, so the concept is strongly established in the code across all classes.

using case sensitive find and replace:

highRangeHigh -> highCriticalRangeHigh
highRangeLow -> highCriticalRangeLow
lowRangeHigh -> lowCriticalRangeHigh
lowRangeLow -> lowCriticalRangeLow

HighRangeHigh -> HighCriticalRangeHigh
HighRangeLow -> HighCriticalRangeLow
LowRangeHigh -> LowCriticalRangeHigh
LowRangeLow -> LowCriticalRangeLow

src/main/java/org/openelisglobal/common/services/ResultService.java
```diff
@@ -0,0 +1,42 @@
+package org.openelisglobal.common.services;
```
CalebSLane, line 1: I don’t see a reason for this class to exist and in general we should avoid static variables for services like this static, but managed by spring (some pre-existing classes have this as an artifact from the migration to Spring, but it's use should not be encouraged).

src/main/java/org/openelisglobal/resultlimit/service/ResultLimitService.java
```diff
@@ -45,4 +45,13 @@ public interface ResultLimitService extends BaseObjectService<ResultLimit, Strin
     List<ResultLimit> getResultLimits(Test test);
 
     ResultLimit getResultLimitForAnalysis(Analysis analysis);
+
+    List<? extends Number> getHighRangeHigh();
```
CalebSLane, line 49: getHighRangeHigh, getHighRangeLow, getLowRangeHigh, and getLowRangeLow I would expect these to take a resultLimit object to get the configured values for a single resultLimit, not the entire list of ranges for all result limits, but that is also unnecessary as we can get the value directly from the resultLimit then

src/main/java/org/openelisglobal/resultlimits/dao/ResultLimitDAO.java
```diff
@@ -38,4 +38,12 @@ public interface ResultLimitDAO extends BaseDAO<ResultLimit, String> {
     List<ResultLimit> getAllResultLimitsForTest(String testId) throws LIMSRuntimeException;
 
     ResultLimit getResultLimitById(String resultLimitId) throws LIMSRuntimeException;
+
```
CalebSLane, line 41: These would be deleted along with the methods in the ResultLimitService

src/main/java/org/openelisglobal/testconfiguration/controller/TestAddController.java
```diff
@@ -198,10 +198,10 @@ private List<TestSet> createTestSets(TestAddParams testAddParams) {
         if (numericResults) {
             lowValid = StringUtil.doubleWithInfinity(testAddParams.lowValid);
             highValid = StringUtil.doubleWithInfinity(testAddParams.highValid);
-            lowReportingRange = StringUtil.doubleWithInfinity(testAddParams.lowReportingRange);
-            highReportingRange = StringUtil.doubleWithInfinity(testAddParams.highReportingRange);
-            lowCriticalRange = StringUtil.doubleWithInfinity(testAddParams.lowCriticalRange);
-            highCriticalRange = StringUtil.doubleWithInfinity(testAddParams.highCriticalRange);
+            lowRangeLow = StringUtil.doubleWithInfinity(testAddParams.lowRangeLow);
```
CalebSLane, line 201: Is there a reason lowReportingRange and highReportingRange have been converted to lowRangeLow and highRangeHigh?

Personally I would omit passing the Doubles into createResultLimits and only pass in TestAddParams so all of the StringUtil.doubleWithInfinity(value) can be inside createResultLimits, with independant null safety checks on each pair (reportingRange pair, lowCriticalRange pair, etc) (NOTE! Confirm where these values are stored in the JSONObject some might be at the top level (ie obj.get("lowValid")) while others may be inside an array inside the object (ie obj.get("resultLimits")... limitArray.get(i).get("lowNormal")).

src/main/java/org/openelisglobal/testconfiguration/controller/TestAddController.java
```diff
@@ -284,11 +285,11 @@ private ArrayList<ResultLimit> createResultLimits(Double lowValid, Double highVa
             limit.setHighNormal(StringUtil.doubleWithInfinity(params.highLimit));
             limit.setLowValid(lowValid);
             limit.setHighValid(highValid);
-            if(lowReportingRange != null && highReportingRange != null && highCriticalRange != null && lowCriticalRange != null){
-            limit.setLowReportingRange(lowReportingRange);
-            limit.setHighReportingRange(highReportingRange);
-            limit.setLowCriticalRange(lowCriticalRange);
-            limit.setHighCriticalRange(highCriticalRange);
+            if(limit != null)    {
```
CalebSLane, line 288: I think this null check is unnecessary and should be converted to the 3 independent null checks mentioned above (3 separate ifs).

src/main/java/org/openelisglobal/testconfiguration/controller/TestModifyEntryController.java
```diff
@@ -219,6 +219,8 @@ public int compare(ResultLimit o1, ResultLimit o2) {
                     significantDigits, "-"));
             bean.setReportingRange(SpringContext.getBean(ResultLimitService.class).getDisplayReportingRange(limit,
                     significantDigits, "-"));        
+            bean.setReportingRange(SpringContext.getBean(ResultLimitService.class).getDisplayCriticalRange(limit,
```
CalebSLane, line 222: I assume we don’t want to overwrite the reportingRange with the critical range, but instead mean to mirror the finctionality here by adding a g/setHighCriticalRange and g/setLowCriticalRange to ResultLimitBean?

src/main/java/org/openelisglobal/testconfiguration/controller/TestModifyEntryController.java
```diff
@@ -542,8 +546,10 @@ private List<TestSet> createTestSets(TestAddParams testAddParams) {
             highValid = StringUtil.doubleWithInfinity(testAddParams.highValid);
             lowReportingRange = StringUtil.doubleWithInfinity(testAddParams.lowReportingRange);
             highReportingRange = StringUtil.doubleWithInfinity(testAddParams.highReportingRange);
-            lowCriticalRange = StringUtil.doubleWithInfinity(testAddParams.lowCriticalRange);
-            highCriticalRange = StringUtil.doubleWithInfinity(testAddParams.highCriticalRange);
+            lowRangeLow = StringUtil.doubleWithInfinity(testAddParams.lowRangeLow);
```
CalebSLane, line 549: Same note as above, omit passing the doubles into createResultLimits and move the conversion of StringUtil.doubleWithInfinity(value) into the method

src/main/webapp/pages/accessionvalidationrange/accessionValidationRange.jsp
```diff
@@ -35,6 +37,10 @@
 	String searchTerm = request.getParameter("searchTerm");
 	String url = request.getAttribute("javax.servlet.forward.servlet_path").toString();	
 	//boolean showTestSectionSelect = !ConfigurationProperties.getInstance().isPropertyValueEqual(Property.configurationName, "CI RetroCI");
+	List<? extends Number> highRangeHighResult = ResultService.getHighRangeHigh();
```
CalebSLane, line 40: it is generally discouraged to add more scriptlets <% %> into the JSPs as it is an old style of doing things. Note below about converting this to use JSTL and mirroring the existing validation logic

src/main/webapp/pages/accessionvalidationrange/accessionValidationRange.jsp
```diff
@@ -104,11 +114,36 @@ function /*void*/ enableDisableCheckboxes( matchedElement, groupingNumber ){
 	jQuery("selectAllAccept").checked = false;
 }
 
-function criticalResultCheck( index ){
-	var result = jQuery("#resultId_" + index).val();
-	if((result >= 0) && (result <= 10) || (result >= 95)){
-		alert( `<%=MessageUtil.getContextualMessage("result.critical")%>` );
-	} 
+function criticalResultCheckLowRangeHigh( index ){
```
CalebSLane, line 117: This checking if a result is in a critical range will check to see if the value for this test is a critical range for ANY test, not just the one it is configured for (unless I am missing something). This logic seems like it should go beside normal, and reporting range validation inside testResults.js validateResults(). 

Values can be collected and embedded in the page via TestResultItem.java in the same way that lowerAbnormalRange is via ResultLoadUtility.setResultLimitDependencies. Search resultListView.jsp for <c:set var="lowerBound"  to see how these values are included on the page.

CalebSLane (changes requested):
More notes, but getting closer. Thanks @gitcliff!

caseyi:
@gitcliff do you think you will have time to finish this in the next few weeks?

gitcliff:
> @gitcliff do you think you will have time to finish this in the next few weeks?

yes @caseyi i am going to get some time in the coming days to resume work on this . Thanks

gitcliff:
hello @caseyi ,,,, 
@mozzy11 requested to finish up work on this since he has been working on OE of lately

PR #380, opened 2022-10-07
Reject Sample

mozzy11:


mozzy11:
here is a sample csv report 
[SampleRejection.csv](https://github.com/I-TECH-UW/OpenELIS-Global-2/files/9754788/SampleRejection.csv)

PR #383, opened 2022-10-18
revert org.jsoup version upgrade

mozzy11:


mozzy11:
@CalebSLane , uograding the `org.jsoup`
see comment https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/373#discussion_r989994399
and also an [issue ](https://www.pivotaltracker.com/story/show/183516959) i fixed

mozzy11:
we would need to do some refactor if we are to upgrade this

PR #429, opened 2023-03-15
critical2:implementing ciritical ranges

gitcliff:
https://docs.google.com/document/d/1JlK3nPXwIg1U3CmRW0fTaVE9zj4A19Ew/edit# 
implemented critical ranges for the results person first


mozzy11:
@gitcliff can you first resolve these conflicts ??

gitcliff:
@mozzy11 kindly review

mozzy11 (dismissed):

mozzy11:
Thanks @gitcliff 
LGTM

src/main/java/org/openelisglobal/common/provider/query/TestResultLimitsProvider.java
```diff
@@ -107,6 +107,14 @@ public int compare(ResultLimit e1, ResultLimit e2) {
                 resultLimitObject.put("highNormal", resultLimit.getHighNormal());
                 resultLimitObject.put("lowValid", resultLimit.getLowValid());
                 resultLimitObject.put("highValid", resultLimit.getHighValid());
+                resultLimitObject.put("highCriticalRangeHigh", resultLimit.getHighCriticalRangeHigh());
+                resultLimitObject.put("highCriticalRangeLow", resultLimit.getHighCriticalRangeLow());
+                resultLimitObject.put("lowCriticalRangeHigh", resultLimit.getLowCriticalRangeHigh());
+                resultLimitObject.put("lowCriticalRangeLow", resultLimit.getLowCriticalRangeLow());
+                resultLimitObject.put("highCriticalRangeHigh", resultLimit.getHighCriticalRangeHigh());
```
mozzy11, line 114: Looks like you added these ranges twice into the `resultLimitObject`

src/main/resources/liquibase/2.7.x.x/critical_range.xml
```diff
@@ -0,0 +1,54 @@
...
+<databaseChangeLog
+	xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+	  http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+	<property name="now" value="now()" dbms="postgresql" />
+
+	<changeSet author="cliff" id="1">
+		<validCheckSum>8:d30ab032096adb6b489edac763c55588</validCheckSum>
```
mozzy11, line 11: you also don't need to add the validCheckSum , unless your modifying an existing changeset that already ran in production DBs

src/main/java/org/openelisglobal/common/provider/query/TestResultLimitsProvider.java
```diff
@@ -107,6 +107,10 @@ public int compare(ResultLimit e1, ResultLimit e2) {
                 resultLimitObject.put("highNormal", resultLimit.getHighNormal());
                 resultLimitObject.put("lowValid", resultLimit.getLowValid());
                 resultLimitObject.put("highValid", resultLimit.getHighValid());
+                resultLimitObject.put("highCriticalRangeHigh", resultLimit.getHighCriticalRangeHigh());
```
mozzy11, line 110: @gitcliff , i think we agrred with @caseyi to refactor this abit .
We don't really need to define new High and Low critical **Ranges** explicitly ie
`highCrticalRangeHigh-HighCrticalRangeLow` . and `lowCrticalRangeHigh-lowCrticalRangeLow`.
Technically we already define min and Max valid range ie `highValid-lowValid`

So in summary to define critical rages , you only need to define only High critical and low critical values. ie `highCritical `and `lowCritica`l

- That way the low critical range becomes `lowValid-lowCrtical `instead of  `lowCrticalRangeHigh-lowCrticalRangeLow`

- The High critical range become `highValid-highCritical` instead of `highCrticalRangeHigh-HighCrticalRangeLow`

mozzy11: That way will only need to validate that the high and low critical values are with in the valid range

mozzy11 (approved):

mozzy11:
@gitcliff , this Looks good.
can you just resolve this conflict and I merge ??

PR #437, opened 2023-03-28
Fix bugs encountered while testing with OE9 databases 

Pkom17:


dev/properties/common.properties
```diff
@@ -31,4 +31,7 @@ org.openelisglobal.requester.phone=230 5123 1234
 
 # org.openelisglobal.providerlist.fhirstore=http://localhost:8081/fhir/
 # org.openelisglobal.requester.identifier=Practitioner/f9badd80-ab76-11e2-9e96-0800200c9a66
-facilitylist.schedule.fixedRate=864000000
\ No newline at end of file
+facilitylist.schedule.fixedRate=864000000
+
+#context for liquibase
+spring.liquibase.contexts=ci_regional
```
mozzy11, line 37: @Pkom17 i gues you could comment out this property since its implementation specific

docker-compose.yml
```diff
@@ -32,59 +32,63 @@ services:
...
-            - CATALINA_OPTS= -Ddatasource.url=jdbc:postgresql://database:5432/clinlims -Ddatasource.username=clinlims
-        volumes:
-            -  cert-volume:/etc/openelis-global
-            - ./volume/plugins/:/var/lib/openelis-global/plugins
-            - ./volume/tomcat/oe_server.xml:/usr/local/tomcat/conf/server.xml
-        secrets:
-            - source: datasource.password
-            - source: common.properties
+#    oe.openelis.org:
+#        build:
```
mozzy11, line 36: you could discard these changes too in this file since they were specific to your local enviroment

src/main/java/org/openelisglobal/common/provider/query/SampleSearchPopulateProvider.java
```diff
@@ -200,7 +200,7 @@ private void createReturnXML(Sample sample, String patientId, StringBuilder xml)
     private Organization getOrganizationForSample(Sample sample) {
         SampleOrganization so = new SampleOrganization();
         so.setSample(sample);
-        sampleOrganizationService.getDataBySample(so);
+        sampleOrganizationService.getDataBySample(sample);
```
mozzy11, line 203: the original code seemed right. i dont see how the `so` object gets assigned the Organization object fetched from the DB.
unless you do `so = sampleOrganizationService.getDataBySample(sample);`

src/main/java/org/openelisglobal/dataexchange/service/order/ElectronicOrderServiceImpl.java
```diff
@@ -30,136 +34,206 @@
...
+	@Transactional(readOnly = true)
+	public List<ElectronicOrder> searchForStudyElectronicOrders(ElectronicOrderViewForm form) {
+
+		IGenericClient fhirClient = fhirUtil.getFhirClient(fhirConfig.getLocalFhirStorePath());
+		Bundle searchBundle = fhirClient.search().forResource(Task.class)// .where(Task.AUTHORED_ON.after().day(form.getStartDate())).and(Task.AUTHORED_ON.before().day(form.getEndDate()))
+				.returnBundle(Bundle.class).execute();
+		searchBundle.getEntry();
+		for (BundleEntryComponent bundleEntry : searchBundle.getEntry()) {
+			if (bundleEntry.hasResource() && ResourceType.Task.equals(bundleEntry.getResource().getResourceType())) {
+				System.out.println("=============");
```
mozzy11, line 179: you could use the Loger instead

src/main/java/org/openelisglobal/dataexchange/service/order/ElectronicOrderServiceImpl.java
```diff
@@ -30,136 +34,206 @@
...
+					System.out.println("Org:"+ o1.getName());
+					System.out.println("Org2:"+ o1.getIdentifier().get(0).getValue());
+				}
+				System.out.println("Display: "+task.getRequester().getDisplay());
+			}
+		}
+		return null;
+
+//        switch (form.getSearchType()) {
+//        case IDENTIFIER:
```
mozzy11, line 197: discard these commented code if its not needed

src/main/java/org/openelisglobal/dataexchange/service/order/ElectronicOrderServiceImpl.java
```diff
@@ -30,136 +34,206 @@
...
+				Task task = (Task) bundleEntry.getResource();
+				System.out.println(task.getAuthoredOn());
+				System.out.println(task.getPriority());
+
+				Bundle bundle =  fhirClient.search().forResource(Organization.class).where(Organization.IDENTIFIER.exactly().code(task.getRequester().getReference())).returnBundle(Bundle.class).execute();
+				
+				Organization o1= (Organization) bundle.getEntryFirstRep().getResource();
+				if(ObjectUtils.isNotEmpty(o1)) {
+					System.out.println("Org:"+ o1.getName());
+					System.out.println("Org2:"+ o1.getIdentifier().get(0).getValue());
```
mozzy11, line 189: you could get rid of the `System.out.print` lines

src/main/java/org/openelisglobal/common/provider/query/SampleSearchPopulateProvider.java
Pkom17: The original code cause `null pointer exeception` while retrieving organization for sample. Because `so` does not keep the `sampleId`  from the sample object in parameter

dev/properties/common.properties
Pkom17: Ok right

src/main/java/org/openelisglobal/dataexchange/service/order/ElectronicOrderServiceImpl.java
Pkom17: Yeah this line should be removed

Pkom17:
Close

PR #443, opened 2023-04-05
Fix bugs encountered while testing with OE9 databases - reviewed

Pkom17:


mozzy11:
Thanks @Pkom17

mozzy11:
Your liquibase changeset seems to have an error , that's why the tests are failing

mozzy11 (approved):

PR #447, opened 2023-05-02
Sample Order Payload initiation

abertnamanya:


CalebSLane (approved):
Some comments that need to be addressed, but merging in to move this ahead

PR #451, opened 2023-05-11
Improved VL study forms and reports and fixed some bugs

Pkom17:


src/main/java/org/openelisglobal/image/service/ImageServiceImpl.java
CalebSLane, line 1: If we are going to catch the generic exception, we should do a LogEvent.LogErrorStack to make sure we can trace the error if it happens

src/main/resources/static/scripts/entryByProjectUtils.js
CalebSLane, line 1: what is the value 1251 supposed to be? if it's an id is it going to be static across all instances? if not, can it be declared as a named variable to give it more context?

src/main/resources/static/scripts/neon/entryByProjectUtils.js
CalebSLane, line 1: what is the value 1251 supposed to be? if it's an id is it going to be static across all instances? if not, can it be declared as a named variable to give it more context?

src/main/java/org/openelisglobal/patient/controller/BasePatientEntryByProject.java
CalebSLane, line 1: what is distinct about genders_2 from genders?

CalebSLane (changes requested):

src/main/java/org/openelisglobal/image/service/ImageServiceImpl.java
Pkom17: Okay... LogEvent has been added

src/main/resources/static/scripts/entryByProjectUtils.js
Pkom17: 1251 is the dictionary ID for "No" response  in the dictionary table and it is supposed to be static accross instances. 
it is used here to check "No" response by default in the study forms

src/main/resources/static/scripts/neon/entryByProjectUtils.js
Pkom17: 1251 is the dictionary ID for "No" response  in the dictionary table and it is supposed to be static accross instances. 
it is used here to check "No" response by default in the study forms

src/main/java/org/openelisglobal/patient/controller/BasePatientEntryByProject.java
Pkom17: finally commented out the first one as it is not used elsewhere.

CalebSLane (approved):

PR #464, opened 2023-05-30
modifyOrder:Initial work on modifying the order

gitcliff:


mozzy11:
@gitcliff , can you fix the merge conflicts ??
we changed the format of the menu from a top nav to a sid nav

mozzy11:
just copy the content [here ](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/2.8/react-ui/src/components/layout/Header.js). and add your menu item

gitcliff:
@mozzy11 its done

react-ui/src/components/modifyOrder/ModifyOrderForm.js
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  handlePerPageChange = (newPerPage) => {
+    this.setState({ perPage: newPerPage })
+  }
+
+  render () {
+    const { page, pageSize } = this.state
+    return (
+
```
mozzy11, line 86: @gitcliff , Looks like you're implementing a Patient search component again here.
we have already written most of these common re-usable components [here](https://github.com/I-TECH-UW/OpenELIS-Global-2/tree/2.8/react-ui/src/components/common). 
you'll need to re-use those components for creating patients , creating orders etc

mozzy11 (changes requested):
and also @abertnamanya already created components to do with Creating orders [here](https://github.com/I-TECH-UW/OpenELIS-Global-2/tree/2.8/react-ui/src/components/addOrder). .
you would mostly reuse most of those

gitcliff:
cc @mozzy11 
you can review

gitcliff:
cc @mozzy11 
you can review

PR #656, opened 2023-12-30
Add support for URL search params in Validation components

rahul6603:
Issue - https://github.com/I-TECH-UW/OpenELIS-Global-2/issues/613

### DIFF
--- a/frontend/src/components/validation/SearchForm.js
+++ b/frontend/src/components/validation/SearchForm.js
@@ -107,7 +107,8 @@ const SearchForm = (props) => {
       ? values.accessionNumber.split("-")[0]
       : "";
     var unitType = values.unitType ? values.unitType : "";
-    var date = testDate ? testDate : values.defaultDate;
+    var defaultDate = values.defaultDate ? values.defaultDate : "";
+    var date = testDate ? testDate : defaultDate;
     let searchEndPoint =
       "/rest/accessionValidation?" +
       "accessionNumber=" +
--- a/frontend/src/components/validation/SearchForm.js
+++ b/frontend/src/components/validation/SearchForm.js
@@ -161,7 +161,7 @@ const SearchForm = (props) => {
     switch(searchBy){
       case "routine":
         let testSectionId = new URLSearchParams(window.location.search).get(
-          "test_section_id"
+          "testSectionId"
         );
         testSectionId = testSectionId ? testSectionId : "";
         getFromOpenElisServer("/rest/user-test-sections", (fetchedTestSections) => {
--- a/frontend/src/components/validation/SearchForm.js
+++ b/frontend/src/components/validation/SearchForm.js
@@ -35,6 +35,11 @@ const SearchForm = (props) => {
   const [searchBy, setSearchBy] = useState();
   const [doRange, setDoRagnge] = useState(true);
   const [testSections, setTestSections] = useState([]);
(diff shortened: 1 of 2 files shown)


rahul6603:
https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/128216878/f6712717-47cf-4fa0-a840-8b7faaafac7a

frontend/src/components/validation/SearchForm.js
```diff
@@ -152,16 +158,60 @@ const SearchForm = (props) => {
     if (param === "order") {
       setDoRagnge(false);
     }
-    getFromOpenElisServer("/rest/user-test-sections", fetchTestSections);
+    switch(searchBy){
+      case "routine":
+        let testSectionId = new URLSearchParams(window.location.search).get(
+          "test_section_id"
```
mozzy11, line 164: rename the  `testSectionId` in order to be consistent with the old ui

mozzy11:
Thanks @rahul6603 . added a small comment above

frontend/src/components/validation/SearchForm.js
rahul6603: Thanks, updated to `testSectionId`

mozzy11:
Thanks @rahul6603 , 
While this works fine , 
In principle , when lexical declarations are used inside a case block  , they should be wrapped inside use a block statement ({}) or curly braces  to create a scope for the variable .

ie instead of 
```
switch (someValue) {
  case 1:
    let x = 10; // Error: Unexpected lexical declaration in case block
    break;
  // other cases
}
```

it should be 

```
switch (someValue) {
  case 1: {
    let x = 10; // OK
    // code specific to case 1
    break;
  }
  // other cases
}
```

frontend/src/components/validation/SearchForm.js
```diff
@@ -152,16 +158,60 @@ const SearchForm = (props) => {
     if (param === "order") {
       setDoRagnge(false);
     }
-    getFromOpenElisServer("/rest/user-test-sections", fetchTestSections);
+    switch(searchBy){
+      case "routine":
+        let testSectionId = new URLSearchParams(window.location.search).get(
```
mozzy11, line 163: Since youve used Lexical Declarations inside your case block ,
can you just wrap your case blocks inside curly braces /block statements ie 

```
case "routine":{
        let testSectionId = new URLSearchParams(window.location.search).get(
          "testSectionId"
        );
        testSectionId = testSectionId ? testSectionId : "";
        getFromOpenElisServer("/rest/user-test-sections", (fetchedTestSections) => {
          let testSection = fetchedTestSections.find(testSection => testSection.id === testSectionId);
          let testSectionLabel = testSection ? testSection.value : "";
          setDefaultTestSectionId(testSectionId);
          setDefaultTestSectionLabel(testSectionLabel);
          fetchTestSections(fetchedTestSections);
        })
        if(testSectionId){
          let values = { unitType: testSectionId };
          handleSubmit(values);
        }
        break;
}
      case "order":
```

rahul6603:
That makes sense, thanks @mozzy11 !

mozzy11:
I made a follow up commit [here](https://github.com/I-TECH-UW/OpenELIS-Global-2/commit/a716c44766b7bb4cdc89afe311b2e229cab55792)

PR #660, opened 2024-01-03
Add Expandanble tile (WIP)

tendomart:
This PR is  replaces carbon Clickable tiles with Expandable tiles as shown in the diagrams below
cc @mozzy11 @caseyi  @pmanko 

issue : [Complete the Home DashBoard#594](https://github.com/I-TECH-UW/OpenELIS-Global-2/issues/594)

Screenshot from 2024-01-03 14-45-47

Screenshot from 2024-01-03 14-45-35


mozzy11:
Thanks @tendomart , 
I did a PR to do more work on this .https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/666
The Expandable tiles wouldn't achieve what we want or they would need a lot of tweaking.

https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/29783151/783ec413-1810-4fb9-9214-faef94fc3e5a

mozzy11:
But theres still more work to be done

mozzy11:
@tendomart fee free to suggest any Carbon component that can switch between a tile into a full Page view

tendomart:
Well I think from the changes done, it looks more professional. 
It can always change if we discover something better .

thanks

tendomart:
> @tendomart fee free to suggest any Carbon component that can switch between a tile into a full Page view

Data table is the other competitor I see for the above though..But since the above works, we can keep it that way for now ??

https://carbondesignsystem.com/components/data-table/usage/

tendomart:
> But theres still more work to be done

Unfortunately wasn't, but will you break it into more sub-tasks.

tendomart:
@mozzy11  am closing off this PR since a better solution was provided

PR #695, opened 2024-01-23
Fix bugs in patient search form

rahul6603:
This PR addresses couple of bugs in [patient search form](https://openelis28.openelis-global.org/PatientManagement)

1. When all the fields are left empty and the form is submitted, it gets stuck in loading screen.
2. When the form is submitted with an invalid Previous Lab Number, the form gets stuck in loading screen.


mozzy11:
Thanks  @rahul6603

PR #721, opened 2024-02-21
Whonet

rossumg:
WHONET CSV export


dev/database/dbInit/OpenELIS-Global.sql
```diff
@@ -3254,7 +3254,7 @@ COMMENT ON COLUMN clinlims.orders.is_external IS 'Indicates if current order wil
 -- Name: COLUMN orders.external_order_number; Type: COMMENT; Schema: clinlims; Owner: clinlims
 --
 
-COMMENT ON COLUMN clinlims.orders.external_order_number IS 'External order number';
+COMMENT ON COLUMN clinlims.orders.external_order_number IS 'External lab number';
```
sherrif10, line 3257: Why lab number instead of order number here.

src/main/java/org/openelisglobal/reports/action/implementation/ExportWHONETReportByDate.java
```diff
@@ -0,0 +1,162 @@
...
+        errorFound = !validateSubmitParameters();
+        if (errorFound) {
+            return;
+        }
+
+        createReportItems();
+    }
+
+    /**
+     * check everything
```
sherrif10, line 72: Do you need this comment ?

sherrif10:
Hello @rossumg , Seems there are merge conflicts.

CalebSLane:
Tested the pull request locally. It looks like it is incomplete, so I am marking this as a draft. Notably, I had tho change some code in  `ReportImplementationFactory.java` to let it grab the report. Additionally the page for running the report doesn't allow the entry of fields (see screenshot attached) which causes the report to error (second screenshot). @caseyi I'm not sure if Greg didn't include some files and if so it would be useful to get those, but otherwise was this Abert's task originally and would it fall back on his plate?

image
image

mozzy11:
@CalebSLane i gues i can close this PR as you have already added these changes

PR #734, opened 2024-02-29
Improved Notification UX

konavivekramakrishna:
This PR adds the Pnotify package, which is the most preferred package for notifications.

The current implementation of notifications is overly complex, with excessive redundant code in place.

The overall unpacked size of Pnotify is 265 KB, which is comparatively less. By introducing this package, we can significantly reduce the usage of many hooks.

Furthermore, it improves UX by limiting the maximum number of notifications shown to 3. Previously, there was no limit, which was detrimental to the user experience.

### DIFF
--- a/frontend/src/components/common/Notification.js
+++ b/frontend/src/components/common/Notification.js
@@ -1,0 +1,54 @@
+import { alert, Stack, defaultModules } from "@pnotify/core";
+import * as PNotifyMobile from "@pnotify/mobile";
+import _ from "lodash-es";
+
+// Set up PNotify with mobile module
+defaultModules.set(PNotifyMobile, {});
+
+// Create a notification stack
+const notifyStack = new Stack({
+  dir1: "down",
+  dir2: "left",
+  firstpos1: 25,
+  firstpos2: 25,
+  modal: false,
+  maxOpen: 3,
+  maxStrategy: "close",
+  maxClosureCausesWait: false,
+  push: "top",
+});
+
+// Function to create and show a notification
+const notify = (text, type) => {
+  const notification = alert({
+    type: type,
+    text: text,
+    styling: "brighttheme",
+    mode: "light",
+    sticker: false,
+    buttons: {
+      closer: false,
+      sticker: false,
+    },
+    stack: notifyStack,
+  });
+  notification.refs.elem.addEventListener("click", () => {
+    notification.close();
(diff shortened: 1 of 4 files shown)


frontend/src/components/common/Notification.js
```diff
@@ -0,0 +1,54 @@
+import { alert, Stack, defaultModules } from "@pnotify/core";
+import * as PNotifyMobile from "@pnotify/mobile";
+import _ from "lodash-es";
+
+// Set up PNotify with mobile module
+defaultModules.set(PNotifyMobile, {});
+
+// Create a notification stack
```
sherrif10, line 8: Lets also try to add react component, i think we might want for every PR added, we should include react in our Pr.

frontend/src/components/common/Notification.js
```diff
@@ -0,0 +1,54 @@
...
+  firstpos1: 25,
+  firstpos2: 25,
+  modal: false,
+  maxOpen: 3,
+  maxStrategy: "close",
+  maxClosureCausesWait: false,
+  push: "top",
+});
+
+// Function to create and show a notification
```
sherrif10, line 21: Few suggestions.
- import React, {useSate} from React

Write the entire of this change as pure react. cc @mozzy11

sherrif10:
Some notifcation looks bigger than others, Keep the size of the alerts atleast with same size.

frontend/src/components/common/Notification.js
sherrif10: My whole point is . lets try to include react in our PRs.

sherrif10:
Can you share a vedio, it will be better to see notifications directly.

sherrif10 (commented):
Added a comment.

sherrif10:
Appologies for oversight . This Looks good. For better designs, i think we need to decide which kind of   react notifications to use across the entire systems. Either to use completely showToast notification or showSnackBar notication. cc @mozzy11

mozzy11:
@konavivekramakrishna we basically use [Carbon design system ](https://carbondesignsystem.com/)for all the compunenets to have consist design across the project

mozzy11:
I can close this PR .you can work on re-writitn some of our report in React . 
Pick any report and create an issue . 
see the reports  https://oe3.openelis-global.org/RoutineReports

Issue #747, opened 2024-03-04
 Fix Responsiveness of Modify Order Page


atulyadav745:
Issue Description:
The Modify Order page currently lacks responsiveness (on max-width <1056px), making it difficult for users to interact with the page across various devices and screen sizes. To enhance user experience and ensure usability, we need to address the responsiveness issues on this page.

To See issue visit: https://localhost/SampleEdit?type=readwrite
Current Behaviour
![befi](https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/106299466/0430b1d8-ff6d-4e74-811b-13a883114afe)
Expected Behaviour
![After1](https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/106299466/8db4ed47-4eaf-4d63-84f1-358414531b73)

AyushMaurya001:
Hi @atulyadav745 I would like to work on this issue.

PR #748, opened 2024-03-04
Fixed responsiveness of modify order page

atulyadav745:
This PR fixes the issue #747  which involves improving overall Viewport Compatibility, Layout Adjustment across all devices of modify order page.  (on max-width <1056px),

Before:
befi

After:
After1

@mozzy11 This PR is ready to review.
Thank you

### DIFF
--- a/frontend/src/components/modifyOrder/SearchOrder.js
+++ b/frontend/src/components/modifyOrder/SearchOrder.js
@@ -39,13 +39,10 @@ function SearchOrder() {
     <>
       <div className="orderLegendBody">
         <Form onSubmit={handleSearch}>
-          <Grid>
-            <Column lg={16}>
               <h4>
                 <FormattedMessage id="sample.label.search.labnumber" />
               </h4>
-            </Column>
-            <Column lg={4}>
+              <div className="inlineDiv">
               <CustomLabNumberInput
                 placeholder={"Enter Lab No"}
                 id="labNumber"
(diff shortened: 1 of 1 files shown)


frontend/src/components/modifyOrder/SearchOrder.js
```diff
@@ -39,13 +39,10 @@ function SearchOrder() {
     <>
       <div className="orderLegendBody">
         <Form onSubmit={handleSearch}>
-          <Grid>
-            <Column lg={16}>
```
adityadeshlahre, line 43: Hey @atulyadav745, in our codebase, grids and columns are used to maintain consistency across the code. Removing them isn't a good idea as they serve a purpose. We may need to find a better way to fix the responsiveness of the following pages!
Thank You :slightly_smiling_face:

CalebSLane: I don't have a strong preference one way or the other, but using Grid does allow a nice level of control at different viewport sizes. To fix the viewport issue while still using Grid would require specifying the column span at medium and small viewport sizes (ex. `<Column lg={16} md={8} sm={4}></Column>`) In React Carbon, the grid system is 16 wide at the large size, 8 at the medium, and 4 at the small. Details of Grid are [here](https://react.carbondesignsystem.com/?path=/story/components-ui-shell--header-base)

@mozzy11 do you have any opinions on which strategy should be preferred?

mozzy11: @CalebSLane  i agree with using the Carbon Grid. it also makes it easier to have  consistence across the project . Innitialy we used custom classes but it creates more issues as compared to using the Carbon Grid

codewithunknown (approved):
LGTM

mozzy11 (approved):

Issue #763, opened 2024-03-05
Fix Responsiveness of Home page


PR4NJ41:
### Issue Description
Homepage Dashboard lacks responsiveness for max screen width < 425px
### Steps to Reproduce
Visit: https://localhost/
### Current Behaviour
![Screenshot from 2024-03-06 04-35-23](https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/96368921/d24f6bc6-2584-45e9-95cd-90663e6100a6)

nishant1194:
can I take up this issue

PR4NJ41:
@nishant1194 This issue has already been resolved. So, I am closing this issue.

PR4NJ41:
@nishant1194 This issue has already been resolved. So, I am closing this issue.

PR #764, opened 2024-03-05
Fixed Responsiveness of home page

PR4NJ41:
Resolves Issue #763  

### Before
Screenshot from 2024-03-06 04-35-23

### After
Screenshot from 2024-03-06 04-33-33


mozzy11:
can you also attach a screenshot of the desktop screen?

Issue #762, opened 2024-03-05
Re-Write the Export Routine CSV report in the new Rect UI


mozzy11:
https://oe3.openelis-global.org/api/OpenELIS-Global/Report?type=patient&report=CISampleRoutineExport

Agaba-derrick:
helllo @mozzy11  am working on this ticket please assign

PR #766, opened 2024-03-07
Rewritten Export Routine CSV report in the new Rect UI

konavivekramakrishna:
This pr adds the Export Routine CSV report page

fixes : #762 

Before:
image

After:
image

### DIFF
--- a/frontend/src/components/Reports/Routine.js
+++ b/frontend/src/components/Reports/Routine.js
@@ -82,8 +82,7 @@ export const RoutineReportsMenu = {
       icon: IbmWatsonNaturalLanguageUnderstanding,
       SideNavMenuItem: [
         {
-          link:
-            config.serverBaseUrl + "/Report?type=patient&report=referredOut",
+          link:"/RoutineReport?type=patient&report=referredOut",
           label: <FormattedMessage id="sideNav.label.referredtestreport"/>,
         },
       ],
--- a/frontend/src/components/Reports/routine/Index.js
+++ b/frontend/src/components/Reports/routine/Index.js
@@ -15,6 +15,7 @@ import config from "../../../config.json";
 import StatisticsReport from "./StatisticsReport";
 import SummaryOfAllTest from "./SummaryOfAllTest";
 import HIVTestSummary from "./HivTestSummary";
+import ReferredOutTestSummary from "./ReferedOutTestSummary";
 
 const RoutineIndex = () => {
   const intl = useIntl();
@@ -88,6 +89,11 @@ const RoutineIndex = () => {
             {type === "indicator" &&
               report === "indicatorCDILNSPHIV" &&
               (<HIVTestSummary/>)}
+            {
+                type === "patient" && 
+                report === "referredOut" &&
+                (<ReferredOutTestSummary/>)
+              }
 
           </>
         )}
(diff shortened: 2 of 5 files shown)


mozzy11:
Thanks @konavivekramakrishna , 
Basically menu items are added using liquibase changesets to the DB ie [here](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop_3x/src/main/resources/liquibase/2.8.x.x/menu.xml#L102-L129) to I order to be compatible with the old jsp UI .,so we have the a single source of  menu items for both the new and old UI

mozzy11:
Which menu item did you need to add ??

mozzy11 (changes requested):
see above comments

mozzy11:
actually, use the 
config.serverBaseUrl + "/RoutineReport?type=patient&report=referredOut"
i renamed it so that all new RoutineReports in the Rect UI can use that 
https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop_3x/frontend/src/App.js#L382

konavivekramakrishna:
Okay @mozzy11 

I have another recommendation for the frontend code. How about using Husky as a pre-commit hook to automatically format the frontend code using Prettier? This way, the code remains consistent, and the formatting will be consistent as well. What do you think about it?

Can I create a separate PR for the adding pre commit hook to auto format using prettier?

konavivekramakrishna:
@mozzy11 

I have implemented the changes and attached the screenshot could you please review

frontend/src/components/Reports/routine/ReferedOutTestSummary.js
mherman22, line 1: i would probably go ahead and name my file as `CISampleRoutineExport.js`

mherman22 (changes requested):

frontend/src/components/Reports/Routine.js
```diff
@@ -82,8 +82,7 @@ export const RoutineReportsMenu = {
       icon: IbmWatsonNaturalLanguageUnderstanding,
       SideNavMenuItem: [
         {
-          link:
-            config.serverBaseUrl + "/Report?type=patient&report=referredOut",
+          link:"/RoutineReport?type=patient&report=referredOut",
```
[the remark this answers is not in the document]
mherman22: as you can see this `/RoutineReport?type=patient&report=referredOut` is for the referredOut yet this pull request is supposed to address something about ExportRoutineCSV report.

konavivekramakrishna:
@mozzy11 

can you please review

frontend/src/components/Reports/Routine.js
mozzy11, line 85: @konavivekramakrishna , 
@mherman22 is right . you worked on the wrong report menu item . 
YOu were supposed to work on [this ](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop_3x/frontend/src/components/Reports/Routine.js#L133-L144)  ie it shoulb become 
`"/RoutineReport?type=patient&report=CISampleRoutineExport",`

frontend/src/components/Reports/routine/Index.js
```diff
@@ -88,6 +89,11 @@ const RoutineIndex = () => {
             {type === "indicator" &&
               report === "indicatorCDILNSPHIV" &&
               (<HIVTestSummary/>)}
+            {
+                type === "patient" && 
+                report === "referredOut" &&
```
mozzy11, line 94: the reort is supposed to be `CISampleRoutineExport`

mozzy11 (changes requested):
Use the right report menu item

PR #784, opened 2024-03-09
Add Husky Pre-Commit Hook for Code Styling and ESLint Enforcement

konavivekramakrishna:
This PR adds Husky as a pre-commit hook, which is commonly used and preferred for enforcing linting stages. It further formats the code using Prettier and enforces rules using ESLint.

This ensures consistent code styling, removing blank spaces and additional formatting. ESLint will also flag any unused variables or functions, thereby improving efficiency by eliminating unnecessary imports and unused variables.

Implementing this can streamline the code review process significantly.


konavivekramakrishna:
@mozzy11 

can you please review this pr

as Far as our fronted code is concerned there are many unnecessary imports present in the code which causes the build takes more time and less efficiency and no pre-formatting

husky performs lint so that if there any unused imports in the staging files it rejects the git commit and also formats the code before commit so there by no unnecessary blank spaces will be there

can you please let me know your suggestions

mozzy11:
Thanks @konavivekramakrishna . 
The docker build fails probably because of Husky enforcing rules. 
it will require cleaning out all those errors for the build to pass. 
see logs here https://github.com/I-TECH-UW/OpenELIS-Global-2/actions/runs/8338958896/job/22820126387?pr=784

frontend/package.json
```diff
@@ -42,16 +44,15 @@
     "styled-components": "^5.3.6",
     "swr": "2.0.3",
     "web-vitals": "^2.1.4",
-    "yup": "^0.29.2",
```
mozzy11, line 45: Why did you remove these dependencies ??
Am sure they are used some where in the project

mozzy11 (changes requested):
@konavivekramakrishna , 
we can not use   `/* eslint-disable */` .

Ankit-Matth:
Hey @konavivekramakrishna,
It seems like the Docker build is still failing, possibly due to Husky enforcing rules. If you're currently working on fixing it, that's great! However, if you don't mind, I'd like to take a look and see if I can help resolve the issue. Feel free to let me know if you'd like me to give it a try.

Ankit-Matth:
> Hey @Ankit-Matth,
> 
> Let me explain in detail about everything.
> 
> Currently, when we commit, technically Husky will enforce rules using the command `npx lint-staged` to check for any unused variables or if we're conditionally using React hooks, which is not preferred.
> 
> for example :
> 
> ```js
> if (props.isTrue === true){
>   const [val, setVal] = useState(true);
> }
> ```
> 
> In such cases, Husky will prevent the commit and also format before committing, ensuring code consistency.
> 
> To achieve this behavior, we have to modify the `eslintconfig`. If you review our frontend codebase, you'll notice many unused variables, imports, functions, and even instances of conditionally using hooks. Therefore, before this PR can be functional, we need to remove all unused variables and functions from the frontend codebase, which constitutes a significant part of our changes.
> 
> If you understand the above and are willing to contribute, please drop a comment indicating your willingness. I'll open an issue, and through multiple PRs, we'll carefully evaluate and remove all unused variables and functions, as well as eliminate conditionally used hooks.
> 
> cc : @mozzy11

Hey @konavivekramakrishna,

You're absolutely correct; there are indeed many unused variables, imports, functions, and even instances of conditionally using hooks in our frontend codebase. However, it seems the current Docker build failure might be unrelated and likely due to a mismatch between the **package.json** and **lock file**. Below, I've attached some screenshots of the **build-and-push-image** logs for reference:

build-and-push-image-1

build-and-push-image-2

To address this issue, you need to update the **lock file** by running `npm install` locally to ensure it reflects the dependencies specified in our **package.json**.   

And regarding the removal of these unused variables, imports, functions, and even instances of conditionally using hooks, I agree with your approach. It's essential to involve @mozzy11 in this discussion to gather their insights and perspectives. Collaboratively approaching this task will help us consider various viewpoints and potential risks involved.

PR #815, opened 2024-03-15
[test]:  fixed failing test & add delete patient test

valens200:
Fixed the failing test which was failing all the times.

Initially, this test expects to receive several patients equal to one but in real implementation, the output of `"patientService.getAllPatients().size()"` is zero which leads to the failure of the test. 

I also added delete patient test to ensure the high quality of the functionality.


valens200:
@mozzy11  @reagan-meant you can review this PR and suggest changes. 
Thank you.

volume/tomcat/hapi_server.xml
```diff
@@ -19,17 +19,17 @@
      define subcomponents such as "Valves" at this level.
      Documentation at /docs/config/server.html
  -->
-<Server port="-1" shutdown="SHUTDOWN">
```
mozzy11, line 22: get rid of these changes , its probably updated the by the container volume.

mozzy11:
Thanks @valens200 , 
did your tests pass locally ??

valens200:
Yes @mozzy11, I think the action build is failing due to changing this port as you requested.
Screen Shot 2024-03-18 at 21 57 53

Setting it to -1 makes the tests to failure because the data source can't be found.

valens200:
> Thanks @valens200 , did your tests pass locally ??



> @mozzy11 @reagan-meant you can review this PR and suggest changes. Thank you.

yes

mozzy11:
@valens200 i dont think thats right , 
These are integration tests at the service layer and they dont have any thing to do with the servlet container .

src/test/java/org/openelisglobal/patient/PatientServiceTest.java
```diff
@@ -35,8 +38,27 @@ public class PatientServiceTest {
...
+		Assert.assertEquals(1, patientService.getAllPatients().size());
+	}
+	@Test
+	public void deletePatient_shouldDeleteNewPatient() throws Exception {
+		String firstName = "Peter";
+		String lastname = "Master";
+		String dob = "9/12/1993";
+		String gender = "M";
+		Patient pat = createPatient(firstName, lastname, dob, gender);
+		String patientId = patientService.insert(pat);
```
mozzy11, line 55: can you assert the patient exists before delete ?

volume/tomcat/hapi_server.xml
mozzy11: we dont need these changes here for the tests to pass

mozzy11 (changes requested):
Get rid of the server,xml changes and also add an assertion before delete

valens200:
I see @mozzy11 , let me apply the changes

valens200:
> @valens200 i dont think thats right , These are integration tests at the service layer and they dont have any thing to do with the servlet container .

Yes sir @mozzy11

valens200:
@mozzy11 I think that's all.

valens200:
@mozzy11 you can check for changes.

mozzy11 (changes requested):
Thanks @valens200 . 
you still have format changes in hapi_server.xml file . 
amd also can you also put the Guide changes in a separate PR ??

valens200:
@mozzy11 If you are here please you can see re-review this.

PR #891, opened 2024-03-25
Adding unit tests for address

sayed4900:
This the same code of this pull request [](https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/793) but I move the files to folder unitTests as the tests I made is unit test not integration test.

@mozzy11


mozzy11 (changes requested):
Thanks @sayed4900 for attemtoing this. 
For consistence , 
test methods should follow a naming convetion like 
`MethodName_ExpectedBehavior` ie _getOrganization_shouldReturnOrganization_

mozzy11:
Thanks @sayed4900 .We needed to real integration tests that run in the testdb not mocked tests

Issue #788, opened 2024-03-10
Duplicate National IDs are persisted even after setting the 'duplicate national ID' property false 


common-cold:
Issue Description:

The Patient Configuration Menu has fields such as 
- Allow duplicate national ids 
- Allow duplicate subject number

Even after toggling their values to false, new patients can be persisted  with duplicate national IDs and subject Number. 
Currently there is no Validator present in the backend during the process of saving a new patient.

1. Patient Configuration Menu can be accessed by :

Admin -> Legacy Admin -> Patient Entry Configuration

Here the properties can be toggled to true or false.

Screenshots of the issue:

Duplicate National ID Property is set to false:

![Screenshot 2024-03-11 000026](https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/71568044/8ed9e754-b610-434f-b77f-83fac28505c5)

Patients with duplicate national IDs are being persisted:

![Screenshot 2024-03-11 000217](https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/71568044/7c1028c1-2d29-4831-bedd-b46ad30fdbf8)

common-cold:
@mozzy11 can you assign this issue to me.

codewithunknown:
@mozzy11 @common-cold 

just a heads up maybe, what I would say is there can be mutiple patients with same national ID

@mozzy11  I would suggest by default we can keep allowing mutiple patients with same national ID because say I am from `uganda` then there would be no point if only one patient from uganda can use our application

just reconsidering the edge cases

common-cold:
@codewithunknown from what I could understand the national ID in the codebase refers to the unique identification no. given by the government to its citizen. In India we have Aadhaar Card as our National ID. I don't think the code base refers to the Country Code as national Id. Correct me if I'm wrong.

mozzy11:
@common-cold any progress or blocker here ??

mozzy11:
hello @common-cold any progress ??

common-cold:
@mozzy11 yes i have done changes in the code-base I'll generate a PR soon.

PR #909, opened 2024-03-26
Fix Duplicate National Id/Subject No. persistence 

common-cold:
# Pull Requests Requirements

- [x] The changes confirm to the OpenElis Global x3 Styleguide and design documentation.
- [x] I have read and agree to the Contributing Guidelines of this project.

## Summary
fixes: [#788 ](https://github.com/I-TECH-UW/OpenELIS-Global-2/issues/788)

The current validator named **validatePatientInfo()** in **PatientServiceRestController** class is not properly handling this issue. It only focusses on the duplicate SubjectNumber's value. 

While setting  
- duplicate Subject Number = false , and
- duplicate National ID = true

I try to save a new patient with new Subject Number and a duplicate national ID I get an error while persisting this patient. This happens due to incomplete execution of **preparePatientData()** in **PatientManagementRestController** class. This further leads to an error from hibernate saying "attempt to create saveOrUpdate event from null entity". Which should not happen.  

This pull request adds a new validator named **validatePatientInfo()** in **persistPatientData()** inside **PatientServiceImpl** class, and sets the response status to 400 when the validator fails.
This new validator is taking account both subject Number and National ID. And works for all permutations for TRUE/FALSE.

## Screenshots

1. 

Screenshot 2024-03-24 041500
Screenshot 2024-03-24 041607
Screenshot 2024-03-24 041623

2.
Screenshot 2024-03-24 041739
Screenshot 2024-03-24 041919
Screenshot 2024-03-24 042005

3.
Screenshot 2024-03-24 042049
Screenshot 2024-03-24 042339
Screenshot 2024-03-24 042424
Screenshot 2024-03-24 042516

4.
Screenshot 2024-03-24 042645
Screenshot 2024-03-24 042820

## Other
The current validator is not properly validating the conditions. And also focusses on subjectNumber parameter only. 
GIven that the new validator is working fine I suggest we can replace the old validator from the new validator entirely.


src/main/java/org/openelisglobal/patient/service/PatientServiceImpl.java
```diff
@@ -832,4 +844,45 @@ public Patient getByExternalId(String id) {
         return baseObjectDAO.getByExternalId(id);
     }
 
+
+    private Errors validatePatientInfo(PatientManagementInfo patientInfo) {
```
mozzy11, line 848: Why dont you just add this validation function in the controller wherw it was before ??

src/main/java/org/openelisglobal/patient/controller/rest/PatientManagementRestController.java
```diff
@@ -101,10 +103,10 @@ public void savepatient(HttpServletRequest request, @Validated(SamplePatientEntr
     private void preparePatientData(Errors errors ,HttpServletRequest request, PatientManagementInfo patientInfo,
             Patient patient) throws IllegalAccessException, InvocationTargetException, NoSuchMethodException {
 
-        validatePatientInfo(errors, patientInfo);
-        if (errors.hasErrors()) {
-           return;
-        }
+//        validatePatientInfo(errors, patientInfo);
```
mozzy11, line 106: can you just update this function to catch duplicate IDs ?

mozzy11 (changes requested):
I think move the validation function back to the controller , it will also fix the CI build

common-cold:
@mozzy I have commented out the old validator instead of deleting it completely. Let me know if you want it to be removed

common-cold:
@mozzy11 please review this

common-cold:
@mozzy11 any feedback on why the frontend framework is failing?

src/main/java/org/openelisglobal/patient/controller/rest/PatientManagementRestController.java
```diff
@@ -216,4 +216,43 @@ private Patient loadForUpdate(PatientManagementInfo patientInfo) {
         return patient;
     }
 
+    private void validatePatientInfo(Errors errors, PatientManagementInfo patientInfo) {
```
mozzy11, line 219: Thanks @common-cold , can you just update the existing functionality rather than commenting it out ??

mozzy11 (changes requested):
you can  just refactor and update the existing function rather than commenting it out and creating another

common-cold:
@mozzy11 I have done the changes

mozzy11 (approved):

PR #956, opened 2024-04-02
Add unit and integration tests for ReferenceTableService.

valens200:
## Summary

This pull request introduces unit and integration tests for the ReferenceTables service.
Mockito was used to mock external dependencies when executing unit tests focusing on the service’s functionality. The DAO functionalities are included during integration tests.

Related Issue: [Add Unit  and integration Tests for ReferenceTableService ](https://github.com/I-TECH-UW/OpenELIS-Global-2/issues/870)


valens200:
@sherrif10 @HerbertYiga you can review this pull request and suggest changes.

Thank you!

src/test/java/org/openelisglobal/integrationTests/referenceTables/ReferenceTablesServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+    private ReferenceTables result;
+
+
+    @Test
+    public void insert_shouldInsertReferenceTable(){
+        expected = ObjectsInitiator.createReferenceTable("X-Table","N","Y");
+        id = referenceTablesService.insert(expected);
+
+        result = referenceTablesService.get(id);
+        assertEquals(expected.getIsHl7Encoded(),result.getIsHl7Encoded());
```
CalebSLane, line 40: Could you create a method that does assertEquals for every field except id that we can re-use between tests? Ideally we would like to know that all fields are entered successfully on insert, not just ishl7Encoded

src/test/java/org/openelisglobal/integrationTests/referenceTables/ReferenceTablesServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+        id = referenceTablesService.insert(expected);
+        expected = referenceTablesService.get(id);
+
+        result = referenceTablesService.update(expected);
+
+        assertEquals(expected.getTableName(), result.getTableName());
+        assertEquals(expected.getKeepHistory(), result.getKeepHistory());
+        assertEquals(expected.getIsHl7Encoded(), result.getIsHl7Encoded());
+        assertEquals(expected.getId(),result.getId());
```
CalebSLane, line 61: can you move the changes (line 49-51) so that they happen in between the insert and the update, so we can test that the object does get successfully updated? 

maybe test
- insert New table
- assert insert success as is
- change values
- call update
- assert update success on result

Then maybe create second method that does all the same steps except for the last one and instead:
- get updated values by id
- assert success on result of get
- assert success that result of get = result of update

This would enable us to see that the result returned by the update is the same as what is saved to what can be fetched from the database

src/test/java/org/openelisglobal/integrationTests/referenceTables/ReferenceTablesServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+        assertNotNull(result);
+        assertEquals(expected.getTableName(), result.getTableName());
+        assertEquals(expected.getKeepHistory(), result.getKeepHistory());
+        assertEquals(expected.getIsHl7Encoded(), result.getIsHl7Encoded());
+    }
+    @Test
+    public void getAllForHl7Encoding_shouldGetAllReferenceTablesForHl7Encoding() {
+        List<ReferenceTables> referenceTablesList = referenceTablesService.getAllReferenceTablesForHl7Encoding();
+        assertNotNull(referenceTablesList);
+        assertTrue(referenceTablesList.size() > 0);
```
CalebSLane, line 80: can you assert that the value is exactly right instead of just `>0` if the test misses an entry that it should grab this would still pass

src/test/java/org/openelisglobal/integrationTests/referenceTables/ReferenceTablesServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    public void getAll_shouldGetAllReferenceTables() {
+        List<ReferenceTables> referenceTablesList = referenceTablesService.getAllReferenceTables();
+        assertNotNull(referenceTablesList);
+        assertTrue(referenceTablesList.size() > 0);
```
CalebSLane, line 87: can you assert that the value is exactly right instead of just `>0` if the test misses an entry that it should grab this would still pass

src/test/java/org/openelisglobal/integrationTests/referenceTables/ReferenceTablesServiceTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    public void getTotalTablesCount_shouldGetTotalReferenceTablesCount() {
+        Integer totalCount = referenceTablesService.getTotalReferenceTablesCount();
+        assertNotNull(totalCount);
```
CalebSLane, line 93: can you assert that the totalCount is exactly right instead of just not null? this won't detect a miscount.

CalebSLane (commented):
A few comments of things to change. I am not the most familiar with Mockito tests, so I haven't added comments there, but it looks good!

valens200:
Thank you @CalebSLane , I will take a look into this.

mozzy11:
Thanks @valens200 . We currently rely on datasets for data innitialization

Issue #959, opened 2024-04-02
Fix Login E2E Tests


abertnamanya:
Fix E2E tests for the user login events, Here is the current running code. 
https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop_3x/frontend/cypress/e2e/orderEntity.cy.js

manishjha-04:
Hey @abertnamanya is the login E2E tests still failing? i have fixed it in PR #946 is that what you are talking about?

Also in my local fork i have ran the QA workflow multiple times its passing all the tests!

sherrif10:
is this ticket still valid ? cc @abertnamanya

abertnamanya:
@sherrif10 Its already fixed

PR #963, opened 2024-04-03
Fix Login E2E Tests

manishjha-04:
- [ ]  #959 

**>Performed QA workflow runs 10-15 times to ensure it does not fail.** 

Here is the attached screenshot

Screenshot 2024-04-03 211444


manishjha-04:
Hi @mozzy11 , I noticed that during multiple runs, the login test was encountering a DOM detach error. I've now re-queried the element before each action, and after running the QA workflow multiple times to check, it's performing perfectly. I hope this resolves the issue.

mozzy11 (approved):

manishjha-04:
Also, @mozzy11, please let me know if any issues arise!

Issue #897, opened 2024-03-25
Re -write Non Conform Section to the new React UI


mozzy11:
Go to the Main Menu - > Non Conform 
ie 
* https://oe3.openelis-global.org/api/OpenELIS-Global/ReportNonConformingEvent
*  https://oe3.openelis-global.org/api/OpenELIS-Global/ViewNonConformingEvent
*  https://oe3.openelis-global.org/api/OpenELIS-Global/NCECorrectiveAction

YelyzavetaKo:
@mozzy11 
Hi, I would like to work on this issue if still available.

karthikrajpalavalasa:
@mozzy11 can you assign this issue to me?

konavivekramakrishna:
@mozzy11 

can you assign this to me, I have implemented the UI currently working POST requests

konavivekramakrishna:
@mozzy11 

sorry for the delay. I'm encountering few blockers while dealing with `nceNumber`. I'll complete `viewNonConformingEvent` by today

konavivekramakrishna:
- ![image](https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/101407963/89c64b39-56da-4fe1-9249-980e8e5661a3)
- working on `viewNonConformingEvent` post request

cc : @mozzy11

PR #1036, opened 2024-05-07
Implement Corrective Actions 

konavivekramakrishna:
Fixes : #897


src/main/java/org/openelisglobal/common/provider/query/NonConformingEventSearchProvider.java
```diff
@@ -66,10 +66,6 @@ public void formatResults(List<NcEvent> results, StringBuilder xml) {
             XMLUtil.appendKeyValue("unit", reportingUnit, xml);
             XMLUtil.appendKeyValue("color", nce.getColorCode(), xml);
             xml.append("</nce>");
-//            xml.append("<nce><date>").append(nce.getDateOfEvent()).append("</date><ncenumber>");
```
mozzy11, line 69: why did you have to comment these out ??

mozzy11 (approved):

PR #1064, opened 2024-06-03
Implement persistent Lucene indexing with Docker volumes

rahul6603:

### Summary
- To ensure that the search indexes persist when the backend container (_openelisglobal-webapp_) restarts, we will use Docker volumes.

- We will create a directory to store Lucene indexes when the container is created, and then define a volume for this directory to ensure that the indexes are not lost when the container is restarted or recreated.


rahul6603:
Hi @mozzy11 , could you please review this PR?
Thank you.

PR #1083, opened 2024-06-10
Add bidirectional association for patients in person mapping

rahul6603:

### Summary

- We will add `<set>` element to _Person_ mapping to establish bidirectional association with _Patient_.
- This change is needed to prevent the Hibernate Search exception caused when using `@IndexedEmbedded` with a unidirectional association. We will be using `@IndexedEmbedded` to embed the fields of _Person_ object in the main _Patient_ object.
- [**Reference documentation**](https://docs.jboss.org/hibernate/search/6.1/reference/en-US/html_single/#:~:text=the%20association%20must%20be%20bidirectional)


CalebSLane (commented):
A few questions for @rahul6603:

How does this work with the `Provider` class? I see that the `Person` class does include a `Set` for holding Patients (which should be `Set<Patient>` instead of just `Set` as it is now), but the `Person` object is also used by the `Provider` class. Does this class need to get a mapping like this for `Provider` as will, and does adding the mapping cause either class to be unhappy?

Does saving logic need to be reviewed and changed? Will saving a `Person` with an empty `Set<Patient>`, then adding the `Person` to the `Patient`, and saving the `Patient` (without adding the `Set<Patient>` to the `Person`) allow entities to be saved properly?  Does the bi-directionality demand that the `Patient` be added into the `Person` object's `Set<Patient>` as well? 

Are there integration or unit tests to test these cases? There are a few classes with methods that call the insert, save, and/or update on these objects that I'm unsure of how they'll handle the change ex `persistPatientData` in `PatientManagementUpdate.java`

rahul6603:
Hi @CalebSLane, thanks for the review!

1. We need bidirectional association between _Patient_ and _Person_ entity for `@IndexedEmbedded` to work ([reference](https://docs.jboss.org/hibernate/search/6.1/reference/en-US/html_single/#:~:text=the%20association%20must%20be%20bidirectional)), you can see the use of `@IndexedEmbedded` in this draft PR - https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/1095. This should not affect other unidirectional associations such as the one between _Provider_ and _Person_.
PS - Thanks for mentioning `Set` should be `Set<Patient>`, I have made that change in the mentioned PR.

2. Yes, bidirectional associations require updates on both sides of the association when we persist an entity. But in our case, there is no usage of `person.getPatients()` in the codebase, so I don't think change in the persistence logic is needed.

3. I don't think the unit tests and integration tests exist for these cases. I will add them as part of my GSoC project in the upcoming weeks.

Please let me know if you can think of any side effects this change might cause.
Thank you.

rahul6603:
Hi @mozzy11 and @CalebSLane, could you please review this PR?
Thank you.

mozzy11:
Thanks @rahul6603 . 
The chane here cals for modifying the patients field in the person class from  `Set patients` to `Set<Patient> patients`  .
Please do all that in this same PR and some integration tests on adding multiple patients for a single person to ensure no Logic breaks. 
We already have Patient and Person integration tests set up . so its a matter of adding a few tests to ensure we are safe

mozzy11:
@CalebSLane , In theory , no change is needed in the saving Logic and neither  for the Provider Class.
The maapings are generally used internally by hibernate to map associated parent and child objects .  in otherwords instead of writing manual queries to get associated objects in case no hibernamte maaping exists , one can just do` person.getPatients()` , and hibernate does the querying internally

mozzy11 (changes requested):
see above comment

src/main/resources/hibernate/hbm/Person.hbm.xml
```diff
@@ -61,5 +61,11 @@
         <property name="email" type="java.lang.String">
             <column name="EMAIL" />
         </property>
+        <set name="patients" table="PATIENT" inverse="true" fetch="select" lazy="false">
```
mozzy11, line 64: modify the patients property field in the Person class accordingly

PR #1095, opened 2024-06-13
Add mapping for indexing Patient and Person entities

rahul6603:

### Summary

- We will use `@Indexed` and `@*Field` (such as `@KeywordField`) annotations to map _Patient_ and _Person_ entities and their fields to Lucene indexes.
- We will create a custom _normalizer_ that includes _lowercase_ and _ascii-folding_ filters for the _Keyword_ fields. These filters will convert every character to lowercase and replace characters with diacritics ("é", "à", …​) with their ASCII equivalent ("e", "a", …​).

### DIFF
--- a/src/main/java/org/openelisglobal/person/valueholder/Person.java
+++ b/src/main/java/org/openelisglobal/person/valueholder/Person.java
@@ -17,6 +17,11 @@ import java.util.HashSet;
 import java.util.Set;
 import javax.validation.constraints.Email;
 import javax.validation.constraints.Pattern;
+
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.AssociationInverseSide;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.KeywordField;
+import org.hibernate.search.mapper.pojo.mapping.definition.annotation.ObjectPath;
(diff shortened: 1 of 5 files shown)


rahul6603:
Hi @mozzy11 , could you please review this PR?
Thank you.

src/main/java/org/openelisglobal/person/valueholder/Person.java
```diff
@@ -62,7 +69,8 @@ public class Person extends BaseObject<String> {
   private String fax;
   @Email private String email;
 
-  private Set patients = new HashSet(0);
+  @AssociationInverseSide(inversePath = @ObjectPath(@PropertyValue(propertyName = "person")))
+  private Set<Patient> patients = new HashSet<>(0);
```
mozzy11, line 73: Can you move this change ie `Set<Patient> patients  `  to this [PR ](https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/1083)  so that you can add some few Tests on `person.getPatients() `

mozzy11 (changes requested):
move the field modification change to its corresponding [PR](https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/1083) and only leave changes realted to search engine

PR #1180, opened 2024-07-14
Implement Lucene queries in getSearchResults and getSearchResultsExact methods of LuceneSearchResultsDAOImpl

rahul6603:

### Summary

- We will implement Lucene queries in `getSearchResults` and `getSearchResultsExact` methods of `LuceneSearchResultsDAOImpl` class.
- This change will also include tests for the `LuceneSearchResultsServiceImpl` which will replace the existing placeholder tests in `SearchResultsServiceTest`.


src/main/java/org/openelisglobal/sample/daoimpl/DBSearchResultsDAOImpl.java
```diff
@@ -321,57 +321,29 @@ private String buildQueryString(boolean lastName, boolean firstName, boolean STN
         queryBuilder.append(ID_TYPE_FOR_GUID);
         queryBuilder.append(" where ");
 
-        if (anyID) {
```
mozzy11, line 324: @rahul6603  where these changes necesarry ??

rahul6603: I have removed this _if-else_ check because the logic in both the _if_ and _else_ blocks is same. It is not related to the main changes in the PR but I  included it because I want the `buildQueryString` methods in `LuceneSearchResultsDAOImpl` and `DBSearchResultsDAOImpl` to have the same logic.
Let me know if this change needs to be reverted or to be put in a separate PR.

mozzy11: @rahul6603 , yeah , Thanks for catching That

PR #1201, opened 2024-07-23
Minor Fix 

manishjha-04:
> Removed Unnecessary links from readme


mozzy11:
Thanks @manishjha-04 for the catch

PR #1323, opened 2024-11-21
Creating the Patient view React ui

Isabirye1515:
Screenshot from 2024-11-21 06-51-53


Isabirye1515:
@mozzy11 , its alot in your hands as per now but some thing is not coming with the selections for hiv status and like , if you will get some time and i get some advice on that

PR #1345, opened 2024-12-10
1339: Save Button Disabled During API Call, Disllowing Multiple Requests

aayush6194:
# Pull Requests Requirements

      Issue number if applicable.
- [ ] The changes confirm to the OpenElis Global x3 Styleguide and design
      documentation.
- [X] I have read and agree to the Contributing Guidelines of this project.

## Summary
- Added `isSubmitting`  to track the state of the API call.
- The Save Button disabled while `isSubmitting` is true.
- `postToOpenElisServer()` return a Promise, allowing to chain additional .then() handler


mozzy11:
Thanks @aayush6194

mozzy11 (dismissed):

frontend/src/components/utils/Utils.js
```diff
@@ -35,7 +35,7 @@ export const postToOpenElisServer = (
   callback,
   extraParams,
 ) => {
-  fetch(
+  return fetch(
```
mozzy11: @aayush6194 , we already pass a [call back](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop/frontend/src/components/utils/Utils.js#L54) to the fetch function when it returns the Response.

you can revert this change here so that we are consistent across the API call utilities

frontend/src/components/patient/CreatePatientForm.js
```diff
@@ -330,7 +336,9 @@ function CreatePatientForm(props) {
           days: "",
         });
       },
-    );
+    ).then(() => {
+      setIsSubmitting(false);
```
mozzy11: you can instead set this directly in the call back [here](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop/frontend/src/components/patient/CreatePatientForm.js#L325) ,when the call returns the Response , like we do for the other post-Submit  actions

mozzy11:
Thanks @aayush6194 , just added a few minor comments , just make sure we have consistence across all the API call utilities

mozzy11 (dismissed):

PR #1404, opened 2025-01-14
Added comprehensive tests for AddressPartService

Lemeri123:
Added tests for handling Null Data, Empty Part Name, Duplicate Address Part, Invalid Display Order,


Lemeri123:
Please check this @rahul6603

src/main/java/org/openelisglobal/address/service/AddressPartService.java
```diff
@@ -8,4 +8,5 @@ public interface AddressPartService extends BaseObjectService<AddressPart, Strin
     List<AddressPart> getAll();
 
     AddressPart getAddresPartByName(String name);
+    AddressPart getAddresPartById(String id);
```
mozzy11, line 11: did you add this new method for the sake of tests ??

Lemeri123: Yes.  I was getting an error retrieving an AddressPart object by its id. Should I remove it??

PR #1415, opened 2025-01-19
Fix DatePickerInput to Validate Numeric and Slash Characters Only

### LINKED ISSUE #1413: Validate DatePickerInput to Allow Partial Input with Numbers and Slashes While Typing
# How to Reproduce this issue
1. login
2. navigate to results --> Results --> By Patient
3. Try inputting alphabets in the date picker input field

# Current behaviour
Users are allowed to input alphabets even though they shouldn't be

# Expected behaviour 
User should only be allowed to input numbers and slashes (/) in the correct format

# Screen Recording
https://github.com/user-attachments/assets/46488d86-eb67-4b2d-8561-baa5e3e97612

harshitg927:
Fixes: [issue#1413](https://github.com/I-TECH-UW/OpenELIS-Global-2/issues/1413)

## Summary

This PR implements input validation for the DatePickerInput component in the Results/By Patient view. The changes enforce that users can only enter numbers and forward slashes while typing, preventing alphabetic and special character input. This improvement ensures proper date format entry and enhances form validation.

The User will no longer be allowed to input characters like (abc) etc. Instead when the user types a character that is not allowed, the input field will default to an empty string.

## Screen Recording

[Screencast from 20-01-25 00:51:56.webm](https://github.com/user-attachments/assets/c7bd703e-e236-4d86-a62f-e26e3abb0838)

## Related Issue

https://github.com/I-TECH-UW/OpenELIS-Global-2/issues/1413

### DIFF
--- a/frontend/src/components/common/CustomDatePicker.js
+++ b/frontend/src/components/common/CustomDatePicker.js
@@ -20,6 +20,25 @@ const CustomDatePicker = (props) => {
     props.onChange(currentDate);
   }
 
+  function handleInputChange(e) {
+    const inputValue = event.target.value;
+
+  const isFrenchLocale = configurationProperties.DEFAULT_DATE_LOCALE === "fr-FR";
+  const partialDateRegex = isFrenchLocale
+    ? /^(\d{0,2})(\/(\d{0,2})(\/(\d{0,4})?)?)?$/ 
+    : /^(\d{0,2})(\/(\d{0,2})(\/(\d{0,4})?)?)?$/;
+
+  const fullDateRegex = isFrenchLocale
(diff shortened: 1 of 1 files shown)


harshitg927:
@mozzy11 Can you please review this pull request, Thank You.

vsvishalsharma:
Hi @harshitg927,

Just checking in to confirm—have the checks been applied to all the custom date inputs?

Also, instead of clearing the input entirely when there's an invalid entry, could we adjust it to simply ignore the wrong input and retain the valid part? I think this might improve the user experience.

mozzy11 (approved):

Issue #1426, opened 2025-01-22
Repeat password field is turning red although it has not been changed


rahulkundu1209:
# How to Reproduce
1. Login using user name 'admin'
2. Go to Admin -> User Management
3. Click on the Add button to add new user
4. Enter Login Name and then password in the input fields

# Current Behaviour
The Repeat Password field is turning red as if invalid value has been entered although no value has been entered in the field.

# Expected Behaviour
The Repeat Password field should turn red only when a value has been entered and it is not valid.

# Screen Recording

https://github.com/user-attachments/assets/7f0b9ae3-d925-49bb-b5e1-7f623efe7910

rahulkundu1209:
Work in progress.

mozzy11:
Thanks @rahulkundu1209

rahulkundu1209:
Thank you very much @mozzy11 for assigning me the issue and approving the pull request.

PR #1427, opened 2025-01-22
Repeat password input field turns red when a value is entered in the repeat password field and it differs from the password.

rahulkundu1209:
## Summary
The condition of invalid repeat password has been changed. Now the repeat password field becomes red only when the repeat password input field has been changed and the repeat password is invalid as per the earlier conditions.

## Screen Recording
https://github.com/user-attachments/assets/b8dde5f6-77bc-49e7-a08c-9092539b3ef8

## Related Issue

#1426

## Other
This change will just enhance user experience.


mozzy11:
Thanks @rahulkundu1209

mozzy11 (approved):

PR #1453, opened 2025-01-27
Develope

Isabirye1515:
[Screencast from 27-01-25 23:23:36.webm](https://github.com/user-attachments/assets/c1f41d27-2159-4a3c-9e74-2eb4189ebf76)


frontend/src/components/patient/patientEdit.js
```diff
@@ -0,0 +1,255 @@
...
+  MultiSelect,
+  RadioButton,
+  RadioButtonGroup,
+  TextInput,
+} from "@carbon/react";
+import React, { useState, useEffect } from "react";
+import { postToOpenElisServerJsonResponse } from "../utils/Utils";
+
+const PatientEditByProjectForm = ({ formData, projectData }) => {
+  if (!formData) {
```
vsvishalsharma, line 18: @Isabirye1515 after looking at your build error

The component's behavior appears to **conditionally call hooks** based on the presence of `formData`. When formData is `null `or `undefined,` the component returns early, preventing the hooks from being invoked. However, when formData exists, the hooks are executed.

This conditional execution of hooks could be a potential cause of the build error, as React hooks must be called unconditionally and in the same order during every render cycle.

PR #1455, opened 2025-01-28
[fix] Updated the patientEntry E2E

Bahati308:
# [fix] Updated the patientEntry E2E

## Screenshots
Screenshot from 2025-01-28 18-57-51

## Related Issue
https://uwdigi.atlassian.net/browse/OGC-13
## Other
https://uwdigi.atlassian.net/browse/OGC-28


frontend/cypress/pages/WorkPlan.js
```diff
@@ -11,15 +11,42 @@ class WorkPlan {
...
+
+  getTestTypeUnit(unitType) {
+    cy.get("#select-1").select(unitType);
+  }
+
+  getTestTypePriority(priority) {
+    cy.get("#select-1").select(priority);
   }
 
   getPrintWorkPlanButton() {
```
CalebSLane, line 30: This methods should be renamed since it is no longer a "get" operation

frontend/cypress/pages/WorkPlan.js
```diff
@@ -11,15 +11,42 @@ class WorkPlan {
...
-    return cy.get("button#print");
+    cy.get("#print", { timeout: 10000 })
+      .should("exist") 
+      .should("be.visible")
+      .click();
   }
-  getWorkPlanResultsTable() {
-    return cy.get(".cds--data-table.cds--data-table--sm");
+
+  getFinalPrintWorkPlanButton() {
```
CalebSLane: This methods should be renamed since it is no longer a "get" operation

frontend/cypress/pages/ResultsPage.js
```diff
@@ -7,18 +7,20 @@ class Result {
...
 
-  acceptSample(index = 0) {
-    cy.get(`.cds--checkbox-label`).eq(index).click();
+  clickExpanderBtn() {
+    cy.get('[data-testid="expander-button-0"]').click();
+  }
+  acceptSample() {
+    //div.cds--form-item.cds--checkbox-wrapper > input[type="checkbox"]
+    cy.get(
+      'div.cds--form-item.cds--checkbox-wrapper > input[type="checkbox"]',
```
CalebSLane, line 16: This still looks like it is using the page format selectors. Is there a reason why an id can't be used?

frontend/cypress/pages/PatientEntryPage.js
```diff
@@ -1,170 +1,164 @@
...
+  selectNationality(nationality) {
+    cy.get(this.nationality).select(nationality);
+  }
+
+  selectOtherNationality(otherNationality) {
+    cy.get(this.otherNationality).type(otherNationality);
+  }
+
+  selectPatient() {
+    cy.get("table tbody tr").first().find("td label span").click();
```
CalebSLane, line 161: can an id selector be used instead of page format selectors here?

frontend/cypress/pages/OrderEntityPage.js
```diff
@@ -46,14 +47,21 @@ class OrderEntityPage {
...
+  }
+
+  requesterLName(requesterLName) {
+    cy.get("input#requesterLastName").type(requesterLName);
   }
   clickSubmitOrderButton() {
-    cy.getElement(
-      ".navigationButtonsLayout [type='button']:nth-of-type(2)",
-    ).click();
+    cy.get("button.forwardButton.cds--btn--primary")
```
CalebSLane: can this be an id selector instead?

frontend/cypress/e2e/result.cy.js
```diff
@@ -32,9 +32,9 @@ describe("Result By Unit", function () {
   });
 
   it("should accept the sample, refer the sample, and save the result", function () {
+    result.acceptSample();
+    result.expandSampleDetails();
     cy.fixture("result").then((res) => {
-      result.acceptSample();
-      result.expandSampleDetails();
       result.selectTestMethod(0, res.pcrTestMethod);
       cy.get(":nth-child(3) > .cds--form-item > .cds--checkbox-label").click();
```
CalebSLane, line 39: missed selector

CalebSLane (changes requested):
PR looks like it's coming along! I particularly like the use of the test-specific id selectors (data-testid, data-cy) It would be good if we could align on which specific one we want to use and stick to it. There's a few selectors that are still page-setup/class specific that I would like moved to ids if possible. And in general, updating function names to something that captures the updated functionality should always happen.

Individual comments are below. Great work!

frontend/cypress/pages/WorkPlan.js
Bahati308: surely, lemme adjust accordingly. thanks

frontend/cypress/pages/ResultsPage.js
Bahati308: true, will harden it

PR #1478, opened 2025-01-31
Some Changes done in the  hapi_server.xml 

SOHAMPAL23:
# Pull Requests Requirements

- There was problem related to ports which was told earlier .So raising a PR regarding it .
- Server ports has been changed from  -1 to 8005

## Summary

- Server Ports  changed from -1 to  8005

## Screenshots

image


mozzy11 (changes requested):
Thanks @SOHAMPAL23 .
But i gues this change is done by a docker volume. can you add this file to giginore instead ??

volume/tomcat/hapi_server.xml
```diff
@@ -19,7 +19,7 @@
      define subcomponents such as "Valves" at this level.
      Documentation at /docs/config/server.html
  -->
-<Server port="-1" shutdown="SHUTDOWN">
+<Server port="8005" shutdown="SHUTDOWN">
```
mozzy11, line 22: you can discard this change and instead add the file to gitignore

SOHAMPAL23: Yeah ! sure

PR #1493, opened 2025-02-02
Updated dashboard E2E

Bahati308:
# Updated dashboard E2E

## Summary
Audit Existing Cypress Tests for Selector Issues

## Screenshots
Screenshot from 2025-02-02 17-32-37

## Related  issue or mention it here if applicable]

https://uwdigi.atlassian.net/browse/OGC-28
## Other
https://uwdigi.atlassian.net/issues/OGC-31?jql=issueKey%20in%20%28OGC-28%2COGC-30%2COGC-29%2COGC-31%2COGC-32%29


mozzy11 (changes requested):
Thanks @Bahati308 .
Looks like instead of just updating the selectors , you removed much of the test cases already added.
ie adding a an order with a program like pathogy or Imunochemistry and confrming that order shows up on the dashboards

frontend/cypress/pages/DashBoard.js
```diff
@@ -1,81 +1,23 @@
 class DashBoardPage {
-  addOrder(Program) {
```
mozzy11, line 2: @Bahati308  why ddnt you just update the selctors instead of removing this entire test case.

Issue #1523, opened 2025-02-09
Manually specified Cypress test files need to be replaced with a dynamic pattern


hxrshxz:
Currently, the specPattern in the Cypress configuration manually lists individual test files. This approach is not good as it requires frequent updates when new tests are added. The configuration should be updated to dynamically include all test files inside cypress/e2e

mozzy11:
Thanks @hxrshxz .
But  our tests are not run in isolation of each other , rather they depend on a specific order of execution which is curently defined . By replacing the current file definiitons with a regex expression ,the execution order will be unpredictable and Tests will fail

PR #1524, opened 2025-02-09
Refactor: Run all Cypress test files dynamically

hxrshxz:
## References
#1523 
---
## Summary
I have Updated the Cypress configuration to dynamically include all test files inside cypress/e2e/ and its subdirectories using **/*.cy.js. This simplifies test execution by removing the need for manually specifying individual test files.

### DIFF
--- a/frontend/cypress.config.js
+++ b/frontend/cypress.config.js
@@ -9,22 +9,7 @@ module.exports = defineConfig({
   e2e: {
     setupNodeEvents(on, config) {
       // implement node event listeners here
-      config.specPattern = [
-        "cypress/e2e/login.cy.js",
-        "cypress/e2e/home.cy.js",
-        "cypress/e2e/patientEntry.cy.js",
-        "cypress/e2e/orderEntity.cy.js",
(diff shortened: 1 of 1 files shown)


mozzy11:
Thanks @hxrshxz .
I gurantee this will fail because currently , our tests are not run in isolation of each other , rather they depend on a specific order of execution which is curently defined . By replacing the current file definiitons with a regex exprssion  the execution order will be unpredictable and Tests will fail

hxrshxz:
> our tests are not run in isolation of each other , rather they depend on a specific order of execution which is curently defined . By replacing the current file definiitons with a regex exprssion the execution order will be unpredictable and Tests will fail

@mozzy11 thanks for providing this little but very helpful detail we can add this to the documentation as well? also this explains why when i was running some tests individually they were failing

mozzy11:
@hxrshxz yes exactly. probably we need to break that inter dependence

hxrshxz:
> @hxrshxz yes exactly. probably we need to break that inter dependence

@mozzy11 i would like to work on this can I create a new issue For the same? and could you assign it to me?

Issue #1508, opened 2025-02-05
Add Password Expiry Notification for a System User


mozzy11:
This PR https://github.com/I-TECH-UW/OpenELIS-Global-2/pull/332 attempted to add a User Password Expiry Notification but in the old JSP UI.
Now that we have A Notifcation Framepwork , see example of how a new [User Notification is created](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop/src/main/java/org/openelisglobal/sample/controller/rest/SamplePatientEntryRestController.java#L304-L309). 

The User Notification should be created every time a user Logins in with in 72 Hours of Pawword Expiry with a meassage 
ie `12/01/2022 12:00 AM  Your Passoword will expire with in 72 Hours. Please set a new Password`

AdarshSamani:
Can i work on this @mozzy11 ?

mozzy11:
Thanks @HuzaifaOmar for working on this

PR #1587, opened 2025-02-20
Issue #1508: add password expiry notification on session initialization

HuzaifaOmar:
## Screenshots
image

## Related Issue
#1508

### DIFF
--- a/src/main/java/org/openelisglobal/security/login/CustomFormAuthenticationSuccessHandler.java
+++ b/src/main/java/org/openelisglobal/security/login/CustomFormAuthenticationSuccessHandler.java
@@ -2,6 +2,7 @@ package org.openelisglobal.security.login;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.time.OffsetDateTime;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openelisglobal/login/controller/LoginPageController.java
```diff
@@ -136,6 +145,17 @@ public UserSession getSesssionDetails(HttpServletRequest request, CsrfToken toke
         session.setSessionId(request.getSession().getId());
         if (authenticated) {
             SystemUser user = systemUserService.get(getSysUserId(request));
+            LoginUser loginUser = loginService.getUserProfile(user.getLoginName());
```
mozzy11: This endpoint is called very often during a single user session .
We would want to create the Notification only once during Login [here](https://github.com/I-TECH-UW/OpenELIS-Global-2/blob/develop/src/main/java/org/openelisglobal/security/login/CustomFormAuthenticationSuccessHandler.java#L136).

HuzaifaOmar: Thanks for the feedback, @mozzy11. I'll refactor the logic to trigger the notification only once during login, moving it to the CustomFormAuthenticationSuccessHandler as suggested. Appreciate the guidance!

mozzy11:
Thanks @HuzaifaOmar

PR #1623, opened 2025-02-28
login with only correct credentials

Bahati308:
# login with only correct credentials

## Summary
Initially, all tests had to go through a series of all users in the users' list taking ling for tests to run
## Screenshots

## Related Issue

## Other


frontend/cypress.config.js
```diff
@@ -8,29 +8,14 @@ module.exports = defineConfig({
...
-      ];
+      // Listeners
       return config;
     },
     baseUrl: "https://localhost",
     testIsolation: false,
     env: {
       STARTUP_WAIT_MILLISECONDS: 300000,
     },
+    specPattern: "cypress/e2e/**/*.cy.js", // Automatically includes all test specs
```
mozzy11: This will definately fail as For our use case, the tests are not run in isolattion of each other, so Cypress has to depend on a pre-defined order of execution

frontend/cypress/pages/LoginPage.js
```diff
@@ -81,8 +81,8 @@ class LoginPage {
     cy.url().then((url) => {
       if (url.includes("/login")) {
         cy.contains("button", "Login", { timeout: 10000 }).should("be.visible");
-        this.enterUsername(this.testProperties.getUsername());
-        this.enterPassword(this.testProperties.getPassword());
```
mozzy11: Hello @Bahati308 .
What issue where you trying to solve ??

Bahati308: with the previous code, during login there are iterates through the list of users which takes longer for all the tests

Issue #1465, opened 2025-01-29
Replace window.location.href with useNavigate() in PatientHistory.js to Avoid Hard Refresh


BhanuPrakash2047:
Problem Statement:
Currently, the application uses window.location.href for navigation in the PatientHistory.js component, which causes a hard refresh (full page reload). This approach is inefficient and disrupts the Single Page Application (SPA) behavior, leading to a poor user experience.

Location to Update:

OpenELIS-Global-2/frontend/src/components/patient/PatientHistory.js
After redirection, the application uses the component: frontend/src/components/patient/resultsViewer/results-viewer.tsx

Proposed Solution:

Replace window.location.href with the useNavigate() hook provided by React Router v6. This will enable client-side navigation without triggering a hard refresh.

Advantages:

=>Avoid Hard Refresh: The application will no longer reload the entire page during navigation, improving performance and user experience.
=>SPA Behavior: Maintains the SPA architecture, ensuring a seamless and faster navigation experience.
=>No Side Effects: This change is isolated and does not impact other components or functionality.

mozzy11:
Thanks @BhanuPrakash2047 .
Can you work on this ??

BhanuPrakash2047:
Yes i can work this issue.

mozzy11:
you can also look into other places where this isbeing used in the source code

BhanuPrakash2047:
sure, i will look into remaining places and i will edit them in source code.

Lemeri123:
Hello @BhanuPrakash2047 are you still on this?

Lemeri123:
Hello @mozzy11 should I open a PR for this or its still not necessary??
Before tweaking the changes

https://github.com/user-attachments/assets/e08c312a-fdfd-413a-a93d-d52c3768f941

After tweaking the changes

https://github.com/user-attachments/assets/510f3bc1-43e1-4b5f-ad86-40f2febc995d
I see a good user experience change

Acesif:
The current project uses React Router v5.2.0, but the requirement mentions `useNavigate()` from v6. We can look `useHistory()` from v5

Agaba-derrick:
I'm closing this issue, for now, we can always open it incase we want to work on it

PR #1640, opened 2025-03-05
Revised PatientHistory.js to use `useHistory()`

Lemeri123:
As per issue #1465 we can not use useNavigate() since we are using react-router-dom v5 and useNavigate was introduced in v6 hence why I replaced window.href.loaction with useHistory. 
useHistory just like useNavigate doesnt hard refresh


mozzy11:
Thanks @Lemeri123 . can you search other places where we use Window.location and also replace those ie [here](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/frontend/src/App.js#L102) ??

Lemeri123:
@mozzy11

mozzy11:
@Lemeri123 can create a screen recording of all the all the changed files to show everything works fine ??

Lemeri123:
[Screencast from 2025-03-25 10-08-20.webm](https://github.com/user-attachments/assets/86db4e38-0fa7-4e7e-b22d-d6970120b86f)

mozzy11:
Thanks @Lemeri123 but your video does seem be showing the actual effect of your changes

Lemeri123:
Pardon @mozzy11

mozzy11:
Thanks @Lemeri123 
Currently not needed. you can try another issue

Issue #1619, opened 2025-02-27
refactoring cypress.config.js to track specs automatically instead of adding them manually


Bahati308:
with the current code shown  [here](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/frontend/cypress.config.js), after working on a test, one adds it manually on the list. Should a name be misspelt, or a letter missed, the test will not be captured by cypress

SOHAMPAL23:
@Bahati308 

 We can use Glob -pattern  Instead  which would automatically include all test files in a directory.

Bahati308:
> [@Bahati308](https://github.com/Bahati308)
> 
> We can use Glob -pattern Instead which would automatically include all test files in a directory.

indeed we can

SOHAMPAL23:
So , Are you working on this issue or Can I put a PR on this issue

mozzy11:
The challenge is currently tests are supposed to be ran in a specific order since they depend on each other

mozzy11:
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/1658

SOHAMPAL23:
@mozzy11 So how should I go further in this 
I have create a sequential pattern so then it will work

SOHAMPAL23:
If we add new files so how it will get serialised? Can you guide me in that part so that I can understand where I am lagging

Issue #971, opened 2024-04-06
Enhancement: Making Cypress More Developer Friendly


Ankit-Matth:
# Description

Currently, Cypress lacks **auto-suggestion functionality** while writing test cases, leading to a less efficient development experience. Additionally, a significant issue arises when a minor change is made in a test case, **triggering the automatic execution** of all test cases, which disrupts workflow and consumes unnecessary resources. This issue aims to address these shortcomings and enhance the developer experience with Cypress.

## A screen recording illustrating this issue is provided below:

https://github.com/I-TECH-UW/OpenELIS-Global-2/assets/146843890/d0adb88f-c1a5-40e0-bfbb-ce5a47fe15c8

IamLRBA:
@Ankit-Matth , if this issue is still open I'd like it to be assigned to me please.
Thanks!

mozzy11:
Any one to take this on ??

hxrshxz:
@mozzy11 can i work on this?

PR #1653, opened 2025-03-07
Refactored the Cypress file with spec-pattern 

SOHAMPAL23:
# Pull Requests Requirements

- Updated specPattern in cypress.config.js to use a glob pattern that automatically includes all test files located in cypress/e2e/ directory.

- Eliminated the need for manual tracking of test files
- Reduces configuration overhead when adding new tests.
- Potentially less human error in missing test files during configuration updates.

## Summary

Refactored cypress.config.js to track spec files dynamically instead of manual entry. This improves maintainability and ensures that all relevant test files are included without modifying configuration every time new tests are added.

## Related Issue

** Both the issue are same . The Pr solve the problem for manually updating the file  ** 

https://github.com/DIGI-UW/OpenELIS-Global-2/issues/971
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/1619


mozzy11:
This will fail because the test cases depend on a specific order of execution so they cant run in isolation . see issue https://github.com/DIGI-UW/OpenELIS-Global-2/issues/1658

SOHAMPAL23:
@mozzy11 I have corrected the sequential movement of the test case and checks have passed 

Thanks for the help 

Can give a review on this

Issue #1594, opened 2025-02-22
Phone Number Field Allows Non-Numeric Input in Patient Form


harshitg927:
# How to Reproduce

1. Login 
2. Go to patient ---> Add/Edit patient
3. Click on New Patient 
4. Try entering alphabets in the phone number field

# Current Behaviour

The User is allowed to input alphabets and spaces in the phone number field

# Expected Behaviour 

The User should only be allowed to input numeric characters in the phone number field 

# Screen Recording

https://github.com/user-attachments/assets/394a16d9-cd77-423b-b660-334521682544

harshitg927:
I am working on this issue and will raise a pull request for it once done

peach280:
I saw the phone number validation on old UI but it was saving the details even though I entered the wrong phone format.

https://github.com/user-attachments/assets/13b6f19d-ad9c-4f7e-86f1-3fa4a8206604

So I tried different validation on new UI through my pull request. Please guide me if I am wrong

lokaesshwar:
@mozzy11  Hi Mam, if this issue is still open and unassigned, could you please assign it to me? My estimated time to complete the task is within 2 days

Lemeri123:
Hello @harshitg927, @blitzboah how far per this issue??

harshitg927:
Hello @lokaesshwar @Lemeri123 I did raise a pull request for this issue

PR #1720, opened 2025-03-19
Issue 1594 - Enhance Phone Number Input Handling with State Management and Validation

blitzboah:
# Pull Requests Requirements

      Issue number if applicable.
- [x] The changes confirm to the OpenElis Global x3 Styleguide and design
      documentation.
- [x] I have read and agree to the Contributing Guidelines of this project.

## Summary
This PR improves the phone number input handling by:

- Using React state to manage the phone number value.
- Allowing only valid characters (+, numbers, and -).

## Related Issue

#1594


frontend/src/components/patient/CreatePatientForm.js
```diff
@@ -614,7 +644,7 @@ function CreatePatientForm(props) {
                 <Field name="primaryPhone">
                   {({ field }) => (
                     <TextInput
-                      value={values.primaryPhone || ""}
+                      value={phoneNumber}
```
mozzy11: This doesnt set a phone values for an existing patient

frontend/src/components/patient/CreatePatientForm.js
```diff
@@ -222,6 +223,35 @@ function CreatePatientForm(props) {
...
+      // we can adjust threshold as needed
+      try {
+        const escapedPhoneFormat = configurationProperties.PHONE_FORMAT.replace(
+          /[.*+?^${}()|[\]\\]/g,
+          "\\$&",
+        );
+        const phoneRegex = new RegExp(escapedPhoneFormat);
+
+        if (!phoneRegex.test(value)) {
+          console.warn("Invalid phone number format");
```
mozzy11: There should be a visible message on the UI not just in the console

frontend/src/components/patient/CreatePatientForm.js
```diff
@@ -222,6 +223,35 @@ function CreatePatientForm(props) {
...
+
+    if (value.startsWith("+")) {
+      value = "+" + value.slice(1).replace(/[^0-9-]/g, "");
+    } else {
+      value = value.replace(/[^0-9-]/g, "");
+    }
+
+    setPhoneNumber(value);
+
+    if (configurationProperties.PHONE_FORMAT && value.length > 10) {
```
mozzy11: why would you hard code a specific lenght

frontend/src/components/patient/CreatePatientForm.js
blitzboah: does alert works?

frontend/src/components/patient/CreatePatientForm.js
blitzboah: the length check was initially added to prevent premature validation while the user was still typing

mozzy11: On the old UI , there was an [ajax call](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/src/main/resources/static/scripts/ajaxCalls.js#L390-L405) to validate the phone numbers. see where [its used ](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/ac7e9c4184f3cf5db63676a0b463da8ec5ea3c4f/src/main/webapp/pages/patient/patientManagement.jsp#L864-L866)

blitzboah: I checked the old UI and found the Ajax call for phone number validation, but I couldn't get it to work in the latest version. It seems like the Ajax validation might not be in use anymore. For now, I have ensured that only numeric input is allowed in the phone field. Let me know if there's a specific API or validation method I should use instead!

blitzboah: @mozzy11  I noticed that the phone number validation API returns true only when the phone number is encoded (e.g., %2B225-12-23-56-23) but returns false for the same number when sent as +225-12-23-56-23 or without + as well

here's the controller [`CommonValidationsRestController.java`](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/src/main/java/org/openelisglobal/common/rest/provider/CommonValidationsRestController.java#L170).
.

image
image

For handling this on the frontend, should I:

1. Call the API on form submission and show a notification if the phone number is invalid?
2. Validate using regex on the frontend only and display an error message directly below the phone number field?

blitzboah:
@mozzy11, I've implemented changes so that on form submission, the API will validate the phone number. The backend service expects the + to be URL-encoded and the number to follow the format 225-xx-xx-xx-xx, as mentioned in my previous response.

Here are some screenshots for reference:

image
image

let me know if you're good with these changes, or I can roll back to the previous implementation, which only accepted numeric input.

Thanks!

PR #1755, opened 2025-03-22
Create testmanagement.cy.js

Saarim02:
This PR improves the Cypress E2E tests for TestManagementPage by:

-- Replacing aria-label selectors with data-cy attributes for better test stability.

-- Fixing breadcrumb key typos (breadcrums.admin.managment → breadcrumbs.admin.management).

-- Improving assertions to check element visibility and prevent flaky tests.

-- Handling "No Data Available" scenarios with additional validation.

Above are the some suggestions i made in @SOHAMPAL23 PR.


Bahati308 (commented):
thanks for the work done @Saarim02 , you could improve on the following;
1. Have the TestManagementPage included in the workflow, yu seem to be accessing its methods yet it is not included 
2. You could separate the selectors from the test file for maintanability.
3. The stable selectors being used here need to be added in the corresponding js file handling that component

mozzy11:
can you npm run format to format your frontend changes ??

PR #1766, opened 2025-03-24
Update PathRoute.js

Saarim02:
Replace window.location.href with useHistory to prevent hard refresh (This file was missed in Pr of issue #1640)


frontend/src/components/utils/PathRoute.js
```diff
@@ -1,9 +1,9 @@
+import { useHistory } from "react-router-dom";
+
 const PathRoute = ({ path, children }) => {
-  var fullPath = window.location.href;
-  if (window.location.href.includes("?")) {
-    // Remove Query Params from the path
-    fullPath = window.location.href.split("?")[0];
-  }
+  const history = useHistory();
+  let fullPath = history.location.pathname;
```
mozzy11, line 5: This wont work as expected. its because `history.location.pathname` does not include the hash (#) portion of the URL. In React Router, the pathname only contains the part of the URL before the #

PR #1870, opened 2025-04-07
Improved error messages

akshayareddyp18:
Enhanced error handling, added user-friendly alerts, and improved API failure messages.
-Added user-friendly alerts to inform users about issues when fetching test sections.
-Improved error messages in the console for better debugging.
-Prevented the UI from displaying incomplete data when the API request fails.

Issues in the previous approach:
-Only a basic error message was logged to the console.
-Users had no indication that an error occurred.
-If the API failed, the dropdown could appear empty or broken without explanation.

Improvements in the new approach:
-More detailed error messages in the console for better debugging.
-User-friendly alert informs users when the API fails, avoiding confusion.
-Fallback state handling ensures the UI remains functional even if the API request fails.

### DIFF
--- a/volume/tomcat/hapi_server.xml
+++ b/volume/tomcat/hapi_server.xml
@@ -19,17 +19,17 @@
      define subcomponents such as "Valves" at this level.
      Documentation at /docs/config/server.html
  -->
-<Server port="-1" shutdown="SHUTDOWN">
-  <Listener className="org.apache.catalina.startup.VersionLoggerListener" />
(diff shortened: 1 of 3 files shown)


volume/tomcat/hapi_server.xml
```diff
@@ -19,17 +19,17 @@
      define subcomponents such as "Valves" at this level.
```
adityadeshlahre, line 19: discard this changes { whole file }!

PR #1892, opened 2025-04-10
add CityStateZipService integration test

josephbate:
# Pull Requests Requirements

      Issue number if applicable.
- [ ] The changes confirm to the OpenElis Global x3 Styleguide and design
      documentation.
- [ ] I have read and agree to the Contributing Guidelines of this project.

## Summary

- created a primary key for `city_state_zip` and set it as `id` to avoid the `org.dbunit.dataset.NoPrimaryKeyException: city_state_zip`
- rearranged the dummy data and allocate unique Id for each entry so as to have no duplicates
- corrected the sql syntax error in the `getCityStateZipByCityAndZipCode()` of the `CityStateZipDAOImpl()`
- corrected the `getZipCode()` method of the `CityStateZipDAOImpl()` to set the required parameter into the correct field name so as to call the right field when invoked 
- created a HQL query to directly use the `createQuery()` method to define the query to solve the `org.hibernate.QueryException: Named query not known` in the `getCountyCodeByStateAndZipCode()` method

## Screenshots

## Related Issue

## Other


Isabirye1515:
thanks @josephbate ,
one man army, challenging the original copy

src/test/resources/postgre-db-init/OpenELIS-Global.sql
```diff
@@ -8987,6 +8987,111 @@ SELECT pg_catalog.setval('clinlims.city_seq', 1, false);
 --
 
 COPY clinlims.city_state_zip (id, city, state, zip_code, county_fips, county, region_id, region, state_fips, state_name, lastupdated) FROM stdin;
+76708	BROADWAY	WA	98102	33	KING	\N	\N	53	WASHINGTON          	2006-12-04 12:46:48
```
mozzy11, line 8990: was there any reason for making changes to this sql dump file ??

volume/database/dbInit/OpenELIS-Global.sql
```diff
@@ -9007,112 +9113,7 @@ COPY clinlims.city_state_zip (id, city, state, zip_code, county_fips, county, re
 78724	TUKWILA	WA	98108	33	KING	\N	\N	53	WASHINGTON          	2006-12-04 12:46:48
 78726	SEATTLE	WA	98109	33	KING	\N	\N	53	WASHINGTON          	2006-12-04 12:46:48
 78706	SEATTLE	WA	98101	33	KING	\N	\N	53	WASHINGTON          	2006-12-04 12:46:48
-78721	BALLARD	WA	98107	33	KING	\N	\N	53	WASHINGTON          	2006-12-04 12:46:48
```
mozzy11, line 9010: was there any reason for making changes to this sql dump file ??

src/main/java/org/openelisglobal/citystatezip/daoimpl/CityStateZipDAOImpl.java
```diff
@@ -604,9 +604,9 @@ private CityStateZip setCityStateZipValueholder(CityStateZip csz, String city, S
     public String getCountyCodeByStateAndZipCode(CityStateZip cityStateZip) throws LIMSRuntimeException {
         String countyCode = null;
         try {
+            String hql = "SELECT csz.countyFips FROM CityStateZip csz WHERE upper(csz.state) = :param AND csz.zipCode = :param2";
 
-            List<?> list = entityManager.unwrap(Session.class)
```
mozzy11, line 608: why did you have to modify this ??

src/test/resources/postgre-db-init/OpenELIS-Global.sql
josephbate: i set the Primary key as id this create a duplucate exception when you try to build the sytem because most of the entries in the dump files have the same id (duplicated data) yet the primary key has to be unique

src/main/java/org/openelisglobal/citystatezip/daoimpl/CityStateZipDAOImpl.java
josephbate: please take a look at the pr description it gives a brief explanation of each change madeand the blockers encountered  @mozzy11

josephbate: maybe you can advise accordingly

volume/database/dbInit/OpenELIS-Global.sql
josephbate: again if you take a look at the dump files most of these entries are duplicated, if am not mistaken 3times so the ids (which are now primar keys) are required to be unique hence the changes

src/main/java/org/openelisglobal/citystatezip/daoimpl/CityStateZipDAOImpl.java
HerbertYiga: i notice you are changing the Data access layer yet your pr is mainly for tests,are you able to write tests for the existing methods without changing them

src/main/java/org/openelisglobal/citystatezip/daoimpl/CityStateZipDAOImpl.java
```diff
@@ -179,7 +179,7 @@ public List<CityStateZip> getAllStateCodes() throws LIMSRuntimeException {
     public CityStateZip getCityStateZipByCityAndZipCode(CityStateZip cityStateZip) throws LIMSRuntimeException {
         CityStateZip csz = null;
         try {
-            String sql = "from CityStateZip csz where upper(csz.city)) = :param and csz.zipCode = :param2";
+            String sql = "from CityStateZip csz where upper(csz.city) = :param and csz.zipCode = :param2";
```
josephbate, line 182: @HerbertYiga i believe i  this case you cannot write test against this method and get the desired result becuase the there is a sql  syntax error hence the minor correction

src/main/java/org/openelisglobal/citystatezip/daoimpl/CityStateZipDAOImpl.java
```diff
@@ -289,7 +289,7 @@ public CityStateZip getZipCode(CityStateZip cityStateZip) throws LIMSRuntimeExce
                 // give the cityStateZip object an artificial id needed for
                 // autocomplete but not stored anywhere
                 cszip.setId(String.valueOf(i));
-                cszip.setCity(zipCode);
+                cszip.setZipCode(zipCode);
```
josephbate, line 292: @HerbertYiga 
for this one i also believe it would be a bit confusing to write a test that fetches a ZipCode from the data base and  then set it in the city Field
in that the test will call `object.getCity()` to confirm if a zipcode is correct.that can be confussion

src/main/java/org/openelisglobal/citystatezip/daoimpl/CityStateZipDAOImpl.java
josephbate: @HerbertYiga 
as for this this change i dont really think i could write a test against a method because the error `org.hibernate.QueryException: Named query not known` i encountered was due to fact that the named query `"cityStateZip.getCountyCodeByStateAndZipCode"` is not defined anywhere in your codebase or mapped entity.
here there two option either to define the query so as to be used or use a regular query instead so i decided to use a regular hql query to fetch the desired out come 

@HerbertYiga @mozzy11 please do advise accordingly

Issue #1988, opened 2025-05-12
GSoC 2025 : Rewrite Test Management Page & Components from JSP to ReactJS


adityadeshlahre:
**Project Title:** Rewriting Test Management Pages and Components From JSP to ReactJS

**Project Description:** Most of the test management pages and their Functionalities are still on the OLD JSP framework. This Issue/Project aims to rewrite the required test management pages and their components from JSP to ReactJS.

**Milestone:** Complete the possible page with the workflow and functionalities migrated to the new React UI

**Mentor:** @mozzy11 

**NOTE:** This issue serves as the main tracker for the project, where all related Issues & Pull requests will be linked.

**UPDATES:** I will also be sharing regular weekly updates in the thread - [GSoC 2025 Rewriting Test Management pages and components from JSP to React](https://talk.openelis-global.org/t/gsoc-2025-rewrite-test-management-components-pages-from-jsp-to-reactjs/1545)

Thank You 🙂 
@adityadeshlahre

PR #2081, opened 2025-06-26
fix: page skipping logic for freeText as resultType

adityadeshlahre:
- referencing #1988 

- fixing the page skip logic


adityadeshlahre:
### Workflow video : ) [ wait 5 - 10 second let the video load ]

testAddComplexPage

PR #2329, opened 2025-11-17
OGC-52: Add Freezer data monitoring and dashboard

mherman22:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other

### DIFF
--- a/src/main/java/org/openelisglobal/coldstorage/valueholder/FreezerReading.java
+++ b/src/main/java/org/openelisglobal/coldstorage/valueholder/FreezerReading.java
@@ -1,0 +1,62 @@
+package org.openelisglobal.coldstorage.valueholder;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EnumType;
+import jakarta.persistence.Enumerated;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.Index;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.SequenceGenerator;
+import jakarta.persistence.Table;
+import java.math.BigDecimal;
+import java.time.OffsetDateTime;
+import lombok.Getter;
+import lombok.Setter;
+import org.openelisglobal.common.valueholder.BaseObject;
+
+@Getter
+@Setter
+@Entity
+@Table(name = "freezer_reading", indexes = {
+        @Index(name = "idx_freezer_reading_name_time", columnList = "freezer_id, recorded_at"),
+        @Index(name = "idx_freezer_reading_status", columnList = "status") })
+public class FreezerReading extends BaseObject<Long> {
+
+    public enum Status {
+        NORMAL, WARNING, CRITICAL
+    }
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "freezer_reading_generator")
+    @SequenceGenerator(name = "freezer_reading_generator", sequenceName = "freezer_reading_seq", allocationSize = 1)
+    @Column(name = "id")
+    private Long id;
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "freezer_id", nullable = false)
+    private Freezer freezer;
+
+    @Column(name = "recorded_at", nullable = false)
+    private OffsetDateTime recordedAt;
+
+    @Column(name = "temperature_celsius")
+    private BigDecimal temperatureCelsius;
+
+    @Column(name = "humidity_percentage")
+    private BigDecimal humidityPercentage;
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "status", length = 16, nullable = false)
+    private Status status = Status.NORMAL;
+
+    @Column(name = "transmission_ok", nullable = false)
+    private Boolean transmissionOk = Boolean.TRUE;
+
+    @Column(name = "error_message")
+    private String errorMessage;
+}
--- a/src/main/java/org/openelisglobal/coldstorage/valueholder/ThresholdProfile.java
+++ b/src/main/java/org/openelisglobal/coldstorage/valueholder/ThresholdProfile.java
@@ -1,0 +1,110 @@
+package org.openelisglobal.coldstorage.valueholder;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToMany;
+import jakarta.persistence.SequenceGenerator;
+import jakarta.persistence.Table;
+import java.math.BigDecimal;
+import java.time.OffsetDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import lombok.Getter;
+import lombok.Setter;
+import org.openelisglobal.common.valueholder.BaseObject;
+import org.openelisglobal.systemuser.valueholder.SystemUser;
+
+@Getter
+@Setter
+@Entity
+@Table(name = "threshold_profile")
+public class ThresholdProfile extends BaseObject<Long> {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "threshold_profile_generator")
+    @SequenceGenerator(name = "threshold_profile_generator", sequenceName = "threshold_profile_seq", allocationSize = 1)
+    @Column(name = "id")
+    private Long id;
+
+    @Column(name = "name", nullable = false, unique = true, length = 128)
+    private String name;
+
+    @Column(name = "description")
+    private String description;
+
+    @Column(name = "warning_min")
+    private BigDecimal warningMin;
+
+    @Column(name = "warning_max")
+    private BigDecimal warningMax;
+
+    @Column(name = "critical_min")
+    private BigDecimal criticalMin;
+
+    @Column(name = "critical_max")
+    private BigDecimal criticalMax;
+
+    @Column(name = "min_excursion_minutes")
+    private Integer minExcursionMinutes = 5;
+
+    @Column(name = "max_duration_minutes")
+    private Integer maxDurationMinutes;
+
+    @Column(name = "humidity_warning_min")
+    private BigDecimal humidityWarningMin;
+
+    @Column(name = "humidity_warning_max")
+    private BigDecimal humidityWarningMax;
+
+    @Column(name = "humidity_critical_min")
+    private BigDecimal humidityCriticalMin;
+
+    @Column(name = "humidity_critical_max")
+    private BigDecimal humidityCriticalMax;
+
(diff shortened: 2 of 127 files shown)


src/main/java/org/openelisglobal/freezermonitoring/config/FreezerMonitoringSchedulingConfiguration.java
```diff
@@ -0,0 +1,17 @@
...
+ * feature.
+ *
+ * Note: Loads unconditionally to support dynamic enable/disable via admin UI.
+ * Individual @Scheduled services check feature flag at runtime before
+ * executing.
+ */
+@Configuration
+@EnableScheduling
+public class FreezerMonitoringSchedulingConfiguration {
+}
```
ibacher: Do we actually need this? It seems like Spring Scheduling is already enabled.

src/main/java/org/openelisglobal/freezermonitoring/controller/FreezerMonitoringController.java
```diff
@@ -0,0 +1,597 @@
...
+/**
+ * REST controller for cold storage monitoring endpoints.
+ *
+ * Note: Loads unconditionally. Accessible to all authenticated users. Feature
+ * can be enabled/disabled via admin UI without restart.
+ */
+@Validated
+@RestController
+@RequestMapping("/rest/coldstorage")
+public class FreezerMonitoringController extends BaseRestController {
```
ibacher: This seems have too many responsibilities in a single file. Can we split this out into multiple controllers?

src/main/java/org/openelisglobal/freezermonitoring/dao/AlertActionDAO.java
```diff
@@ -0,0 +1,17 @@
+package org.openelisglobal.freezermonitoring.dao;
+
+import java.time.OffsetDateTime;
+import java.util.List;
+import org.openelisglobal.common.dao.BaseDAO;
+import org.openelisglobal.freezermonitoring.entity.AlertAction;
+
+public interface AlertActionDAO extends BaseDAO<AlertAction, Long> {
```
ibacher: All this alert stuff should live in a top-level alert domain.

src/main/java/org/openelisglobal/freezermonitoring/dao/AuditLogEntryDAO.java
```diff
@@ -0,0 +1,10 @@
+package org.openelisglobal.freezermonitoring.dao;
+
+import java.util.List;
+import org.openelisglobal.common.dao.BaseDAO;
+import org.openelisglobal.freezermonitoring.entity.AuditLogEntry;
+
+public interface AuditLogEntryDAO extends BaseDAO<AuditLogEntry, Long> {
```
ibacher: Shouldn't this be integrated with the existing audit log stuff?

src/main/java/org/openelisglobal/freezermonitoring/dao/ReportScheduleDAO.java
```diff
@@ -0,0 +1,10 @@
+package org.openelisglobal.freezermonitoring.dao;
+
+import java.util.List;
+import org.openelisglobal.common.dao.BaseDAO;
+import org.openelisglobal.freezermonitoring.entity.ReportSchedule;
+
+public interface ReportScheduleDAO extends BaseDAO<ReportSchedule, Long> {
```
ibacher: Why aren't we using the existing reporting stuff?

src/main/java/org/openelisglobal/freezermonitoring/controller/FreezerMonitoringController.java
mherman22: Working on the fix!

src/main/java/org/openelisglobal/freezermonitoring/dao/AlertActionDAO.java
mherman22: top-level, do you mean outside the freezermonitoring stuff?

ibacher: Yes. Alerts of some sort are a pretty common feature. Although, I suppose the data model here has some freezer-specific things. But maybe we can attach the data to an alert in some other way...

ibacher: Basically, rather than writing an alert-subsystem-per-system, it would be great if we had a single alert-system and could slot these notifications into that.

freezer_monitoring_cleanup.sql
ibacher, line 1: This feels like its either in the wrong place or shouldn't exist at all. It definitely shouldn't be in the root.

freezer_monitoring_demo_data.sql
ibacher, line 1: Same with this one...

src/main/java/org/openelisglobal/reports/action/implementation/FreezerExcursionReport.java
```diff
@@ -0,0 +1,267 @@
...
+        reportItems = new ArrayList<>();
+
+        if (GenericValidator.isBlankOrNull(startDate) || GenericValidator.isBlankOrNull(endDate)) {
+            add1LineErrorMessage("report.error.message.noPrintableItems");
+            return;
+        }
+
+        try {
+            OffsetDateTime start = OffsetDateTime.parse(startDate + "T00:00:00Z");
+            OffsetDateTime end = OffsetDateTime.parse(endDate + "T23:59:59Z");
```
ibacher: Shouldn't the timezone information be set based on the local timezone so that the report is for `n` days in the local timezone rather than UTC?

src/main/java/org/openelisglobal/hibernate/type/JsonBinaryType.java
```diff
@@ -0,0 +1,108 @@
...
+import java.sql.SQLException;
+import java.sql.Types;
+import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.usertype.UserType;
+
+/**
+ * Custom Hibernate UserType for mapping PostgreSQL JSONB columns to String
+ * fields. This handles the type conversion between Java String and PostgreSQL
+ * JSONB. Uses reflection to avoid compile-time dependency on PostgreSQL driver.
```
ibacher: Why does this use reflection to avoid a compile-time dependency? The PG driver is part of the source code for this, so it would be better not to do that.

src/main/java/org/openelisglobal/coldstorage/valueholder/ThresholdProfile.java
```diff
@@ -0,0 +1,79 @@
...
+    @Column(name = "id")
+    private Long id;
+
+    @Column(name = "name", nullable = false, unique = true, length = 128)
+    private String name;
+
+    @Column(name = "description")
+    private String description;
+
+    @Column(name = "warning_min", precision = 6, scale = 2)
```
ibacher: Is this precision value coming from a specific requirement?

src/main/java/org/openelisglobal/notification/dao/AnalysisNotificationConfigDAOImpl.java
```diff
@@ -12,6 +12,7 @@
 import org.springframework.stereotype.Component;
 
 @Component
+@SuppressWarnings("unused")
```
ibacher: Removing this kind of refactoring in an otherwise untouched class could go a long way towards making this PR more reviewable.

src/main/java/org/openelisglobal/notification/valueholder/NotificationConfigOption.java
```diff
@@ -28,7 +28,10 @@ public class NotificationConfigOption extends BaseObject<Integer> {
     private static final long serialVersionUID = -6242849348547228319L;
 
     public enum NotificationNature {
-        RESULT_VALIDATION
+        RESULT_VALIDATION, FREEZER_TEMPERATURE_ALERT, // Task T039: Alert system integration
```
ibacher, line 31: I don't think we need these comments long-term

```suggestion
        RESULT_VALIDATION, 
        FREEZER_TEMPERATURE_ALERT, // Task T039: Alert system integration
```

Also what is a "GENERAL_ALERT"? Is that something we specifically need?

src/main/java/org/openelisglobal/coldstorage/valueholder/FreezerReportSchedule.java
```diff
@@ -0,0 +1,169 @@
...
+    @OneToOne
+    @JoinColumn(name = "cron_scheduler_id", referencedColumnName = "id")
+    private CronScheduler cronScheduler;
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "report_type", nullable = false, length = 50)
+    private ReportType reportType;
+
+    @Column(name = "freezer_id")
+    private Long freezerId;
```
ibacher: Shouldn't this be a link to the actual freezer?

src/main/java/org/openelisglobal/coldstorage/valueholder/FreezerReportSchedule.java
```diff
@@ -0,0 +1,169 @@
...
+import jakarta.persistence.OneToOne;
+import jakarta.persistence.SequenceGenerator;
+import jakarta.persistence.Table;
+import java.sql.Timestamp;
+import org.openelisglobal.common.valueholder.BaseObject;
+import org.openelisglobal.scheduler.valueholder.CronScheduler;
+
+@Entity
+@Table(name = "freezer_report_schedule")
+public class FreezerReportSchedule extends BaseObject<Long> {
```
ibacher: I'm having a hard time understanding the point of this class. Surely the existing scheduling subsystem should be used for actual scheduling and maybe we have some class to store per-freezer stuff? I don't see other domains here adding custom "*Schedule" classes.

src/main/java/org/openelisglobal/coldstorage/valueholder/FreezerReading.java
```diff
@@ -0,0 +1,62 @@
...
+    private Long id;
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "freezer_id", nullable = false)
+    private Freezer freezer;
+
+    @Column(name = "recorded_at", nullable = false)
+    private OffsetDateTime recordedAt;
+
+    @Column(name = "temperature_celsius", precision = 6, scale = 2)
```
ibacher: Again, is the precision here defined somewhere in the requirements? This feels overly precise...

ibacher (commented):
So a few things:

1. I have no idea how to actually review this PR. Is it possible to break this down into smaller PR chunks focused on specific features?
2. I noticed in the reports that we seem to be assuming that they should be run between midnight and midnight UTC. It seems like this should be tied to the local timezone.
3. This adds SQL files to the root and those definitely shouldn't be there.

mherman22:
> So a few things:
> 
> 1. I have no idea how to actually review this PR. Is it possible to break this down into smaller PR chunks focused on specific features?
> 2. I noticed in the reports that we seem to be assuming that they should be run between midnight and midnight UTC. It seems like this should be tied to the local timezone.
> 3. This adds SQL files to the root and those definitely shouldn't be there.

let me keep th ui changes in this pull request + backend implementation of corrective actions and devices. And move the rest of the stuff like reports, alerts, into other PRs. Hope that makes life easy!

mherman22:
i am putting this in draft. i have divided it into two PRs. Feel free to have the reviews over at:
- https://github.com/DIGI-UW/OpenELIS-Global-2/pull/2359 for backend changes
- https://github.com/DIGI-UW/OpenELIS-Global-2/pull/2358 for UI changes.

cc: @ibacher

PR #2343, opened 2025-11-22
feat(results): add multi-select support for result entry (fixes #2198)

### LINKED ISSUE #2198: Add Support for Selecting Multiple Values from DropDown for Multi Select Result Type
Currently on the new Result Entry Page (https://localhost/AccessionResults)  , A user is only able to selct a single Value from the Drop Down ie same as for **D - Select list** Result Type,
Add Support for  a user to be able to select Multiple Values fron the Drop down for **M - Multi-select list** Result type  .

See Result Entry Page on the old ui (https://localhost/api/OpenELIS-Global/LogbookResults?type=)

Image

krishpatel2-prog:
## Summary
This Pull Request implements Multi-Select functionality for Result Entry, resolving Issue #2198. It upgrades the UI for "Multi-select list" (Type M) results, allowing users to select multiple dictionary values for a single test parameter.

## The Problem
Previously, tests configured with the "Multi-select list (M)" result type were incorrectly rendered using a standard single-select dropdown.
- **User Impact:** Users could only select one value. Selecting a second value would overwrite the first.
- **Technical Constraint:** The frontend treated "M" types identically to "D" (Single Select) types in the `renderCell` logic, lacking the state management to hold multiple values.

## The Solution
This fix involves significant updates to `frontend/src/components/resultPage/SearchResultForm.js` and `Style.css` to support array-based result selection while maintaining backward compatibility with the backend's string-based storage.

### 1. UI Component Architecture
- Implemented a composite UI pattern for "M" and "C" (Cascading) result types using the Carbon Design System.
- **Selection:** A `<Select>` component allows the user to pick a value from the dictionary.
- **Display:** Selected values are rendered as removable Carbon `<Tag>` components (green for Multi-select) immediately below the dropdown.
- **Styling:** Added `flex-wrap` CSS classes (`.resultOptionsDisplay`, `.resultTagDisplay`) in `Style.css` to ensure tags flow correctly on smaller screens without breaking the table layout.

### 2. State Management & Logic
- Introduced `multiSelectResultValues` state to track selected items locally as an array of objects.
- Implemented `validateResults` to handle the addition of unique items to the state (preventing duplicates).
- Implemented `handleTagClose` to allow users to deselect/remove specific values.

### 3. Data Serialization (Crucial Fix)
- The OpenELIS backend expects result values as a single comma-separated string, not a JSON array.
- **Fix:** Updated the `handleSave` function to intercept "M" type results. It maps the array of selected tag IDs and joins them (e.g., `['id1', 'id2']` -> `"id1,id2"`) before the payload is sent to the server. This ensures seamless data persistence without backend changes.

### 4. Code Quality
- Fixed existing ESLint `no-case-declarations` errors in the `renderCell` switch statement by correctly scoping `case` blocks with curly braces `{ ... }`.

## Related Issue
Fixes #2198

## Verification
* **Visuals:** The UI now correctly renders selected items as tags below the dropdown.
* **Logic:** Users can select multiple items, and removing a tag updates the state correctly.
* **Persistence:** Data is formatted as a string before submission to ensure compatibility with the existing database schema.


mozzy11:
Thanks @krishpatel2-prog for working on this ,
can you add a screen recording of this working , ie for both adding the results and fetching the results

mozzy11:
can you run `npm run format`

mozzy11:
@krishpatel2-prog can you run the formatter and also attach a video recording ??

mozzy11:
@krishpatel2-prog this will definately require a little modification on the API side to parse these values properly

krishpatel2-prog:
@mozzy11 
Thanks for the review..
 I was checking the `/AccessionResults` page in the new React UI, and I realized something important:

**The new UI does not currently have the result-entry functionality implemented at all.**  
It only has a search bar and an empty table. I couldn’t find any React code for:

- showing the tests/analyses  
- rendering result fields  
- single-select or multi-select inputs  
- saving result values  

Because of this, the multi-select feature cannot be added yet — the whole result-entry UI from the old Logbook page has not been migrated into React.

**Before I continue, I wanted to confirm:**

1. Should I build the full result-entry UI for `/AccessionResults` in React first, and then add multi-select?  
2. Or is there an existing component/branch for this that I should use?

Just need clarification so I proceed in the right direction. Thanks.

mozzy11:
@krishpatel2-prog you can modify re-use the existing component

krishpatel2-prog:
@mozzy11

I’ve investigated this end-to-end. The multi-select (M-type) functionality added in my PR (tag rendering and save handling) is implemented correctly.

However, I’m currently blocked from attaching a screen recording because, in a clean local setup, both LogbookResults and AccessionResults filter out the test data due to backend workflow/status constraints (sample and analysis eligibility). Even with structurally valid M-type data present in the database, /rest/LogbookResults does not return rows unless specific workflow states are met.

Without guidance on the required workflow statuses or access to seeded/demo data, the behavior shown in the issue screenshot cannot be reliably reproduced or validated locally.

Please advise whether:

- specific workflow states are required to test this, or
- the issue should be re-scoped or closed as environment-dependent, with the multi-select logic already addressed in the PR.

mozzy11:

PR #2349, opened 2025-11-25
OCL Import should stop Failing  Due to Missing Role Handling

### LINKED ISSUE #2346: OCL Import Fails Due to Missing Role Handling - NullPointerException in RoleModule Creation
## Summary

  The OCL import process fails with a NullPointerException when creating RoleModule objects because the required roles ("Results" and "Validation") are not found in the database, resulting in null role references.

  ### Steps to Reproduce

  1. Run OclInnitializerTest.testImportOclPackage_validZip()
  2. OCL import attempts to create panel with role modules
  3. System looks up roles using Constants.ROLE_RESULTS ("Results") and Constants.ROLE_VALIDATION ("Validation")
  4. roleService.getRoleByName() returns null for these roles
  5. Code attempts to create RoleModule with null roles
  6. RoleModuleServiceImpl.insert() calls duplicateRoleModuleExists()
  7. NullPointerException thrown at RoleModuleDAOImpl.java:166 when accessing roleModule.getRole().getId()

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
     
- In OclToOpenElisMapper we added ensureRoleExists(...), which checks for each required role and creates it (grouped under “Lab Unit Roles”) if it isn’t already in system_role. The mapper now calls this helper before building any RoleModule, so the import continues instead of passing null into the service layer.

- we added Defensive checks that  prevent crashes if something slips through (safety net)

- Tests confirm the exact failure scenario is handled (validation)

## Screenshots
Screenshot from 2025-11-25 16-37-18

## Related Issue

Closes #2346

## Other


src/test/java/org/openelisglobal/ocl/OclInnitializerTest.java
mherman22, line 1: i didnt expect you to change the test.

Agaba-derrick: @mherman22  I  did change OclInnitializerTest mainly to add the regression that proves the import auto-creates the missing roles and to drop the unused sample-type field. Those updates were deliberate so the fix is testable

if this doesnt meet our expectations, i can always  revert  to the previous layout !

src/main/java/org/openelisglobal/ocl/OclToOpenElisMapper.java
```diff
@@ -770,6 +777,32 @@ private RoleModule createRoleModule(String userId, SystemModule workplanModule,
         return roleModule;
     }
 
+    private Role ensureRoleExists(String roleName, String description, String displayKey) {
```
mozzy11, line 780: Thanks @Agaba-derrick .
I guess youre trying to fix an error thrown by the test.
Ideally at runtime , the roles are inserted by Liquibase so they exist.
Youll have to fix this for the  test itsself , because theres some kind of db cleaning durong test-time which most probably causes the error .

mozzy11:
This has been fixed in this PR. https://github.com/DIGI-UW/OpenELIS-Global-2/pull/2362. we just needed to have roles loded before this test runs

PR #2371, opened 2025-12-01
Branding

HerbertYiga:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


mozzy11:
Thanks @HerbertYiga .
Can you add a screen recording of the feature ??

HerbertYiga:
> Thanks @HerbertYiga . Can you add a screen recording of the feature ??

sure as soon as i am done with some more testing

mozzy11:
@HerbertYiga can you fix the merge conflics and also run the formatters ?

src/main/java/org/openelisglobal/sitebranding/valueholder/SiteBranding.java
```diff
@@ -0,0 +1,142 @@
...
+ * Task Reference: T011
+ */
+@Entity
+@Table(name = "site_branding")
+public class SiteBranding extends BaseObject<String> {
+
+    private static final long serialVersionUID = 1L;
+
+    @Id
+    @GeneratedValue(generator = "uuid-generator")
```
CalebSLane: uuid is not used anywhere else in the codebase for OpenELIS db ids. Is there a specific reason to use it here?

mozzy11: By the way do we even need to persist this in the DB ?
we could  just keep it as  a frontend config  file , and just mounted on the frontend container.

mozzy11:
@HerbertYiga why not just doing this entirely as a frontend config file instead of persisting it in the db ??
Creating an api for this looks like an overkill for me

src/main/java/org/openelisglobal/sitebranding/valueholder/SiteBranding.java
ibacher: @mozzy11 Because we want to let users dynamically change this in the system and it's more sane to do that in a multi-user system with a database and transactional control. In any case, we also allow users uploading files (for logos) and we need to have server-side validation that those files are correct, at which point, a database table is pretty small overhead.

ibacher:
@mozzy11 Pretty sure this feature doesn't actually work yet...

PR #2460, opened 2025-12-16
fix(test): Add shutdown hook to prevent orphaned Testcontainers containers (OGC-281)

pmanko:
Fixes OGC-281

Adds JVM shutdown hook to BaseTestConfig to ensure PostgreSQL containers are stopped on exit, preventing orphaned containers from interrupted test runs.


pmanko:
Closing after testing previously existing Ryuk cleanup already included in testcontainers. Future psql container pollution needs to be more closely investigated for possible causes - default setup should have cleanup already enabled.

PR #2463, opened 2025-12-17
feat(189):lab unit management

adityadeshlahre:
will update this soon : )


src/main/java/org/openelisglobal/labunit/controller/rest/LabUnitRestController.java
```diff
@@ -0,0 +1,881 @@
...
+    @Autowired
+    private UserRoleService userRoleService;
+
+    /**
+     * GET /api/lab-units - List all lab units with counts and optional status
+     * filter
+     */
+    @GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<?> getLabUnits(@RequestParam(required = false) String status,
+            @RequestParam(defaultValue = "1") int page, @RequestParam(defaultValue = "25") int size,
```
Isabirye1515: Why we dint use the exixting pagination mechanism

mozzy11:

Issue #2259, opened 2025-10-15
Add pre-commit hook to automatically run frontend formatting (npm run format)


Agaba-derrick:
# Add pre-commit hook to automatically run frontend formatting (`npm run format`)

## Description
Currently, developers manually run `npm run format` before committing frontend changes (JS, CSS, JSON, etc.) to ensure code style consistency. To streamline this process and prevent unformatted code from entering the repository, we should automate it using a **pre-commit hook**.

### Tasks
- [ ] Set up a pre-commit hook that runs `npm run format` before each commit.  
- [ ] Ensure the hook:
  - Executes `npm run format` automatically.  
  - Stages only  modified files after formatting.  
  - Allows the commit to proceed once formatting is complete.  
- [ ] **Recommended approach:** Use **Husky** to manage the hook for team-wide consistency:
  - Install Husky (`npm install husky --save-dev`)  
  - Add the hook: `npx husky add .husky/pre-commit "npm run format && git add ."`  
  - Commit the `.husky/pre-commit` file so the hook is versioned and automatically used by all contributors.  
- [ ] Update contributor documentation (`CONTRIBUTING.md`) with setup instructions.

### Optional Enhancement
- Use a **check-only script** (`npm run format:check`) to block commits if code is unformatted, forcing developers to fix it manually.

### Expected Outcome
- Frontend code is always consistently formatted before commits.  
- Developers no longer need to remember to run `npm run format`.  
- PRs are cleaner, with fewer formatting-related changes.  
- Team-wide enforcement ensures consistent style across the repository
- Only code that has been  worked upon should be formated

Agaba-derrick:
this is a good first issue !

mukulpythondev:
Hi , @Agaba-derrick , i found this issue interesting and happy to work on it.

Agaba-derrick:
@mukulpythondev  you have  Two Pull requests open  #2263  and #2286  all solving one Issue, please specify which Pr is updated and  close the other !
Are you able to do that ?

mukulpythondev:
@Agaba-derrick I have closed the #2263

sagarkhandagre998:
@Agaba-derrick I think it's not implemented yet ?

sagarkhandagre998:
@Agaba-derrick  we don't have anything in CONTRIBUTING.md , still should be updated with setup instruction?

sagarkhandagre998:
[...]

PR #2476, opened 2025-12-19
Implement - Add pre-commit hook to automatically run frontend formatting (npm run format) #2259

sagarkhandagre998:
Implementation- 
Add pre-commit hook to automatically format frontend code before each commit.

---

**Changes**
- Installed **Husky (v9.1.7)** as devDependency
- Added `prepare` script to auto-initialize Husky on `npm install`
- Created `.husky/pre-commit` hook: runs `cd frontend && npm run lint-staged`
- Configured lint-staged** following  pattern:
   - ESLint + Prettier for JS/JSX/TS/TSX files
   - Prettier only for CSS/SCSS/JSON/MD files
   - Only processes staged files
- Updated `CONTRIBUTING.md` with setup instructions

---

### ✅ **How It Works**
1. Developer commits code: `git commit -m "message"`
2. Pre-commit hook automatically runs `npm run lint-staged`
3. Formatted files that actually changed/created are staged automatically
4. Commit proceeds

---

### 🔧 **Setup (One-time)**
```bash
cd frontend
npm install
```

Hook is now active and will run automatically on every commit.

---

### ✅ **Tested**
- Created test file with unformatted code
- Committed successfully
- File was automatically formatted before commit


frontend/package.json
```diff
@@ -102,6 +103,7 @@
     "eslint-plugin-prettier": "^5.0.0",
```
Agaba-derrick, line 114: does this make sense in terms of formating only staged files ?
https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/package.json#L83C2-L86C5

sagarkhandagre998: Yeah definitely, It will not format all files. Format only the files that developer has changed or implemented some new files,  so after commiting (all the chagned or new files) will get automatically format so we don't need to format all. Thats why i previously thought to implement this.

Agaba-derrick: are you able to make  your changes look like those referenced  from that block i have shared ?

sagarkhandagre998: Sure I'll do that, Need Some time to analyze. I'll commit changes soon.

sagarkhandagre998: > are you able to make your changes look like those referenced from that block i have shared ?

But they only {ts,tsx} files in their particular package but in our frontend we need to include {js ,jsx ,ts ,tsx} and {css,scss,json,md} also. I'll make sure to include this as well.

Agaba-derrick (approved):

Agaba-derrick:
@sagarkhandagre998  Thanks, we are clossing this PR as this issue was resolved in some earlier work by @pmanko

Issue #2278, opened 2025-10-28
Move away from db to file 'menu configuration'


reagan-meant:
Currently the menu items are loaded from the database in the menu table. We should have a configuration in the frontend that manages this to make it simple and configurable by implementors

Lemeri123:
Hello @reagan-meant let me give this a shot

mozzy11:
We can add support for defining a config file to create an Implementation specific Menu Structure.
This is currently supported in the UI via Menu Configuration , but a config file can help pre define the menu

PatadeSujal:
Hi! I'm checking in to see if there has been any progress on this issue.  
If there is no active work happening, I would be happy to take it and start working. 🙂

mozzy11:
Thanks @PatadeSujal

Lemeri123:
@mozzy11 @reagan-meant I have a couple of questions on this............Should we completely remove the backend menu APIs, or simply stop using them in the React app while leaving them in place for the legacy UI?
How should updates to the new configuration be made? Is it intended to be a static JSON file that implementers edit manually, or should the current admin UI still be able to add/remove/toggle items (and if so, where would it persist those changes)?
Are there other consumers of the /rest/menu endpoints that we need to keep supporting?.

vidishadeswal:
Hey @mozzy11 @reagan-meant, if this issue is still available I'd like to work on this

sagarkhandagre998:
@mozzy11 Is it resolved or Available ?

sagarkhandagre998:
@reagan-meant  Menu configuration already exists via UI and DB. There is partial config-file support today, but it only filters existing menu entries. There is no mechanism to seed or create menus from a config file. Adding config-based seeding would complement the existing UI, not replace it.

sagarkhandagre998:
Pr - https://github.com/DIGI-UW/OpenELIS-Global-2/pull/2480

PR #2480, opened 2025-12-20
Implemented - ( Issue - Move away from db to file 'menu configuration' #2278 )

sagarkhandagre998:
🎯 Feature Overview
**Menu Seeding from JSON Configuration at Application Startup**

Allows OpenELIS implementations to ship with a predefined menu structure by defining menus in a JSON configuration file. Menus are automatically created in the database at application startup without requiring manual UI configuration or Liquibase changesets.

---

## 📋 What Problem Does It Solve?

**Before:**
- Implementations had to manually configure menus via UI after deployment
- Or create complex Liquibase changesets for menu creation
- Time-consuming setup for each deployment

**After:**
- Define menus once in `menu_seed.json`
- Menus automatically created on first startup
- Idempotent - safe to run on every restart (no duplicates)

---

### 1. **Configuration**
Enable in `application.properties`:
```properties
org.openelisglobal.menu.seed.enabled=true
org.openelisglobal.menu.seed.config.path=/var/lib/openelis-global/menu/menu_seed.json
```

### 2. **Menu Definition**
Create `menu_seed.json`:
```json
{
  "menus": [
    {
      "elementId": "menu_home",
      "displayKey": "banner.menu.home",
      "actionURL": "/Dashboard",
      "presentationOrder": 1,
      "isActive": true
    }
  ]
}
```

### 3. **Automatic Seeding**
- On startup, `MenuSeedService` reads the JSON config
- Creates menus that don't exist (checks by `elementId`)
- Skips existing menus (idempotent)
- Supports parent-child hierarchies
- Rebuilds menu cache automatically

---

✨ Key Features

| Feature | Description |
|---------|-------------|
| **Idempotent** | Existing menus are preserved - no duplicates created |
| **Hierarchical** | Supports parent-child menu relationships |
| **Safe** | Non-destructive - only creates, never updates/deletes |
| **Flexible** | JSON-based configuration easy to customize |
| **Optional** | Disabled by default - opt-in via properties |
| **Validated** | Comprehensive test coverage (10 tests)


sagarkhandagre998:
i will fix the test case, Need some time.

src/test/java/org/openelisglobal/menu/service/MenuSeedServiceTest.java
```diff
@@ -0,0 +1,182 @@
...
+    @Autowired
+    private MenuService menuService;
+
+    private Path tempConfigFile;
+    private List<String> createdMenuIds = new ArrayList<>();
+
+    @Before
+    public void setup() throws Exception {
+        // Create a temporary config file for testing
+        tempConfigFile = Files.createTempFile("menu_seed_test", ".json");
```
Agaba-derrick: OpenELIS would want to maintain dataset driven test take 
an example https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/src/test/java/org/openelisglobal/menu/MenuServiceTest.java

Agaba-derrick (commented):
One clarification request: the example menu_seed.json includes metadata fields such as _comment, _usage, and _notes, which are not documented as part of the supported JSON schema. Can we confirm that the menu seed parser explicitly ignores unknown top-level keys? If not, it may be safer to move that information into the README or keep it only in a .example file.

src/test/java/org/openelisglobal/menu/service/MenuSeedServiceTest.java
sagarkhandagre998: Sure I'll check and Implement.

sagarkhandagre998:
> One clarification request: the example menu_seed.json includes metadata fields such as _comment, _usage, and _notes, which are not documented as part of the supported JSON schema. Can we confirm that the menu seed parser explicitly ignores unknown top-level keys? If not, it may be safer to move that information into the README or keep it only in a .example file.

Good catch! You're right that it's better practice to keep the production `menu_seed.json` clean. we''ll recommand to use clean  `menu_seed.json`  in readme and we'll keep them only in `menu_seed.json.example`. 

Safer  - I'll  add `@JsonIgnoreProperties(ignoreUnknown = true)` to both `MenuSeedConfig` and `MenuItemDTO` classes, which explicitly tells Jackson to ignore all unknown fields. The parser has been tested and works correctly with the metadata fields. However, I agree with your concern about best practices.

sagarkhandagre998:
@Agaba-derrick @mozzy11 Can you please review it ?

Agaba-derrick (approved):

Agaba-derrick:
can we have this  branch updated

Agaba-derrick:
are you able to list the issue you are solving ?

mozzy11:

PR #2498, opened 2025-12-27
Simplify SecurityFilter by removing ineffective XSS handling

### LINKED ISSUE #2492: 🔐 Harden SecurityFilter by simplifying XSS handling and request processing
### 🔐 Security: Harden custom SecurityFilter (naive XSS handling & dead code)

**Summary**

The custom SecurityFilter contains naive XSS detection, mutates request parameters, and includes unused logic.  
This creates false security guarantees and unnecessary technical debt in a security-critical path.

The goal is incremental hardening without changing authentication or UI behavior.

---

### 1️⃣ Naive XSS detection & input mutation

Naive XSS detection

- Detects only literal `<script>` tags  
- Misses common XSS vectors  
- Mutates user input inside a security filter  

---

### 2️⃣ Dead / unused exception logic

Unused exceptions field

Exceptions never evaluated

- Values are added but never evaluated  
- Indicates incomplete or misleading security logic  

---

### 3️⃣ Silent redirect on suspected attack

Silent redirect

- No HTTP error status  
- Limited visibility and auditability  

---

### ⚠️ Why this matters

- The filter appears to provide protection but does not reliably do so  
- Developers may assume XSS is handled when it is not  
- Security logic runs outside Spring Security’s standard mechanisms  

---

### ✅ Proposed scope (safe & minimal)

- Avoid mutating request parameters  
- Remove or clarify dead exception logic  
- Improve clarity of attack handling  
- No redesign of authentication or UI behavior

maneshri00:
Summary

This PR simplifies the SecurityFilter by removing naive and misleading XSS detection logic that attempted to sanitize request input and silently redirect users. The filter now behaves as a clean pass-through, avoiding false security guarantees and improving maintainability.

Why this change is needed

The previous implementation only checked for literal <script> tags, which does not provide real XSS protection.

Request parameters were modified inside the filter, violating servlet filter best practices and risking data corruption.

Suspected attacks were handled via silent redirects, which obscured security events and complicated auditing.

Unused and dead code paths increased technical debt and developer confusion.

This logic created a false sense of security without meaningfully improving protection.

What was changed

Removed naive XSS input inspection and request mutation

Eliminated silent redirect behavior on suspected attacks

Removed unused exception-related code

Simplified the filter to a clear, auditable request pass-through

What was intentionally not changed

No UI, authentication, or authorization logic was modified

No new security mechanisms were introduced

Output encoding and framework-level protections remain the responsibility of the view and Spring Security layers

Security rationale

XSS protection is most effective when implemented through:

Proper output encoding in views

Framework-level security controls

Input sanitization at the filter level is insufficient and potentially harmful. This change aligns with OWASP and Spring Security best practices.

Related issue

Fixes #2492


mozzy11:
Thanks for taking a run at #2492 — you're right that the current detection is naive and the exceptions list is dead code. We're consolidating the three overlapping SecurityFilter PRs (#2498, #2500, #2552) and are moving forward with #2552, which keeps a hardened detection layer rather than removing the check outright, which is closer to what the issue asks for. Closing this one to shrink the queue, but please keep contributing — a fresh PR on another open issue would be very welcome.

PR #2499, opened 2025-12-28
fix: Implement escapeRegexChars in NameValidator

airajena:
## Description
Implemented the [escapeRegexChars](cci:1://file:///C:/Users/airaj/OneDrive/Desktop/gsoc/openelis/src/main/java/org/openelisglobal/validation/constraintvalidator/NameValidator.java:28:4-33:5) method in [NameValidator.java](cci:7://file:///C:/Users/airaj/OneDrive/Desktop/gsoc/openelis/src/main/java/org/openelisglobal/validation/constraintvalidator/NameValidator.java:0:0-0:0) which previously had a TODO stub that just returned the input unchanged.

## Changes
- Added null check with empty string fallback
- Escape special regex characters: \ ^ $ . | ? * + ( ) [ ] { }

## Why
The method is used to build regex patterns for name validation. Without proper escaping, special characters in the charset configuration could cause regex parsing errors or unintended pattern matching.

### DIFF
--- a/src/main/java/org/openelisglobal/validation/constraintvalidator/NameValidator.java
+++ b/src/main/java/org/openelisglobal/validation/constraintvalidator/NameValidator.java
@@ -27,8 +27,10 @@ public class NameValidator implements ConstraintValidator<ValidName, String>, Co
     SiteInformationService siteInformationService = SpringContext.getBean(SiteInformationService.class);
 
     private String escapeRegexChars(String regex) {
-        // TODO Auto-generated method stub
-        return regex;
+        if (regex == null) {
+            return "";
+        }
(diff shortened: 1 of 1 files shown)


mozzy11 (commented):
This is still a real gap — escapeRegexChars() on develop remains the auto-generated TODO stub, so a site-configured charset containing regex metacharacters can still corrupt the name-validation patterns. Your implementation looks right (null-safe, escapes the character-class-relevant metachars). The failing QA job is from a December run on our old CI pipeline; could you rebase onto latest develop so the new pipeline re-runs? If it comes back green this should be quick to merge — and it pairs nicely with your test PR #2527.

Issue #2292, opened 2025-11-03
Migrate from Junit4 to Junit5


Agaba-derrick:
JUnit 4 is now in maintenance mode and no longer receiving active updates. Since OpenELIS runs on Java 21 and uses Jakarta EE, it’s time to plan a migration to JUnit 5, which offers modern features, active support, and better Jakarta compatibility.

Goal
Evaluate and begin a phased migration from JUnit 4 to JUnit 5, starting with updating build dependencies and enabling hybrid mode JUnit Vintage) to support both frameworks during transition.

diyajn:
Can I work on that

PR #2510, opened 2025-12-30
issue-2292: Add unit tests for GenericValidator

airajena:
## Description
Adds unit tests for the GenericValidator utility class as part of the JUnit migration effort.
## Changes
- Added `GenericValidatorTest.java` with 7 test methods covering:
  - `isBool()` - validates boolean string values (true/false)
  - `is24HourTime()` - validates 24-hour time format (HH:mm)
## Related Issue
#2292


Agaba-derrick (changes requested):
@airajena  have you checked the existing tests we  have ?

airajena:
> @airajena have you checked the existing tests we have ?

Yes, I reviewed the existing tests. I noticed most tests in OpenELIS extend BaseWebContextSensitiveTest for integration testing with Spring context and database. However, GenericValidator is a simple utility class with static methods that don't require Spring context or database access. I followed the pattern of other simple utility tests - using plain JUnit 4 assertions (org.junit.Assert) without extending a base class. If you'd prefer this to follow a different pattern or extend a specific base class, please let me know and I'll update accordingly.

src/test/java/org/openelisglobal/common/util/validator/GenericValidatorTest.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class GenericValidatorTest {
+
+    @Test
+    public void isBool_shouldReturnTrueForTrue() {
+        assertTrue(GenericValidator.isBool("true"));
```
vsvishalsharma, line 12: @airajena how come we are testing it with three different values of true this needs to be standardised

src/test/java/org/openelisglobal/common/util/validator/GenericValidatorTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertTrue(GenericValidator.isBool("TRUE"));
+        assertTrue(GenericValidator.isBool("True"));
+    }
+
+    @Test
+    public void isBool_shouldReturnTrueForFalse() {
+        assertTrue(GenericValidator.isBool("false"));
+        assertTrue(GenericValidator.isBool("FALSE"));
```
vsvishalsharma, line 27: same here as well

src/test/java/org/openelisglobal/common/util/validator/GenericValidatorTest.java
airajena: Resolved, now  isBool tests is splitted into 4 separate methods

mozzy11:
Thanks @airajena

PR #2527, opened 2026-01-07
test: Add unit tests for NameValidator

airajena:
## Summary
Add test coverage for NameValidator constraint validator.

## Changes
- Created NameValidatorTest.java with 10 test methods
- Cover all NameType enum values (FIRST_NAME, LAST_NAME, USERNAME, FULL_NAME)
- Test null and empty input handling
- Test special cases (UNKNOWN_ for last name, apostrophes, hyphens)

### DIFF
--- a/src/test/java/org/openelisglobal/validation/NameValidatorTest.java
+++ b/src/test/java/org/openelisglobal/validation/NameValidatorTest.java
@@ -1,0 +1,112 @@
+package org.openelisglobal.validation;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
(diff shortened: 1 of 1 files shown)


mozzy11:
can run` mvn spotless:apply` to format your changes ??

mozzy11 (commented):
NameValidator still has zero test coverage on develop and your assertions check out against the implementation (null and malformed input both throw NumberFormatException where expected), so this is still worth landing. Two things before we can merge: (1) the unit-test job is failing — please run mvn spotless:apply and rebase onto latest develop so the current pipeline re-runs; (2) note that NameValidator pulls SiteInformationService from SpringContext at initialization, so if the tests fail at runtime you may need to extend BaseWebContextSensitiveTest. Also worth coordinating with your #2499, since merging that will change the regex behavior these tests pin down — the two PRs make a great pair.

PR #2522, opened 2026-01-05
Fixes (#2521): Clicking on cross icon, Clearing the search input results in error.

shauryag2002:
# Pull Requests Requirements
closes #2521 
- [x] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [x] The PR includes a video showing the changes for the work done.
- [x] The PR title follows conventional commit label standards.
- [x] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [x] The changes include tests or are validated by existing tests.
- [x] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary
This PR contains the fix of frontend bug where :
1. Go to homepage.
2. Click on search icon.
3. Notice a Input box will appear, Fill some words.
4. Then Cross icon will appear ( for clearing the input box )
5. Click on that cross icon.

## Screenshots

image

[Screencast from 2026-01-05 17-33-05.webm](https://github.com/user-attachments/assets/afd6ffd2-6df2-42aa-8477-2f89fb27fed1)

[Add relevant screenshots here if applicable]

## Related Issue
#2521 
[Add a link to the related issue or mention it here if applicable]

## Other

[Add any additional information or notes here]

PR #2536, opened 2026-01-09
solved #2522 Fix crash when clearing search input

Surajiitmjnu:
**What I fixed**

While testing the search feature, the app was crashing when clicking the clear (❌) button.

I found that setTextValue was being called even when it was not available.

**What I changed**

Added a safe check before calling setTextValue

Now the search can be cleared without any error or crash

**Why this change**

useAutocomplete does not always return setTextValue, so calling it directly was causing a runtime error.

**Testing**

Searched for a patient

Clicked clear button multiple times

Verified that the page no longer crashes

**Demo**

📹 Demo video attached above

https://github.com/user-attachments/assets/9721e6eb-1bbe-438a-8c37-19005ba97825

### DIFF
--- a/src/main/java/org/openelisglobal/common/controller/BaseController.java
+++ b/src/main/java/org/openelisglobal/common/controller/BaseController.java
@@ -298,7 +298,7 @@ public abstract class BaseController extends ControllerUtills implements IAction
 
         Errors errors = (Errors) inputFlashMap.get(Constants.REQUEST_ERRORS);
         if (errors != null) {
-            request.setAttribute(Constants.REQUEST_ERRORS, errors); // ✅ fixed
+            request.setAttribute(Constants.REQUEST_ERRORS, errors);
         }
 
         List<String> messages =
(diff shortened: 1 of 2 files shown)


src/main/java/org/openelisglobal/common/controller/BaseController.java
```diff
@@ -281,25 +281,40 @@ protected Errors getErrors() {
     }
 
     // move flash attributes into request
-    protected void addFlashMsgsToRequest(HttpServletRequest request) {
-        Map<String, ?> inputFlashMap = RequestContextUtils.getInputFlashMap(request);
-        if (inputFlashMap != null) {
-            Boolean success = (Boolean) inputFlashMap.get(FWD_SUCCESS);
+  protected void addFlashMsgsToRequest(HttpServletRequest request) {
```
mozzy11, line 284: These changes do not seem related to the current issue being fixed

Surajiitmjnu: @mazy
I’ve tested this work in my local environment and have also attached a video for reference. Please check.

Agaba-derrick:
@Surajiitmjnu  are you able to solve these conflicts ?

Surajiitmjnu:
@Agaba-derrick 

Yes sir

PR #2542, opened 2026-01-11
[FIX] :Enabling MulltiSelect in the old ui

Isabirye1515:
[Screencast from 2026-01-11 15-28-24.webm](https://github.com/user-attachments/assets/97046a55-7e14-4409-8133-a408b6b75ddd)

React Ui

[Screencast from 2026-01-12 23-16-22.webm](https://github.com/user-attachments/assets/6fd6bcb8-24fe-4bba-8708-54603659b456)


frontend/src/components/resultPage/SearchResultForm.js
```diff
@@ -1169,6 +1169,29 @@ export function SearchResults(props) {
...
+
+          case "C":
+            return (
+              <CascadingMultiSelect
+                id={`multiResult${row.id}`}
+                name={`testResult[${row.id}].multiSelectResultValues`}
+                dictionaryValues={row.dictionaryResults}
+                value={row.multiSelectResultValues}
+                onChange={(e) => handleChange(e, row.id)}
+                maxLevels={5}
```
vsvishalsharma: `maxLevels` prop is passed but not implemented check this once

frontend/src/components/common/multiSelect.js
```diff
@@ -0,0 +1,63 @@
+import { Column, MultiSelect } from "@carbon/react";
```
vsvishalsharma, line 1: Check the file name multSelect.js (WRONG - missing 'i')

src/main/java/org/openelisglobal/result/controller/LogbookResultsController.java
```diff
@@ -805,6 +824,41 @@ private ResultSignature createTechnicianSignatureFromResultItem(TestResultItem t
         return sig;
     }
 
+    private String extractMultiSelectJson(String input) {
+        // Pattern matches JSON objects like {"0":"860,1171"} or
+        // {"0":"860","1":"1171",... up to N keys}
+        Pattern pattern = Pattern
```
vsvishalsharma: The REGEX patterns has few problems

- Extremely hard to read and maintain
- Fragile - will break with minor JSON format changes
- Vulnerable to malformed input
- No comments explaining what it matches

Try to Use proper JSON parsing:

src/main/java/org/openelisglobal/result/service/LogbookPersistServiceImpl.java
```diff
@@ -75,8 +74,7 @@ public List<Analysis> persistDataSet(ResultsUpdateDataSet actionDataSet, List<IR
             // Check if result already exists for this specific Analysis (not Sample+Test)
             // This allows different aliquots (SampleItems) of the same sample to have
             // results for the same test type, since each aliquot has its own Analysis
-            checkResult = resultService.getResultsByAnalysis(checkAnalysis);
-            if (checkResult.size() == 0) {
+            if (resultSet.result.getId() == null) {
```
vsvishalsharma, line 74: **Old logic:** Check if results exist for this Analysis

**New logic**: Check if this specific result object has an ID

**These are NOT equivalent:**

- Old: Queries database for existing results
- New: Only checks in-memory object

vsvishalsharma (changes requested):

frontend/src/components/common/multiSelect.js
mozzy11: @Isabirye1515 have you seen this comment ??

src/main/java/org/openelisglobal/result/controller/LogbookResultsController.java
mozzy11: Thanks @Isabirye1515 .
Isnted of using this REGEX, 
simply parse out the list of strings into JSON objects , and just find the object with the largest Lenth for the id List

src/main/java/org/openelisglobal/result/service/LogbookPersistServiceImpl.java
mozzy11: @Isabirye1515 can you adress this comment

Isabirye1515: https://github.com/Isabirye1515/OpenELIS-Global-2/blob/9c109791866e488e50215909828d75045a497342/src/main/java/org/openelisglobal/common/services/ResultSaveService.java ,

Isabirye1515: handled here

Isabirye1515: checking through the analysis for the results, and equating to size zero and the save made the multiselect and cascading multiselect with looping values break

frontend/src/components/common/multiSelect.js
Isabirye1515: @vsvishalsharma , thanks

vsvishalsharma (commented):
Thanks for the clarification, @Isabirye1515.

I understand the issue now the old logic was checking for existing results at the Analysis level, but the new logic only checks whether the in-memory result object already has an ID. Because multi-select creates multiple result entries for the same Analysis, relying only on `result.getId() != null` would skip inserts when the first result already exists, which explains why the multiselect values were not being saved.

My earlier comment was pointing out that these two checks aren’t equivalent:

`getResultsByAnalysis(checkAnalysis)` → checks persisted DB state

`result.getId()` → checks only the current object

If your updated logic in ResultSaveService properly handles multi-select inserts while avoiding duplicates, I think that’s fine but it would help to explicitly document why we no longer rely on the Analysis-level check and how the new flow prevents duplicate inserts.

Could you add a brief comment in the code explaining this behavior?
That will make the intention clearer for future contributors

mozzy11:
Thanks @Isabirye1515

PR #2500, opened 2025-12-28
Refactor SecurityFilter: improve XSS handling, remove dead logic (#2492)

prakhar0608shrma:
## Pull Request Requirements
- [x] The PR title includes a brief description of the work done, including the Issue number.
- [ ] The PR includes a video showing the changes for the work done.
- [x] The PR title follows conventional commit label standards.
- [x] The changes conform to the OpenELIS Global x3 Styleguide and Design documentation.
- [x] The changes include tests or are validated by existing tests.
- [x] I have read and agree to the Contributing Guidelines of this project.

## Summary
Refactor `SecurityFilter` to improve XSS handling and remove dead logic. This change hardens request validation, avoids mutating request parameters, and simplifies filter logic for clarity and maintainability. The refactor ensures better security practices while keeping authentication and UI behavior unchanged.

## Screenshots
[Add relevant screenshots here if applicable]

## Related Issue
Fixes [#2492](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/2492)

## Other
- Branch name: `fix/securityfilter-hardening-2492`
- Target branch: `develop`
- Code formatted with `mvn spotless:apply` (separate commit if required)
- Verified build passes locally with `mvn clean install`
- Manually tested edge cases (missing headers, malformed parameters)
- CI pipeline checks are passing

PR #2552, opened 2026-01-14
issue-2492: Harden SecurityFilter XSS handling and remove dead code

airajena:
## Summary
Addresses #2492 by hardening the SecurityFilter with improved XSS detection and removing dead code.

## Changes
- **Improved XSS detection**: Replaced naive <script> tag check with comprehensive regex pattern that detects:
  - Script tags (with whitespace/attribute variations)
  - Event handlers (onclick, onerror, onload, etc.)
  - JavaScript protocol URLs
  - Data URIs with base64 encoding
  - VBScript protocol
  - CSS expression() attacks
- **Removed dead code**: Deleted unused exceptions ArrayList and related methods
- **Better error handling**: Returns HTTP 400 Bad Request instead of silent redirect to Dashboard
- **No parameter mutation**: Removed replaceAll() call that was modifying request parameters
- **Enhanced logging**: Includes URI, method, remote IP, and attack details

### DIFF
--- a/src/main/java/org/openelisglobal/security/SecurityFilter.java
+++ b/src/main/java/org/openelisglobal/security/SecurityFilter.java
@@ -9,22 +9,20 @@ import jakarta.servlet.ServletResponse;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Enumeration;
+import java.util.regex.Pattern;
 import org.openelisglobal.common.log.LogEvent;
 import org.openelisglobal.common.util.StringUtil;
 
 public class SecurityFilter implements Filter {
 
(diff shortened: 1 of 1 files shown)


mozzy11 (commented):
This is the strongest of the three open SecurityFilter PRs and we'd like to move forward with it — develop still has the old naive filter, so #2492 remains fully unaddressed, and your version covers the vectors the issue calls out (event handlers, javascript: URIs, data:base64) while removing the dead exceptions list and replacing the silent redirect with a 400. Next steps: rebase onto latest develop so the new CI pipeline re-runs, and double-check the regex against a couple of legitimate payloads that laboratories actually submit (e.g. result notes containing 'expression(' or 'on...=' -like text) so we don't introduce false-positive blocks on real data. We're closing #2498 and #2500 in favor of this one.

Agaba-derrick:
@airajena are you able to Address the above reviews Asap ?

PR #2573, opened 2026-01-20
OGC-291: Validation Page Redesign

OsauravO:
## Summary
This PR solves [OGC-291](https://uwdigi.atlassian.net/browse/OGC-291?issueKey=OGC-291&subProduct=jira-software) 
### Frontend (`frontend/src/components/validation/`)

**SearchForm**
- Lab unit dropdown
- Accession number quick search
- Collapsible advanced filters: lab number range, date range, test section, analyzer, entered by user, normal/abnormal/flagged checkboxes

**ValidationPage**
- Carbon DataTable with expandable rows
- Row selection: individual, select all, select all normal
- Batch actions bar: "Accept & Release Selected", "Retest Selected"
- Real-time stats: normal/abnormal/flagged counts
- Pagination controls

**ResultRow (Expanded)**
- Patient banner: name, ID, DOB, sex, age
- Entry info banner: entered by, date, method
- Editable fields: result value, interpretation, notes
- 5 tabs: Method, Order Info, Attachments, History, QC

**RetestModal**
- Confirmation modal with required reason textarea

**Internationalization**
- All strings externalized via React Intl (en/fr)

---

### Backend (`AccessionValidationRestController.java`)

**`GET /rest/AccessionValidation`**
- Query params: `labUnit`, `q`, `labNumberFrom/To`, `dateFrom/To`, `testSection`, `analyzer`, `enteredBy`, `normal`, `flagged`
- Server-side filtering via `applyAdditionalFilters()`

**`POST /rest/AccessionValidation`**
- Accepts `ResultValidationForm` with `isAccepted` flags
- Updates analysis status → `Finalized`, sets `releasedDate`, updates sample status
- Uses `processValidationDirectly()` (bypasses session cache for REST)

**`POST /rest/AccessionValidation/retest`**
- Accepts `RetestRequest` with `resultIds` and `reason`
- Validates only `TechnicalAcceptance` status analyses
- Sets status → `BiologistRejected`
- Creates internal note: `[RETEST REQUESTED] {reason}`

**`GET /rest/AccessionValidation/{id}/details`**
- Returns `ValidationDetailsDTO` with previous results history and order info
- On-demand loading for expanded row details

---

## Screen Recording

https://github.com/user-attachments/assets/31be3fec-e924-4ab5-9b7e-65e04de85092

[OGC-291]: https://uwdigi.atlassian.net/browse/OGC-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/openelisglobal/resultvalidation/service/ValidationDetailsServiceImpl.java
```diff
@@ -0,0 +1,192 @@
+package org.openelisglobal.resultvalidation.service;
```
Agaba-derrick: @OsauravO  thanks, can we have intergration tests that show this works perfect as the requirement.md doc required ?

frontend/src/components/validation/ResultRow.js
```diff
@@ -0,0 +1,322 @@
+import React, { useState } from "react";
```
Agaba-derrick: we can add cypress tests testing this work flow

src/main/java/org/openelisglobal/resultvalidation/service/ValidationDetailsServiceImpl.java
```diff
@@ -0,0 +1,192 @@
...
+import org.openelisglobal.test.service.TestService;
+import org.openelisglobal.test.valueholder.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+public class ValidationDetailsServiceImpl implements ValidationDetailsService {
+
+    private static final DateTimeFormatter DATE_FORMAT = DateTimeFormatter.ofPattern("MM/dd/yyyy");
```
Agaba-derrick: why do we have these hard coded ?
i would have an anticipation around 
_```
String dateString = _DateUtil.formatDateAsText(Date.from(analysis.getCompletedDate().toInstant()));
String dateTimeString = DateUtil.formatDateTimeAsText(Date.from(sample.getCollectionDate().toInstant()));_
```_

for more information you can check src/main/java/org/openelisglobal/common/util/DateUtil.java

File: src/main/java/org/openelisglobal/resultvalidation/service/ValidationDetailsServiceImpl.java [hunk 4]
```diff
@@ -0,0 +1,192 @@
...
+
+        if (previousResults.size() > 10) {
+            previousResults = new ArrayList<>(previousResults.subList(0, 10));
+        }
+
+        return previousResults;
+    }
+
+    @Override
+    @Transactional(readOnly = true)
```
vsvishalsharma, line 153: The UI exposes these tabs, but they're non-functional. please check

vsvishalsharma (commented):
the pr overall looks good to me consider the changes required in revamp of the validation page few stub implementation needs to completed

src/main/java/org/openelisglobal/resultvalidation/service/ValidationDetailsServiceImpl.java
OsauravO: You are correct, these are intentional stub implementations. The UI tabs are scaffolded and ready to display data when these features are implemented in backend, as OpenELIS currently doesn't have these.-
 - QC tracking tied to specific analyses
 - Reagent lot tracking at the analysis level
 - File attachment support for results
 - Delta Check requires test-specific thresholds (e.g., "flag if creatinine changes >30%")

vsvishalsharma (approved):
LGTM

vsvishalsharma (approved):
LGTM

frontend/cypress/e2e/validation.cy.js
```diff
@@ -1,12 +1,50 @@
...
-import PatientEntryPage from "../pages/PatientEntryPage";
-import Validation from "../pages/Validation";
 
 let homePage = null;
 let loginPage = null;
 let validation = null;
-let patientPage = new PatientEntryPage();
+
+// Mock data for validation results
+const mockValidationResults = {
```
mozzy11: Please these are e2e test . We dont mock responses. you would probably mock  in Unit tests but not in e2e tests

mozzy11:
Thanks @OsauravO for working on this . Good work.
I have tested this locally and found a few minor issues

1. You should support loading validation results by search param `accessionNumbe` ie `https://localhost/validation?type=order&accessionNumber=DEV01260000000000002`. On the dashBoard under `Ready for Validation` Card , we suport accesing the validation results via a url.
2. The Test section filter doesnt seem to work properly. its not stable.
3. The `Accept and release ` Button for each specific result entry doesnt seem to function
4. Try to mantain the same screen layout as the other pages ie Results pages.
5. Result modification on the Validation screen doesnt seem to work
6. Plese remove mock values from the E2E tests and ensure to use the actual response returned from the server so that we dont habour any error

OsauravO:
@mozzy11 
Thanks for the feedback. I have addressed all the points. Please let me know if anything else needs adjustment.

pmanko:
Hey @OsauravO — nice work on this redesign! The Carbon DataTable approach, expandable rows, and batch actions look great. A few things to address before we can merge:

**1. `fr.json` should not be modified in PRs**
We've adopted Transifex as the source of truth for non-English translations. Please revert all `fr.json` changes and add new keys to `en.json` only. Your French translations look good — we can upload them to [Transifex](https://explore.transifex.com/openmrs/openelis-1/) separately. A new CI check will enforce this going forward (#3269).

**2. Cypress tests → Playwright**
New E2E tests should use Playwright, not Cypress (Cypress is deprecated in this project). Please either convert `validation.cy.js` to Playwright or remove the Cypress files for now and add Playwright tests as a follow-up.

**3. Retest endpoint has business logic in the controller**
The `requestRetest()` method does `analysisService.update()` and `noteService.insert()` directly. Per our architecture guidelines, this should be in a service class (you already have `ValidationDetailsService` — the retest logic could live alongside it or in a new `RetestService`).

These are the blockers. A couple of non-blocking notes for your awareness:
- `applyAdditionalFilters()` does in-memory filtering after fetching the full result list — worth moving to the query layer for high-volume sites in a follow-up.
- `processValidationDirectly()` duplicates validation logic from the session-cache path — consider refactoring both paths to share core logic to avoid drift.

Happy to help with any of this!

OsauravO:
@pmanko 
Thanks for the detailed review! All three blockers are addressed in the latest push:                                                                                                   
                                                                                                                                                                                         
  1. fr.json — Reverted. New keys are in `en.json` only; happy to upload the French translations to Transifex separately.                                                                  
  2. Cypress → Playwright — Removed the new Cypress files and reverted modified ones. Will add Playwright tests as a follow-up PR once the feature is merged.                            
  3. Retest business logic — Moved out of the controller into `ValidationDetailsService.processRetestRequest()` with @Transactional. Controller now only handles HTTP request/response.    
                                                                                                                                                                                         
  Also took the opportunity to address the two non-blocking notes:                                                                                                                       
  - `applyAdditionalFilters()` moved to `ResultValidationService.applyFilters()` with a `ValidationFilterCriteria` bean — decoupled from the controller and positioned for query-layer pushdown
   as a follow-up.                                                                                                                                                                       
  - `processValidationDirectly()` now delegates to `createUpdateList()` so both save paths share the same core logic.

PR #2608, opened 2026-01-23
fix(postgres): resolve type mismatch in ReferenceTables duplicate check

### LINKED ISSUE #2606: SQL Type Mismatch in ReferenceTablesService
PostgreSQL is throwing a SQL grammar exception during `ReferenceTablesServiceTest` execution due to a type mismatch. The query is trying to compare `numeric` and `character varying` types without proper type casting, which PostgreSQL doesn't allow

##  Error Details

ERROR: operator does not exist: numeric <> character varying
Position: 294

org.hibernate.exception.SQLGrammarException: could not extract ResultSet

**Discovered during:** `mvn clean install` - Test execution  
**Test:** `ReferenceTablesServiceTest`  
**Error Time:** 23 Jan 2026 02:25:46  
**Database:** PostgreSQL  
**Position in Query:** Character 294

##   Investigation Steps
Analyze Test Data - See what parameters are being passed

nigar-08:
Fixes #2606

This PR resolves the PostgreSQL error operator does not exist: numeric <> character varying in ReferenceTablesServiceTest.

Root Cause:  
The test duplicateReferenceTable_shouldThrowException() creates a  ReferenceTables  object without an ID. In duplicateReferenceTablesExists() , this causes the code to default to using "0"  (a String) as the ID value. This string is then passed directly into a Hibernate query that compares it against the  id  column — which is  numeric  in PostgreSQL. Unlike MySQL, PostgreSQL does not allow implicit type conversion between string and numeric types, so the query fails.

Fix:  
Parse the ID as an `Integer` before setting it as a query parameter:

query.setParameter("param2", Integer.parseInt(referenceTablesId));

### DIFF
--- a/src/main/java/org/openelisglobal/referencetables/daoimpl/ReferenceTablesDAOImpl.java
+++ b/src/main/java/org/openelisglobal/referencetables/daoimpl/ReferenceTablesDAOImpl.java
@@ -156,7 +156,7 @@ public class ReferenceTablesDAOImpl extends BaseDAOImpl<ReferenceTables, String>
             }
 
             if (!isNew) {
-                query.setParameter("param2", referenceTablesId);
+                query.setParameter("param2", Integer.parseInt(referenceTablesId));
             }
             list = query.list();
             if (list.size() > 0) {


pmanko:
Is this error captured in a test?

pmanko:
ah I see the referenced issue

pmanko (approved):

PR #2613, opened 2026-01-25
  FHIR link update fails during patient merge when primary patient resource is missing

Surajiitmjnu:
### What changed

I added a pre-merge validation so that if either patient has a FHIR record, both patients must have one before the merge is allowed.
If this condition is not met, the merge is aborted and rolled back

### Why this change
Previously, the database merge could succeed even when the FHIR update failed, producing errors like:
“Primary patient FHIR resource not found”
This left OpenELIS and the FHIR server in an inconsistent state, which is unsafe for patient data.


mozzy11:
Thanks @Surajiitmjnu .
Any reason you deleted a bunch of existing files in this PR ??

Surajiitmjnu:
Hi @mozzy11
I mistakenly committed some unnecessary files yesterday, and this issue was included in that commit. I’ve since deleted the unnecessary files. Please let me know if anything else needs to be changed."

Agaba-derrick:
can you create a fresh clean PR for this issue, use the openelis PR requirement  doc that automatically comes in every new pull request,

Surajiitmjnu:
Hi @Agaba-derrick,
Yes, as per your suggestion, I will generate a fresh PR for thi

src/main/java/org/openelisglobal/patient/merge/service/PatientMergeServiceImpl.java
```diff
@@ -324,8 +324,9 @@ public PatientMergeDetailsDTO getMergeDetails(String patientId) {
      * patient, and creates audit trail. Runs in single transaction with rollback on
      * failure.
      */
-    @Override
-    public PatientMergeExecutionResultDTO executeMerge(PatientMergeRequestDTO request, String sysUserId) {
+    @Transactional(rollbackFor = FhirLocalPersistingException.class)
```
vsvishalsharma, line 327: Removing @Override eliminates compile-time checks and risks signature mismatches.

The method now throws a checked exception that likely isn’t declared in the interface.

Using rollbackFor = FhirLocalPersistingException.class is unnecessary, and if the exception is checked, it complicates rollback semantics.

vsvishalsharma (commented):
@Surajiitmjnu thansk for the fix have a minor suggestion/doubt when you raise a new pr please check if you can address that as well

Surajiitmjnu:
Hi @vsvishalsharma 
  Sure, I’ll take a look and handle this in the new PR.

pmanko:
@Surajiitmjnu closing since the work will be continued in a new PR. Please reference this PR in the new one. Thanks!

Issue #2628, opened 2026-01-27
PostgreSQL Type Mismatch Error in CronSchedulerServiceTest


Agaba-derrick:
A PostgreSQL type mismatch error occurs when running `CronSchedulerServiceTest`. The database is attempting to compare a `character varying` (VARCHAR) column with a `bytea` (binary data) value, causing a SQL operator error.

### Error Output
```
[INFO] Running org.openelisglobal.scheduler.service.CronSchedulerServiceTest
26 Jan 2026 01:04:08 -- ERROR -- ERROR: operator does not exist: character varying = bytea
  Hint: No operator matches the given name and argument types. You might need to add explicit type casts.
  Position: 448
```

### Error Details
- **Error Type:** SQL Operator Error
- **Database:** PostgreSQL
- **Test Class:** `org.openelisglobal.scheduler.service.CronSchedulerServiceTest`
- **Error Position:** Character 448 in SQL query
- **Type Mismatch:** Comparing `character varying` with `bytea`

### Root Cause
PostgreSQL is strict about type comparisons. The error indicates:
1. A VARCHAR column is being compared to binary data (bytea)
2. The query at position 448 contains the problematic comparison
3. An explicit type cast is needed to resolve the mismatch

**Common causes:**
- ORM (Hibernate/JPA) mapping issue where a String field is mapped to a bytea column
- Incorrect parameter binding (String value being passed as byte array or vice versa)
- Database schema mismatch with entity definitions

### Impact
- Indicates potential data type inconsistency in production code
- Could affect scheduler functionality

### Steps to Reproduce
1. Run the test:
   ```bash
   mvn test -Dtest=CronSchedulerServiceTest
   ```
2. Observe the PostgreSQL operator error

ramzende14:
@Agaba-derrick Hey! Can I take this issue? I’d like to investigate and submit a fix.

waleedahmedwaheed:
This error is happening because the database is being asked to compare two different data types that PostgreSQL does not automatically convert:

- One side is character varying (VARCHAR)

- The other side is bytea (binary data)

PostgreSQL is strict with type comparisons, so it throws this error when no matching operator exists.

What is likely going wrong

This usually means one of the following is true:

- The JPA/Hibernate entity maps a String field to a bytea column in PostgreSQL
 
- A query parameter is being bound as a byte array when the column expects text
 
- The database schema does not match the entity field type

[...]

PR #2636, opened 2026-01-28
Fixed issue-2628 - PostgreSQL Type Mismatch Error in CronSchedulerServiceTest

sagarkhandagre998:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Problem
When `null` values were passed to query methods (e.g., `getCronScheduleByJobName(null)`), Hibernate's Criteria API incorrectly inferred the parameter type as `VARBINARY` (PostgreSQL `bytea`) instead of the actual column type (VARCHAR). This caused PostgreSQL to throw:
```
ERROR: operator does not exist: character varying = bytea
  Position: 448
```

### Root Cause
In the generated SQL query:
```sql
WHERE cronschedu0_.job_name = ?
```

Hibernate was binding the `null` parameter as `VARBINARY` (which becomes `bytea` in PostgreSQL) when it should bind as `VARCHAR` for the `job_name` column. 

This occurred in `BaseDAOImpl.addWhere()` where `criteriaBuilder.equal(pathToProperty, null)` didn't provide type information, causing Hibernate to default to binary type inference.

**### Solution**
Added explicit null handling in the `EQ` case to use `criteriaBuilder.isNull()` instead of `criteriaBuilder.equal()` when the value is null. This follows SQL best practices (`IS NULL` vs `= NULL`) and prevents type inference issues.

### Changes
- Modified `BaseDAOImpl.java`** (line 677-679): Added explicit null handling in the `EQ` case to use `criteriaBuilder.isNull()` instead of `criteriaBuilder.equal()` when the value is null
- Added 5 lines of code with explanatory comment
- Modified `SampleProjectServiceTest.java`** (line 64-68): Updated test `getSampleProjectBySampleId_shouldReturnNullForNullInput` to expect `null` return instead of exception

Behavioral Change
**Before Fix:** Passing `null` to service query methods threw `PersistenceException` (unintended side effect of the bytea bug)

**After Fix:** Passing `null` to service query methods returns `null` (no matching results found)

**Rationale:** The exception was never intentional - service methods have no explicit null validation, and the pattern throughout the codebase is to return `null` for "not found" scenarios. The original test was accidentally passing due to the bug, not testing actual business logic. This change makes the behavior consistent with methods like `getSampleProjectBySampleId("999")` which also returns `null` for non-existent IDs.

### Testing
- All 16 tests in `CronSchedulerServiceTest` pass
- Bytea comparison error no longer appears in logs
- Verified SQL logging shows correct `IS NULL` predicate instead of `= ?` with bytea binding

## Screenshots

## Related Issue
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/2628

## Other


Agaba-derrick:
thanks @sagarkhandagre998 the Pr title should follow the naming convetion  not using the issue number

sagarkhandagre998:
> thanks @sagarkhandagre998 the Pr title should follow the naming convetion not using the issue number

Sure sorry for that

pmanko (approved):

mozzy11 (commented):
The core fix here is still valid: develop's BaseDAOImpl EQ branch still builds criteriaBuilder.equal(path, null) with no null guard, which generates a '= NULL' predicate that can never match and can trigger PostgreSQL type-inference errors — switching to isNull is the correct defensive fix, and your CI is fully green. Two asks: (1) drop the SampleProjectServiceTest hunk, which only renames a variable and adds a comment (that test already passes on develop), so the PR is a pure one-spot fix; (2) fix the duplicated line in the code comment. Note we're closing your #2672, which carries the identical change — let's land it through this PR. Ideally add a small test demonstrating getAllMatching with a null value to lock the behavior in.

PR #2691, opened 2026-02-03
refactor(speckit): simplify install with cross-platform Python script

pmanko:
## Summary

- Replace bash/PowerShell install scripts with single Python script
- Move installer to `scripts/install-speckit-commands.py` (repo root level)
- Update core commands to use `.specify/` paths directly (no runtime transforms)
- Add `.python-version` for version managers (pyenv, asdf, uv)

## Why

The previous install script:
1. Was buried in `.specify/scripts/bash/` (circular: inside what it installs)
2. Required parallel bash + PowerShell implementations
3. Did slow runtime path transformations using bash string substitution

The new Python script:
- Works identically on Windows, macOS, and Linux
- Is simpler to maintain (one file vs two)
- Runs faster (no path transformations needed at install time)
- Lives at repo root level (`scripts/`) for better discoverability

## Test plan

- [x] Run `python scripts/install-speckit-commands.py -y all`
- [x] Verify commands installed to `.claude/commands/` and `.cursor/commands/`
- [x] Verify paths in installed commands use `.specify/` prefix


pmanko:
## Addressed Review Comments

Thanks for the thorough review! Here's a summary:

### ✅ Fixed

1. **Injection logic restored** (commit 377969e) - OE overrides now inject at strategic locations:
   - `speckit.tasks.md`: BEFORE "## Task Generation Rules"
   - `speckit.implement.md`: BEFORE "## User Input"
   - Other commands: append at end
   
   This ensures AI agents see OE overrides before the core rules they override.

2. **Docstring updated** - Now correctly says "extensions/overrides" and documents the injection behavior.

3. **Exception handling added** - File I/O operations now have try-except with user-friendly error messages.

### ⏭️ Deferred / Dismissed

4. **Leading slash in paths** (`/.specify/` vs `.specify/`) - These paths came from the original upstream SpecKit files which used `/memory/constitution.md`. Our sed transformation converted `/memory/` → `/.specify/memory/`. While `.specify/memory/...` would be cleaner, AI agents interpret these as documentation references (not literal filesystem paths), so functionality is unaffected. Can clean up in a follow-up PR if desired.

5. **CI validation for `/.specify/` paths** - Adding this check would immediately fail CI since we DO have `/.specify/` paths in the files from the transformation. This would need the path cleanup (item 4) to happen first.

Issue #2648, opened 2026-01-29
Form Validation Failures and Dropdown Data Not Loading Across Multiple Modules


Mouli24:
Several input fields and dropdown components across the application are either accepting invalid input or not functioning / not populated at all. These issues affect data integrity, user experience, and core workflows such as order entry, reporting, and project configuration.

**Issue 1: Time fields accept invalid input (text & incorrect digit length)
Affected Screens**
Add Order → Sample → Collection Time
Batch Order Entry Setup → Reception Time (hh:mm)
Sample Information → Collection Time

**Problem**
Time fields accept:
Arbitrary numeric input (e.g. 12333, 12555)
Alphabetic characters (e.g. 12:ss)
No enforcement of HH:mm format

**Expected Behavior**
Allow only valid time values
Enforce HH:mm format
Block text input and invalid digit lengths

**Actual Behavior**
Inputs behave like unrestricted text fields
Invalid values are accepted without validation

**Issue 2: Export Format dropdown not working**
Location
Inventory → Inventory Management → Reports
**Problem**
Export Format dropdown does not open / does not show any selectable options

**Expected Behavior**
Dropdown should list available export formats (e.g. Excel, CSV, PDF)

**Actual Behavior**
Dropdown is non-functional / empty

**Issue 3: Experiment Type dropdown not working
Location**
Notebook Dashboard → Project Definition Form → MetaData tab

**Problem**
Experiment Type dropdown does not populate or respond
User is unable to select any experiment type

**Expected Behavior**
Dropdown should show available experiment types

**Actual Behavior**
Dropdown appears empty / unusable

 **Issue 4: Ward / Dept / Unit dropdown is empty**
Location
Batch Order Entry Setup → Configure Barcode Entry → Patient Info

**Problem**
Ward / Dept / Unit dropdown shows no options

**Expected Behavior**
Dropdown should be populated with available wards/departments/units

**Actual Behavior**
Dropdown is empty

 **Issue 5: Filter by Location dropdown is empty**
Location
Storage → Samples → Sample Items

**Problem**
Filter by locations dropdown has no selectable options

**Expected Behavior**
Dropdown should list available storage locations (rooms, shelves, racks, etc.)

**Actual Behavior**
Dropdown is empty, making filtering impossible

![Image](https://github.com/user-attachments/assets/7b795670-d6ee-46be-880f-d0d052046861)

[...]

PR #2692, opened 2026-02-03
 Fixed Collection time and Export format

Mouli24:
### Fixed Collection time and Export format
# Pull Requests Requirements

- [✔ ] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [✔ ] The PR includes a video showing the changes for the work done.
- [✔ ] The PR title follows conventional commit label standards.
- [✔ ] The changes confirm to the OpenElis Global x3
      documentation.
- [✔ ] The changes include tests or are validated by existing tests.
- [✔ ] I have read and agree to the Contributing
      project.

## Summary
In this PR, I have worked on issue related to broken inputs and dropdowns across the application. I have successfully fixed Issue 1 and Issue 2, which include:
Proper validation for Time fields (HH format, blocking invalid text and incorrect digit lengths), and
Fixing the Export Format dropdown so it now opens and shows available options.

Issue 3 (Experiment Type dropdown) has not been fixed through code in this PR. After investigating, I found that this is a data configuration issue the dropdown depends on values from notebook_seed.xml, which needs to be executed to populate the options.

Similarly, for Issue 4 (Ward/Dept/Unit dropdown) and Issue 5 (Filter by Location dropdown), I confirmed that these are also data configuration issue, not code bugs:

The Ward/Dept/Unit dropdown works correctly in the code but requires departments to be configured for each site.

The Location Filter dropdown functions properly but needs storage rooms/locations to be created via the Admin panel before options appear.

Therefore, no code changes were required for Issues 3, 4, and 5 — they depend on proper system setup and seed data. I would appreciate maintainer guidance on whether these configuration steps should be documented for future users
## Screenshots
image
[Screencast from 2026-02-03 19-55-28.webm](https://github.com/user-attachments/assets/eeee17b0-7a7a-4c6f-9aa7-c8700c27ffe8)

## Related Issue
#2648 

## Other


Bahati308 (commented):
This component does a good job enforcing the `HH:mm` format, which helps with fixed collection time and export consistency.
However, the `useEffect` that calls `props.onChange(currentTime)` on mount uses an empty dependency array while referencing external values. This can lead to stale closures and potentially duplicate `onChange` calls (since updates already occur in `handleTimePicker`).
Could you clarify the intent of this effect and either:
remove it if redundant, or
include the correct dependencies and explain why the initial `onChange` is required?

Bahati308:
you could update your branch as well. thanks for the contribution

Mouli24:
> This component does a good job enforcing the `HH:mm` format, which helps with fixed collection time and export consistency. However, the `useEffect` that calls `props.onChange(currentTime)` on mount uses an empty dependency array while referencing external values. This can lead to stale closures and potentially duplicate `onChange` calls (since updates already occur in `handleTimePicker`). Could you clarify the intent of this effect and either: remove it if redundant, or include the correct dependencies and explain why the initial `onChange` is required?

The useEffect with [] is intentional. It handles the case where autofillTime sets the initial value via useState during mount. In that scenario, handleTimePicker is not called, so without this effect the parent would not be notified of the initial value.
The effect runs only once to sync that initial autofilled time with the parent. All subsequent updates are handled directly inside handleTimePicker, which already calls onChange.
Previously, using [currentTime] caused duplicate onChange calls once from the handler and once from the effect. This change prevents that issue.
The empty dependency array is deliberate since we only want the initial value from the first render, not future updates.

mozzy11:
Thanks for digging into the Inventory Reports export issue! Closing this one during a triage pass of older PRs: the itemToString fix here overlaps with #2751, which covers both InventoryReports and InventoryReportsModal, and the remaining changes touch the Cypress workflow, which has since been deprecated in favor of Playwright. If the collection-time issue you originally saw still reproduces on current develop, we'd genuinely welcome a fresh, focused PR for just that fix.

PR #2701, opened 2026-02-04
Improve error handling and diagnostics in UserRoleDAOImpl

kmvipin:
### Describe the problem / current behavior

Many methods in `org.openelisglobal.userrole.daoimpl.UserRoleDAOImpl` catch exceptions (HibernateException / RuntimeException), log them via `LogEvent.logError(e)`, and then either:

- silently return partial/empty results, or
- re-throw a generic `LIMSRuntimeException` **without any contextual information** (userId, roleName, etc.)

This makes it very hard to diagnose permission / role lookup failures in production.

Examples of problematic patterns:

1. `getRoleIdsForUser(String userId)`
   - Catches → logs → returns empty list → caller has no idea why user appears to have no roles

2. `userInRole(String userId, String roleName)` and the Collection variant
   - Catch → log → return `false` → looks like user simply doesn't have the role

3. `deleteLabUnitRoleMap(LabUnitRoleMap roleMap)`
   - Silent delete failure possible

When something goes wrong (DB connectivity, cast issues, constraint violations, etc.), support tickets become very hard to investigate because the only clue is a generic log line without context.

### Why this matters

- Security-sensitive area (roles & permissions)
- Silent failures lead to mysterious symptoms: "user can't access feature X but has the role assigned"
- Wastes a lot of time for implementers / support teams
- Harder to create good bug reports / tickets from the field

### Suggested improvement

Add **contextual information** to both logs **and** the thrown `LIMSRuntimeException` so that:

- Logs are immediately actionable
- Exceptions contain enough info to understand what failed without reading code


kmvipin:
@mozzy11, kindly review this!

Bahati308:
hello @kmvipin , could you keep your branch updated to fetch the latest changes, which makes the review updated as well. Thanks for the initiative

kmvipin:
> hello @kmvipin , could you keep your branch updated to fetch the latest changes, which makes the review updated as well. Thanks for the initiative

Sure, I will do merge

mozzy11:
Thank you for the detailed writeup on the swallowed exceptions in UserRoleDAOImpl — the diagnosis is spot-on and better error context there would be valuable. Closing during a triage pass because the backend Build + Test check is failing and changing lookup failures from silent-empty to rethrow alters permission-check behavior, which needs tests to land safely. UserRoleDAOImpl is unchanged on develop, so if you're still interested, a fresh PR rebased on current develop with a green build and a couple of tests covering the new error paths would be very welcome.

PR #2703, opened 2026-02-04
fix: Untrack .env file and add .env.example template

pmanko:
## Problem
The `.env` file was tracked in git, causing it to be deleted by `git reset --hard` and branch switches. This resulted in repeated loss of server-specific configurations (database passwords, domain settings, etc.).

## Solution
1. **Untracked .env** - Removed from version control (already in .gitignore)
2. **Added .env.example** - Template with all required variables, configured for `analyzers.openelis-global.org` domain
3. **Updated documentation** - Added clear prerequisites in README.md and AGENTS.md

## Changes
- ✅ Remove `.env` from git tracking (`git rm --cached`)
- ✅ Add `.env.example` with comprehensive documentation
- ✅ Update README.md: Add prerequisites section with .env setup
- ✅ Update AGENTS.md: Add critical .env configuration section

## Migration Steps for Developers/Servers

After this PR is merged, each developer/server needs to:

\`\`\`bash
# On develop branch after pulling
cp .env.example .env

# Customize .env for your environment:
# - Set LETSENCRYPT_DOMAIN (localhost for dev, analyzers.openelis-global.org for production)
# - Set LETSENCRYPT_EMAIL (required for real SSL certificates)
# - Customize passwords if needed (defaults are fine for dev)
\`\`\`

## Why This Matters
- `.env` contains secrets and should NEVER be in version control
- Each developer/server needs their own customized `.env`
- Docker Compose requires `.env` for `\${VAR}` substitution
- Missing `.env` causes authentication failures and SSL certificate errors

## Test Plan
- [x] Verified `.env` is untracked after this change
- [x] Verified `.env.example` contains all required variables
- [x] Tested `cp .env.example .env` workflow
- [x] Documentation is clear and comprehensive

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -107,6 +107,29 @@ mvn clean install -DskipTests
 - **Node.js 16+**: Frontend development
 - **Git with submodules**: `git submodule update --init --recursive`
 
+### Environment Configuration (.env file) - CRITICAL
+
+**IMPORTANT:** Before running any `docker compose` command, you MUST create a
+`.env` file:
+
+```bash
+cp .env.example .env
+```
+
+Then customize `.env` for your environment (database passwords, domain, etc.).
+
+**Why this matters:**
+
+- `.env` is in `.gitignore` (contains secrets and server-specific settings)
+- **`.env` is intentionally NOT tracked in git** - each developer/server needs
+  their own
+- Docker Compose uses `.env` for `${VAR}` substitution in compose files
+- Missing `.env` causes authentication failures and SSL certificate errors
+
+**Common mistake:** Running `git reset --hard` or switching branches used to
+delete `.env` because it was accidentally tracked in git. This has been fixed -
+`.env` is now untracked.
+
 ---
 
 ## Technology Stack
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -126,10 +126,6 @@ Then customize `.env` for your environment (database passwords, domain, etc.).
 - Docker Compose uses `.env` for `${VAR}` substitution in compose files
 - Missing `.env` causes authentication failures and SSL certificate errors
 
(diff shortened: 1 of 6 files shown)


mozzy11:
Thanks @pmanko . i think i dont see a reason we would remove the default `.env` file for some one to create it manually every time they need to run the default setup.

I would just create instructions to update the credentials for a production setup

pmanko:
I worry about putting .env files in git, since users can put in sensitive information that can then accidentally get commited.

pmanko:
the .env file should have configuration settings, and be checkout-specific in my opinion, not part of git history.

pmanko:
https://medium.com/@nafisafaruna/a-developers-guide-to-safeguarding-env-files-in-git-41161789516e

pmanko:
@mozzy11 and practically, any local changes to the .env file have to currently be ignored or they can be accidentally pushed, regardless of how sensitive they might be.

mozzy11:
Thanks @pmanko

pmanko:
add defaults in docker-compose to prevent missing .env from breaking dev

AGENTS.md
```diff
@@ -107,6 +107,29 @@ mvn clean install -DskipTests
...
+
+- `.env` is in `.gitignore` (contains secrets and server-specific settings)
+- **`.env` is intentionally NOT tracked in git** - each developer/server needs
+  their own
+- Docker Compose uses `.env` for `${VAR}` substitution in compose files
+- Missing `.env` causes authentication failures and SSL certificate errors
+
+**Common mistake:** Running `git reset --hard` or switching branches used to
+delete `.env` because it was accidentally tracked in git. This has been fixed -
+`.env` is now untracked.
```
pmanko: unneccessary content - needs to be simple and clear.

pmanko: @copilot fix please

Issue #2748, opened 2026-02-07
UI issue: Collection Time field overflows section on smaller screen sizes


manroop-2005:
Description

On the Test Request → Sample section, there is a layout issue when the screen dimensions are reduced (tablet/mobile view).

When switching to smaller screen sizes, the Collection Time input field extends outside the sample section container instead of staying aligned within it.

Expected Behavior

The Collection Time field should remain contained within the Sample section.

All date and time inputs should realign or wrap correctly when the screen size changes, maintaining a clean and readable layout.

Actual Behavior

The Collection Time input overflows outside the Sample section when viewed on smaller screens.

This causes misalignment and breaks the visual structure of the form.

Environment

Page: Test Request → Sample

View: Responsive / Mobile view

Browser: Chrome (DevTools responsive mode)

Screenshots

https://github.com/user-attachments/assets/71595913-6cd5-4e2a-996d-7caf46a829bc

manroop-2005:
Hi @Agaba-derrick , I am working on it and I will raise the pr shortly.

PR #2750, opened 2026-02-07
Fix timpicker input field overflow in mobile screen

manroop-2005:
# Changes

Fixed TimePicker input overflow on mobile by applying a scoped, responsive style.

# Pull Requests Requirements

      Issue number if applicable.
- [X] The PR includes a video showing the changes for the work done.
      documentation.
      project.

## Screenshots

https://github.com/user-attachments/assets/40b5add0-6765-4507-b972-ac0bfa55bd9f

## Related Issue

[https://github.com/DIGI-UW/OpenELIS-Global-2/issues/2748]

## Other


manroop-2005:
@pmanko can you review this ?

pmanko (changes requested):
The `width: 200px !important` applies at all viewport sizes — on desktop this shrinks the timepicker from ~510px to 200px, which is a visual regression.

Please wrap this in a media query so it only applies on small screens:
```css
@media (max-width: 672px) {
  .inlineDiv .cds--time-picker__input-field {
    max-width: 200px;
  }
}
```

mozzy11 (commented):
Thanks — I checked develop and this fix is still needed (the global 31.875rem !important width is still there with no scoped override), your branch merges cleanly, and CI is fully green on the new pipeline, nice!

One question before merging: the new rule applies at **every** screen size, so on desktop the time picker inside `.inlineDiv` will shrink from ~510px to 200px as well. If that desktop change is intentional and looks good, could you add a before/after screenshot at desktop width? Otherwise, wrapping the rule in something like `@media (max-width: 672px) { ... }` would keep the fix mobile-only. Either way this is very close — happy to merge once that's settled.

frontend/src/components/addOrder/add-order.scss
```diff
@@ -132,6 +132,10 @@
   width: 31.875rem !important;
 }
 
+.inlineDiv .cds--time-picker__input-field {
+  width: 200px !important;
```
mozzy11, line 136: Still the open item from pmanko's review: the fixed `width: 200px !important` applies at every viewport, regressing the desktop time picker (~510px → 200px). Wrap in `@media (max-width: 672px)` and prefer `width: 100%; max-width: ...` so the input tracks its container instead of fighting the existing `31.875rem !important` global with another `!important`.

mozzy11 (commented):
Checking in on this one — @pmanko's CHANGES_REQUESTED from the earlier review is still the blocking state and **hasn't been addressed**: the only commits since the original change are two develop merges, and the diff still contains the unconditional `width: 200px !important`, which shrinks the desktop time picker from ~510px to 200px at every viewport.

The ask is unchanged and small (inline): wrap the rule in a small-screen media query and use a container-relative width instead of stacking another `!important` on the existing global override. Once that's pushed, re-request pmanko's review.

Issue #2745, opened 2026-02-07
In the inventory  report section Export Format dropdown text is not visible


AnuragSingh2101:
### Steps to reproduce:
1.  Go to inventory section
2.  Go to report section
3. then check out the format dropdown

### screenshot
<img width="1325" height="524" alt="Image" src="https://github.com/user-attachments/assets/9564f09e-22b5-443d-93fe-abee74c18d4e" />

AnuragSingh2101:
@Agaba-derrick , I want to work on this can you please assign this to me.

PR #2751, opened 2026-02-07
Fix Dropdown display by adding itemToString

AnuragSingh2101:
# Description
This PR fixes an issue where the Export Format dropdown in Inventory Reports opened with no visible options.

#2745 

## Screenshots
image

## Cause:
- Carbon Dropdown requires item ToString to render object items

## Fix:
- Added itemToString to export format dropdown

## Related Issue
N/A


mozzy11 (commented):
Good catch — I verified this still reproduces on current develop: the Export Format dropdowns in both InventoryReports.jsx and InventoryReportsModal.jsx are still missing itemToString, so your fix is exactly right. Two things needed to get this in: please rebase on current develop and trim the en.json changes down to only the new keys your fix needs (the large storage.* block in your diff has since landed on develop and would duplicate keys), and drop the fr.json edits — non-English translations flow through Transifex, so new keys go in en.json only. Once that's done this should be a quick review. Thanks for sticking with it!

PR #2767, opened 2026-02-08
feat(011): Madagascar Analyzer Integration, management UI, updated sidenav navigation

pmanko:
## Summary
Complete analyzer integration for the Madagascar deployment: 12 analyzers across 4 protocols (ASTM, HL7, file, serial), full-stack management UI, sidenav revamp, and comprehensive test coverage.

- **Backend** (~200 files): Controllers, DAOs, services, valueholders, forms, analyzer readers, HBM→JPA migration, 26 Liquibase migrations
- **Frontend** (~120 files): 70+ analyzer management components, sidenav layout revamp, reports fixes
- **Tests** (~80 files): 65+ backend tests, 10 Playwright specs, 18 Cypress test updates
- **Infrastructure**: Analyzer docker compose, harness project, volume configs, analyzer-defaults JSON

## Supported Analyzers

| Analyzer | Protocol | Transport |
|----------|----------|-----------|
| FACScalibur | CSV | File |
| Fluorocycler | CSV | File |
| GeneXpert | ASTM / CSV | TCP / File |
| Mindray BC-30s | HL7 | MLLP |
| Mindray BS-240/480 | ASTM | TCP |
| Stago Start 4 | ASTM | TCP |
| CobasTaqman | CSV | File |
| Sysmex XN/XP | ASTM | TCP |
| ABX Micros | ASTM | TCP |
| ABX Pentra 80 | ASTM | TCP |
| Horiba Micros 60 | ASTM | TCP |
| Abbott Architect | HL7 | MLLP |

## Key Architecture Decisions
- 5-layer pattern: Valueholder → DAO → Service → Controller → Form
- Plugin-based analyzer readers (extensible via plugins submodule)
- HBM→JPA annotation migration for Analyzer, AnalyzerResults, AnalyzerTestMapping
- Sidenav uses Carbon Design System components

## Context
This is **PR 4 of 4** splitting `demo/madagascar` (60 commits) into reviewable PRs:
1. Documentation and specs (#2764)
2. Submodule renames and updates (#2765)
3. CI, testing infra, translations (#2766)
4. **This PR** — Analyzer code, tests, and sidenav UI

**Merge order**: This PR should be merged LAST, after PRs 1-3. A rebase onto develop (after PRs 1-3 merge) may be needed to resolve pom.xml and workflow conflicts.

## Test plan
- [ ] `mvn clean install -DskipTests -Dmaven.test.skip=true` compiles
- [ ] `mvn spotless:check` passes
- [ ] `cd frontend && npm run format -- --check` passes
- [ ] Backend unit tests pass: `mvn test`
- [ ] Playwright specs pass for analyzer pages
- [ ] Docker compose analyzer setup validates


frontend/src/components/analyzers/FieldMapping/ValidationDashboard.jsx
```diff
@@ -0,0 +1,321 @@
...
+  TableHead,
+  TableRow,
+  TableHeader,
+  TableBody,
+  TableCell,
+  Tag,
+  ProgressBar,
+  InlineNotification,
+  Loading,
+} from "@carbon/react";
```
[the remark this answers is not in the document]
pmanko: Fixed in commit 55331d7 — removed the unused `DataTable` import from `ValidationDashboard.jsx`.

frontend/src/components/analyzers/FieldMapping/ValidationDashboard.jsx
```diff
@@ -0,0 +1,321 @@
...
+        );
+      }
+      setValidating(false);
+    });
+  };
+
+  const handleViewTestHistory = () => {
+    // TODO: Implement test history modal
+    // This would load historical test execution results
+    console.log("Test history feature not yet implemented");
```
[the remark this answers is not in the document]
pmanko: Fixed in commit 55331d7 — removed the `console.log` call and disabled the button with a TODO comment indicating the feature is not yet implemented.

frontend/src/components/analyzers/FieldMapping/ValidationDashboard.jsx
```diff
@@ -0,0 +1,321 @@
...
+  // Only display when lifecycle stage is VALIDATION (early return AFTER hooks)
+  if (status !== "VALIDATION") {
+    return null;
+  }
+
+  const loadValidationMetrics = () => {
+    setLoading(true);
+    setError(null);
+
+    const endpoint = `/rest/analyzer/analyzers/${analyzerId}/validation-metrics`;
```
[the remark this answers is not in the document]
pmanko: Fixed in commit 55331d7 — switched to relative path. Both `getFromOpenElisServer` and `postToOpenElisServerJsonResponse` prepend the base URL internally, so using `${config.serverBaseUrl}` was causing a double-prefix bug. Now both calls use consistent relative paths.

frontend/src/components/analyzers/FieldMapping/ValidationDashboard.jsx
```diff
@@ -0,0 +1,321 @@
...
+    });
+  };
+
+  const handleValidateAllMappings = () => {
+    setValidating(true);
+    setError(null);
+
+    // Trigger validation for all mappings
+    // This would call a validation endpoint that tests all configured mappings
+    const endpoint = `${config.serverBaseUrl}/rest/analyzer/analyzers/${analyzerId}/validate-all-mappings`;
```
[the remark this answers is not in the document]
pmanko: Same fix as above (commit 55331d7) — both GET and POST now use consistent relative paths to avoid double-prefixing.

frontend/src/components/analyzers/ErrorDashboard/ErrorDetailsModal.jsx
```diff
@@ -0,0 +1,266 @@
...
+        ? "magenta"
+        : "blue";
+
+  // Get error type label
+  const errorTypeKey = `analyzer.errorDashboard.errorType.${errorType.toLowerCase()}`;
+  const errorTypeLabel = intl.formatMessage({ id: errorTypeKey }, errorType);
+
+  // Get severity label
+  const severityKey = `analyzer.errorDashboard.severity.${severity.toLowerCase()}`;
+  const severityLabel = intl.formatMessage({ id: severityKey }, severity);
```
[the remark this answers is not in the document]
pmanko: Fixed in commit 55331d7 — now using `{ id, defaultMessage }` descriptor pattern instead of positional fallback string. Both `errorTypeLabel` and `severityLabel` use `intl.formatMessage({ id: key, defaultMessage: fallback })` as suggested.

frontend/src/App.js
```diff
@@ -418,6 +389,42 @@ export default function App() {
...
+                <SecureRoute
+                  path="/GenericSample/Import"
+                  exact
+                  component={() => {
+                    const GenericSampleOrderImport =
+                      require("./components/genericSample/GenericSampleOrderImport").default;
+                    return <GenericSampleOrderImport />;
+                  }}
+                  role={Roles.RECEPTION}
+                />
```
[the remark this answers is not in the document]
pmanko: Fixed in commit 55331d7 — reverted to `role=""` on these routes. The role change was unintentional and not related to the analyzer routing feature.

frontend/src/components/analyzers/FieldMapping/QueryStatusModal.jsx
```diff
@@ -0,0 +1,149 @@
...
+    };
+    poll();
+    timerRef.current = setInterval(poll, 2000);
+    return () => {
+      if (timerRef.current) {
+        clearInterval(timerRef.current);
+        timerRef.current = null;
+      }
+    };
+  }, [open, analyzerId, jobId]); // Removed onCompleted from dependencies
```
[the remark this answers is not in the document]
pmanko: Intentional. Line 27 uses `completedRef` (`useRef`) to capture `onCompleted` by reference, avoiding restarting the polling interval on parent re-renders. The ref is always updated to the latest `onCompleted` before it's called (line 28: `completedRef.current = onCompleted`), so stale closures are not possible. The comment on line 72 documents this deliberate exclusion from the dependency array.

frontend/package.json
```diff
@@ -56,21 +57,22 @@
...
-    "test:e2e:single": "cypress run --browser electron --config video=false,defaultCommandTimeout=10000 --spec",
-    "test:e2e:full": "cypress run --headless --browser chrome",
     "format": "prettier ./ --write",
     "check-format": "prettier ./ --check",
-    "cy:quick": "cypress run --browser electron --config video=false,defaultCommandTimeout=10000",
-    "cy:single": "cypress run --browser electron --config video=false,defaultCommandTimeout=10000 --spec",
-    "cy:debug": "ELECTRON_ENABLE_LOGGING=1 cypress run --browser electron --config video=false,defaultCommandTimeout=10000",
+    "cy:run": "unset ELECTRON_RUN_AS_NODE && cypress run --headless --browser chrome",
+    "cy:open": "unset ELECTRON_RUN_AS_NODE && cypress open --browser chrome",
     "cy:spec": "unset ELECTRON_RUN_AS_NODE && npx cypress run --browser chrome --headless --spec",
```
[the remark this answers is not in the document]
pmanko: By design per CLAUDE.md and project requirements. `ELECTRON_RUN_AS_NODE=1` is set in our Claude Code CLI environment and breaks Cypress completely. All `cy:*` scripts must unset it. OpenELIS development is macOS/Linux only — there is no Windows dev environment support. The `unset` workaround is documented in CLAUDE.md under "Cypress E2E Test Execution (CRITICAL)".

frontend/cypress/pages/RoutineReportPage.js
```diff
@@ -1,110 +1,104 @@
...
+    this.clickNavLink("#menu_reports_aggregate_statistics");
   }
 
   allReportsSummary() {
-    cy.get("#menu_reports_aggregate_all_nav", { timeout: 15000 })
-      .scrollIntoView({ behavior: "smooth" })
-      .should("be.visible")
-      .click();
+    this.clickNavLink("#menu_reports_aggregate_all");
   }
```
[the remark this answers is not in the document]
pmanko: These selectors match actual DOM IDs rendered by the Carbon SideNav component (e.g., `#menu_reports_aggregate`). The `_nav` suffix pattern used in `HelpPage.js` targets a different component structure — leaf nav items within a sub-menu. The report page selectors here target section-level menu headers, which use different IDs. The selectors are correct and pass in CI E2E tests.

frontend/cypress/pages/RoutineReportPage.js
```diff
@@ -116,16 +110,15 @@ class RoutineReportPage {
...
 
   selectPatientStatusReport() {
-    cy.get("#menu_reports_status_patient", { timeout: 15000 })
-      .scrollIntoView({ behavior: "smooth" })
-      .should("be.visible")
-      .click();
+    this.ensureSidenavMenuExpanded("#menu_reports");
+    this.ensureSidenavMenuExpanded("#menu_reports_routine");
+    this.clickNavLink("#menu_reports_status_patient");
   }
```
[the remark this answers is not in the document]
pmanko: Same as above — these selectors target actual DOM IDs for section-level report menu headers, which are distinct from the `_nav` suffixed leaf items in `HelpPage.js`. Both conventions are correct for their respective component structures.

pmanko:
See madagascar.openelis-global.org for testing

analyzer-defaults/astm/genexpert-astm.json
```diff
@@ -0,0 +1,38 @@
+{
+  "schema_version": "1.0",
+  "analyzer_name": "Cepheid GeneXpert (ASTM Mode)",
+  "manufacturer": "Cepheid",
+  "category": "MOLECULAR",
+  "protocol": "ASTM",
+  "protocol_version": "LIS2-A2",
+  "identifier_pattern": "GENEXPERT|CEPHEID",
+  "transport": "RS-232 Serial or TCP/IP",
```
ibacher: Should this be structured? E.g.,

```json
{
  "transport": [
     "file",
     "RS-232",
     "TCP/IP"
  ]
}
```

analyzer-defaults/astm/genexpert-astm.json
```diff
@@ -0,0 +1,38 @@
...
+  "default_baud_rate": 9600,
```
ibacher: Then this sort of transport-specific config could be structured with the transport I think

analyzer-defaults/astm/genexpert-astm.json
```diff
@@ -0,0 +1,38 @@
+{
+  "schema_version": "1.0",
+  "analyzer_name": "Cepheid GeneXpert (ASTM Mode)",
+  "manufacturer": "Cepheid",
+  "category": "MOLECULAR",
+  "protocol": "ASTM",
+  "protocol_version": "LIS2-A2",
```
ibacher: Similar to the below:


```suggestion
  "protocol": {
    "name": "ASTM",
    "version": "LIS2-A2"
  },
```

projects/analyzer-defaults/astm/genexpert-astm.json
```diff
@@ -0,0 +1,38 @@
+{
+  "schema_version": "1.0",
+  "analyzer_name": "Cepheid GeneXpert (ASTM Mode)",
```
ibacher, line 3: Do we potentially need this to be localizeable?

analyzer-defaults/astm/genexpert-astm.json
```diff
@@ -0,0 +1,38 @@
+{
+  "schema_version": "1.0",
```
ibacher: Last thought on structuring things: there's a well supported way of handling JSON schemas, via a property called `$schema` on the object with a URL. So we could do something here like:

```json
"$schema: https://openelis-global.org/schemas/analyzer-defaults.json/1.0"
```

Which allows us to embed the same information (the schema version) but (later on) we can also publish a JSON schema at that URL go give some guidance to people actually editing the file (I did [something like this](https://github.com/openmrs/openmrs-contrib-json-schemas/blob/main/routes.schema.json) for OpenMRS, but never implemented a real versioning scheme).

projects/analyzer-harness/scripts/test-fluorocycler-simulator.sh
ibacher, line 1: I'm not sure if this is the right place for this... `scripts` seem to be mostly operational, not tests... I suppose we maybe need to put it somewhere, but it seems like it should just be part of the analyzer-mock-server?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
```diff
@@ -0,0 +1,316 @@
...
+    /**
+     * GET /rest/analyzer/errors
+     * 
+     * List analyzer errors with filtering and pagination
+     * 
+     * Query Parameters: - page, size, search, errorType, severity, status,
+     * analyzerId, startDate, endDate, sort
+     */
+    @GetMapping("/errors")
+    public ResponseEntity<Map<String, Object>> getErrors(@RequestParam(required = false) Integer page,
```
ibacher: I find the order of parameters here a little bizarre, though I know that's the most critical thing.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
```diff
@@ -0,0 +1,316 @@
...
+            @RequestParam(required = false) Integer size, @RequestParam(required = false) String search,
+            @RequestParam(required = false) String errorType, @RequestParam(required = false) String severity,
+            @RequestParam(required = false) String status, @RequestParam(required = false) String analyzerId,
+            @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME) Date startDate,
+            @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME) Date endDate,
+            @RequestParam(required = false) String sort) {
+        try {
+            // Parse enum values
+            AnalyzerError.ErrorType errorTypeEnum = errorType != null
```
ibacher: It seems like it should be possible to get Spring to do this conversion for you, i.e., by changing the type from `String` to, e.g., `AnalyzerError.ErrorType` using a [Formatter](https://docs.spring.io/spring-framework/reference/core/validation/format.html). That would somewhat simplify the code needed here.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
```diff
@@ -0,0 +1,316 @@
...
+                        if (error.getAnalyzer() != null && error.getAnalyzer().getName() != null
+                                && error.getAnalyzer().getName().toLowerCase().contains(searchLower)) {
+                            return true;
+                        }
+                    } catch (Exception e) {
+                        // If lazy loading fails, skip analyzer name check
+                    }
+                    return false;
+                }).collect(java.util.stream.Collectors.toList());
+            }
```
ibacher: Several comment here:

1. It's weird to me that we do essentially all the filtering here rather than pushing things down to the database.
2.  `getErrorsByFilters()` looks kidn of deceptively named as the actual implementation only processes a single filter
3. It doesn't seem like we actually do anything with the pagination parameters?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
```diff
@@ -0,0 +1,316 @@
...
+
+            // Calculate statistics - use current filtered results for now
+            // TODO: Implement proper statistics query in DAO
+            List<AnalyzerError> allErrors = errors; // Use filtered results for statistics
```
ibacher: Is this correct? I might expect the statistics to be based on all errors and not just a filtered subset.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
```diff
@@ -0,0 +1,316 @@
...
+            long totalErrors = allErrors.size();
+            long unacknowledged = allErrors.stream()
+                    .filter(e -> e.getStatus() == AnalyzerError.ErrorStatus.UNACKNOWLEDGED).count();
+            long critical = allErrors.stream().filter(e -> e.getSeverity() == AnalyzerError.Severity.CRITICAL).count();
+            long last24Hours = allErrors.stream().filter(e -> {
+                if (e.getLastupdated() == null)
+                    return false;
+                long hoursAgo = (System.currentTimeMillis() - e.getLastupdated().getTime()) / (1000 * 60 * 60);
```
ibacher: Nit:

```suggestion
                long hoursAgo = (System.currentTimeMillis() - e.getLastupdated().getTime()) / (1000L * 60L * 60L);
```

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
```diff
@@ -0,0 +1,316 @@
...
+    public ResponseEntity<Map<String, Object>> reprocessError(@PathVariable String id) {
+        try {
+            boolean success = analyzerErrorService.reprocessError(id);
+
+            Map<String, Object> response = new HashMap<>();
+            Map<String, Object> data = new HashMap<>();
+            data.put("success", success);
+            data.put("message", success ? "Message reprocessed successfully" : "Reprocessing failed");
+            response.put("data", data);
+            response.put("status", "success");
```
ibacher: It's a little weird to me to (potentially) get a message:

```json
{
	"data": {
     "success": false
   },
   "status": "success"
}
```

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
```diff
@@ -0,0 +1,316 @@
...
+     * 
+     * Get error by ID
+     */
+    @GetMapping("/errors/{id}")
+    public ResponseEntity<Map<String, Object>> getError(@PathVariable String id) {
+        try {
+            AnalyzerError error = analyzerErrorService.getErrorById(id);
+
+            if (error == null) {
+                Map<String, Object> errorResponse = new HashMap<>();
```
ibacher: I may be out of sync with OE on this, but I think `LinkedHashMaps` are preferable for building JSON objects... The difference being that the keys are ordered by insertion order rather than hash code order.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+     * accesses relationships
+     * 
+     * Task Reference: T200 - Enhanced to support includeRetired parameter
+     */
+    @GetMapping("/analyzers/{analyzerId}/mappings")
+    public ResponseEntity<List<Map<String, Object>>> getMappings(@PathVariable String analyzerId,
+            @RequestParam(defaultValue = "false") boolean includeRetired) {
+        try {
+            // Service layer eagerly fetches all relationships and compiles complete data
+            List<Map<String, Object>> response = analyzerFieldMappingService.getMappingsForAnalyzer(analyzerId,
```
ibacher, line 79: Shouldn't this be wrapped in the usual "status" / "data" object?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+            @Valid @RequestBody AnalyzerFieldMappingForm form) {
+        try {
+            // Service layer verifies analyzer field belongs to analyzer, validates type
+            // compatibility,
+            // and returns complete compiled data - controller never accesses relationships
+            Map<String, Object> response = analyzerFieldMappingService.createMappingForAnalyzer(analyzerId,
+                    form.getAnalyzerFieldId(), form.getOpenelisFieldId(), form.getOpenelisFieldType(),
+                    form.getMappingType(), form.getIsRequired(), form.getIsActive(), form.getSpecimenTypeConstraint(),
+                    form.getPanelConstraint());
+            return ResponseEntity.status(HttpStatus.CREATED).body(response);
```
ibacher, line 103: Ditto about the response body.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+    public ResponseEntity<Map<String, Object>> activateMappings(@PathVariable String id,
+            @RequestBody Map<String, Object> request) {
+        try {
+            // Validate activation requirements
+            ActivationValidationResult validationResult = analyzerFieldMappingService.validateActivation(id);
+
+            // Check if activation can proceed
+            if (!validationResult.isCanActivate()) {
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Cannot activate: Required mappings missing");
```
ibacher: Shouldn't the error message here come from the `ActivationValidationResult`? Also seems like something we'd want to be able to translate eventually.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+
+            // Get mapping IDs from request (optional - if not provided, activate all draft
+            // mappings)
+            @SuppressWarnings("unchecked")
+            List<String> mappingIds = (List<String>) request.get("mappingIds");
+            String confirmationToken = (String) request.get("confirmationToken");
+
+            // If mappingIds provided, activate specific mappings
+            int activatedCount = 0;
+            if (mappingIds != null && !mappingIds.isEmpty()) {
```
ibacher, line 229: nit: The `isEmpty()` here isn't strictly necessary.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+                    } catch (LIMSRuntimeException e) {
+                        // Check for optimistic locking exception (T168a)
+                        if (e.getCause() instanceof org.hibernate.StaleObjectStateException) {
+                            logger.warn("Concurrent edit detected for mapping " + mappingId);
+                            Map<String, Object> error = new HashMap<>();
+                            error.put("error", "Concurrent edit detected");
+                            error.put("message", "Mapping was modified by another user. Please reload and try again.");
+                            error.put("mappingId", mappingId);
+                            return ResponseEntity.status(HttpStatus.CONFLICT).body(error);
+                        }
```
ibacher: Two worries here:

1. If the mapping doesn't belong to the analyzer that doesn't seem to get reported
2. If a concurrent edit happens, it looks like this just sends back the error, but it seems like it's possible that for some mappings to have already been activated when that happens.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+                }
+            }
+
+            // Build response
+            Map<String, Object> response = new HashMap<>();
+            response.put("activatedCount", activatedCount);
+            response.put("warnings", validationResult.getWarnings());
+            response.put("requiresAdditionalConfirmation", validationResult.getPendingMessagesCount() > 0);
+
+            return ResponseEntity.ok(response);
```
ibacher, line 239: Similar remark about "standard" REST responses included "success" and "data".

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+     * true), skipIncompatible: Boolean (default false) }
+     */
+    @PostMapping("/analyzers/{targetId}/copy-mappings")
+    public ResponseEntity<Map<String, Object>> copyMappings(@PathVariable String targetId,
+            @RequestBody Map<String, Object> request) {
+        try {
+            if (analyzerMappingCopyService == null) {
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Copy mappings service not available");
+                return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(error);
```
ibacher: Minor: This seems more or less like it should be a 404 scenario, i.e., the instance isn't configured to handle this kind of request.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Mapping preview service not available");
+                return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(error);
+            }
+
+            // Validate message size
+            if (form.getAstmMessage() == null || form.getAstmMessage().length() > 10 * 1024) {
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "ASTM message exceeds maximum size of 10KB");
+                return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error);
```
ibacher: 413 seems like the right response here.

ibacher: Just to point this out, but if we only have the size filter here, we've actually done a fair bit of processing around the message, since the request body has been parsed and validated...

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
ibacher: This basically applies to every method in this class.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
```diff
@@ -0,0 +1,582 @@
...
+            Map<String, Object> response = analyzerFieldMappingService.createMappingForAnalyzer(analyzerId,
+                    form.getAnalyzerFieldId(), form.getOpenelisFieldId(), form.getOpenelisFieldType(),
+                    form.getMappingType(), form.getIsRequired(), form.getIsActive(), form.getSpecimenTypeConstraint(),
+                    form.getPanelConstraint());
+            return ResponseEntity.status(HttpStatus.CREATED).body(response);
+        } catch (LIMSRuntimeException e) {
+            logger.error("Error creating mapping: " + e.getMessage(), e);
+            Map<String, Object> error = new HashMap<>();
+            error.put("error", e.getMessage());
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error);
```
ibacher: Are these definitely the result of bad requests?

ibacher: This also applies to pretty much every method in this class.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    /**
+     * GET /rest/analyzer/analyzers Retrieve all analyzers with their configurations
+     */
+    @GetMapping("/analyzers")
+    public ResponseEntity<List<Map<String, Object>>> getAnalyzers(@RequestParam(required = false) String status,
+            @RequestParam(required = false) String search) {
+        try {
+            List<Analyzer> analyzers = analyzerService.getAll();
```
ibacher, line 90: This feels like logic that should be pushed into the DB layer and not handled in a REST controller.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+    public ResponseEntity<Map<String, Object>> createAnalyzer(@RequestBody AnalyzerForm form,
+            HttpServletRequest request) {
+        try {
+            // Manual validation for optional fields
+            if (form.getIpAddress() != null && !form.getIpAddress().matches("^(\\d{1,3}\\.){3}\\d{1,3}$")) {
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Invalid IPv4 address format");
+                return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error);
+            }
+            if (form.getIpAddress() != null && NetworkValidationUtil.isBlockedAddress(form.getIpAddress())) {
```
ibacher, line 153: `form.getIpAddress() != null` is redundant at this point.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+    }
+
+    /**
+     * POST /rest/analyzer/analyzers Create new analyzer with configuration
+     */
+    @PostMapping("/analyzers")
+    public ResponseEntity<Map<String, Object>> createAnalyzer(@RequestBody AnalyzerForm form,
+            HttpServletRequest request) {
+        try {
+            // Manual validation for optional fields
```
ibacher: Might be a good idea to run the validations independently so that multiple issues can be reported in one go?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+
+            String analyzerId = analyzerService.insert(analyzer);
+
+            // Retrieve created analyzer
+            Analyzer createdAnalyzer = analyzerService.get(analyzerId);
+            if (createdAnalyzer == null) {
+                throw new LIMSRuntimeException("Failed to retrieve created analyzer");
+            }
+
+            // Create AnalyzerConfiguration with unified status
```
ibacher: I don't understand what "with unified status" actually means which makes this comment not particularly useful, I think.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+            // Always create configuration to store status, even if IP/Port not provided
+            try {
+                List<String> testUnitIds = form.getTestUnitIds() != null ? form.getTestUnitIds() : new ArrayList<>();
+                AnalyzerConfiguration config = new AnalyzerConfiguration();
+                config.setAnalyzer(createdAnalyzer);
+                config.setIpAddress(form.getIpAddress());
+                config.setPort(form.getPort());
+                config.setProtocolVersion(
+                        form.getProtocolVersion() != null ? form.getProtocolVersion() : "ASTM LIS2-A2");
```
ibacher: Maybe this is ok, but in the JSON wasn't this represented as just "LIS2-A2"?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+                    config.setStatus(AnalyzerConfiguration.AnalyzerStatus.valueOf(status));
+                } catch (IllegalArgumentException e) {
+                    logger.warn("Invalid status value: " + status + ", defaulting to SETUP");
+                    config.setStatus(AnalyzerConfiguration.AnalyzerStatus.SETUP);
+                }
+                config.setSysUserId(getSysUserId(request));
+                analyzerConfigurationService.insert(config);
+            } catch (LIMSRuntimeException e) {
+                // If configuration creation fails, log but don't fail the analyzer creation
+                logger.warn("Failed to create analyzer configuration: " + e.getMessage());
```
ibacher: Minor: generally preferring logging messages with parameters to string concatenation (applies to other messages in this file)

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Analyzer not found: " + id);
+                return ResponseEntity.status(HttpStatus.NOT_FOUND).body(error);
+            }
+            Optional<AnalyzerConfiguration> configOpt = analyzerConfigurationService.getByAnalyzerId(id);
+
+            if (!configOpt.isPresent()) {
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Analyzer configuration not found");
+                return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error);
```
ibacher: Not sure this is a bad request.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+        }
+    }
+
+    /**
+     * GET /rest/analyzer/analyzers/{id} Retrieve analyzer by ID
+     */
+    @GetMapping("/analyzers/{id}")
+    public ResponseEntity<Map<String, Object>> getAnalyzer(@PathVariable String id) {
+        try {
+            Analyzer analyzer = analyzerService.get(id);
```
ibacher, line 319: I'm confused about the contract on this method. In most places, we seem to assume that it returns `null` if the object doesn't exist, but here we expect an `ObjectNotFoundException`? Either the code here or the code in other methods in the file are incorrect.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+    }
+
+    /**
+     * DELETE /rest/analyzer/analyzers/{id} Soft delete analyzer
+     * 
+     * Implements soft delete: sets status to INACTIVE (analyzer can be
+     * reactivated).
+     * 
+     * @param id Analyzer ID to delete
+     * @return 204 No Content on success, 404 if analyzer not found, 500 on error
```
ibacher: Should this also implement the logic for "recent results" from the "legacy" endpoint?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+
+        try {
+            // Attempt TCP connection with 5 second timeout
+            socket = new Socket();
+            socket.connect(new java.net.InetSocketAddress(ipAddress, port), 5000);
+            socket.setSoTimeout(5000); // Read timeout
+
+            // ASTM LIS2-A2 Control Characters
+            byte ENQ = 0x05; // Enquiry - Start transmission
+            byte ACK = 0x06; // Acknowledge - Positive response
```
ibacher: These are constants... shouldn't they be treated as such?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+            return response;
+        }
+
+        try {
+            // Attempt TCP connection with 5 second timeout
+            socket = new Socket();
+            socket.connect(new java.net.InetSocketAddress(ipAddress, port), 5000);
+            socket.setSoTimeout(5000); // Read timeout
+
+            // ASTM LIS2-A2 Control Characters
```
ibacher: The method seems a little incorrectly named since it's ASTM-only so we can't use this to check, e.g., HL7 over TCP/IP.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+            String portName = serialConfig.getPortName();
+
+            if (portName == null || portName.trim().isEmpty()) {
+                response.put("success", false);
+                response.put("message", "Serial port name not configured");
+                return response;
+            }
+
+            // Check if port exists (file descriptor check)
+            java.io.File portFile = new java.io.File(portName);
```
ibacher: It's probably important to note that this assumes *NIX, which is fine, but also means that connections over serial ports are only workable on Linux-based installs, which feels like it belongs in the documentation.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+        return response;
+    }
+
+    /**
+     * POST /rest/analyzer/analyzers/{id}/query Start an asynchronous query job for
+     * an analyzer
+     */
+    @PostMapping("/analyzers/{id}/query")
+    public ResponseEntity<Map<String, Object>> queryAnalyzer(@PathVariable String id) {
+        try {
```
ibacher, line 785: Shouldn't this do some sort of check for push-only analyzers?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+     *
+     * <p>
+     * Task Reference: T216 (M20) - Create REST endpoint for loading specific
+     * template
+     */
+    @GetMapping("/defaults/{protocol}/{name}")
+    public ResponseEntity<?> getDefaultConfig(@PathVariable String protocol, @PathVariable String name) {
+        try {
+            // Validate protocol (allowlist)
+            if (!protocol.equals("astm") && !protocol.equals("hl7")) {
```
ibacher: Should these be case insensitive?

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Invalid protocol: must be 'astm' or 'hl7'");
+                return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error);
+            }
+
+            // Sanitize filename: only alphanumeric, dash, underscore, period
+            if (!name.matches("^[a-zA-Z0-9\\-_.]+$")) {
```
ibacher, line 896: This also seems like it should be better documented. The rule here is substantially stricter than is implied in the method comment for example.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+            }
+
+            // Ensure .json extension
+            String filename = name.endsWith(".json") ? name : name + ".json";
+
+            // Build path
+            String defaultsDir = System.getenv("ANALYZER_DEFAULTS_DIR");
+            if (defaultsDir == null || defaultsDir.isEmpty()) {
+                defaultsDir = "/data/analyzer-defaults";
+            }
```
ibacher, line 909: I've seen this twice now. This should be refactored into a method... Which would also make sense because I'm not sure the REST controller should be doing this much filesystem work.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+
+            java.io.File baseDir = new java.io.File(defaultsDir);
+            java.io.File templateFile = new java.io.File(baseDir, protocol + "/" + filename);
```
ibacher: Hard-coding the `/` here seems off to me.

ibacher: Especially because we use the right constant six lines down.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+            // Verify canonical path (prevents path traversal)
+            String canonicalPath = templateFile.getCanonicalPath();
+            String baseDirCanonical = baseDir.getCanonicalPath();
+
+            if (!canonicalPath.startsWith(baseDirCanonical + java.io.File.separator)
+                    && !canonicalPath.equals(baseDirCanonical)) {
+                Map<String, Object> error = new HashMap<>();
+                error.put("error", "Invalid path: template must be within defaults directory");
+                return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error);
+            }
```
ibacher: I feel like the Java Paths API is generally better for these kinds of checks than the File API.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+        } catch (IOException e) {
+            logger.error("Error reading default config: " + protocol + "/" + name, e);
+            Map<String, Object> error = new HashMap<>();
+            error.put("error", "Failed to read template: " + e.getMessage());
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(error);
+        } catch (Exception e) {
+            logger.error("Error loading default config", e);
+            Map<String, Object> error = new HashMap<>();
+            error.put("error", "Failed to load template: " + e.getMessage());
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(error);
```
ibacher: It feels like the logger methods here should store the same amount of context. I'm actually kind of struggling to understand why we have two `catch` clauses. The difference between them is jsut the first logging message and that seems relevant for general exceptions as well.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
```diff
@@ -0,0 +1,1040 @@
...
+    private void scanTemplates(java.io.File directory, String protocol, List<Map<String, Object>> templates) {
+        java.io.File[] files = directory.listFiles((dir, name) -> name.endsWith(".json"));
+        if (files == null) {
+            return;
+        }
+
+        for (java.io.File file : files) {
+            try {
+                // Read JSON to extract analyzer_name
+                String jsonContent = new String(java.nio.file.Files.readAllBytes(file.toPath()),
```
ibacher: Here I'm left questioning why we we're passing around `File` objects at all, just to drop to using the `java.nio.file.*` stuff.

ibacher (commented):
This is a pretty huge PR, so I dug through what I could in a reasonable amount of time. I think there's quite a bit of clean-up probably necessary here, but I won't block merging this.

analyzer-defaults/astm/genexpert-astm.json
pmanko: Fixed — added `$schema` URL property to all 11 analyzer-defaults JSON files.

analyzer-defaults/astm/genexpert-astm.json
pmanko: Fixed — `protocol` is now a structured object: `{"name": "ASTM", "version": "LIS2-A2"}`. Separate fields for protocol family and version.

analyzer-defaults/astm/genexpert-astm.json
pmanko: Fixed — `transport` is now a typed array: `["RS-232", "TCP/IP"]`. Machine-readable instead of free-text.

analyzer-defaults/astm/genexpert-astm.json
```diff
@@ -0,0 +1,38 @@
...
+  "analyzer_name": "Cepheid GeneXpert (ASTM Mode)",
```
pmanko: No — these are seed defaults for initial DB provisioning, not user-facing labels. Display names go through React Intl.

analyzer-defaults/astm/genexpert-astm.json
pmanko: Fixed — added `transport_config` object keyed by transport type. Transport-specific settings (baud rate, port, framing) are grouped under the correct transport.

projects/analyzer-harness/scripts/test-fluorocycler-simulator.sh
pmanko: Agreed — moved to `projects/analyzer-harness/scripts/`. It's test infrastructure, not operational.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
pmanko: Fixed — reordered to: filters first (analyzerId, errorType, severity, status, startDate, endDate), then search, then pagination (page, size), then sort.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
pmanko: Fixed — changed `@RequestParam` types from `String` to direct enum types (`AnalyzerError.ErrorType`, etc.). Spring's built-in enum converter handles the conversion automatically.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
pmanko: Fixed — filtering pushed to DAO layer via `AnalyzerErrorDAO.findByFilters()` with dynamic HQL. All non-null params combined using AND logic in a single query.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
pmanko: Good catch — fixed. Statistics are now global (independent of filters) via `AnalyzerErrorDAO.getGlobalStatistics()` — single efficient aggregate query returning totalErrors, unacknowledged, critical, and last24Hours counts.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
pmanko: Fixed — `1000 * 60 * 60` → `1000L * 60L * 60L` to prevent int overflow.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
pmanko: Fixed — failed reprocessing now returns 422 UNPROCESSABLE_ENTITY instead of 200 with contradictory `success: false` inside `status: "success"`.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerErrorRestController.java
pmanko: Fixed — all `HashMap` → `LinkedHashMap` across all 3 analyzer controllers for predictable JSON key ordering.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — all methods now use standardized `{status, data}` envelope via `AnalyzerControllerHelper.wrapResponse()`/`wrapError()`.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — applied the `{status, data}` envelope pattern consistently across all methods in this controller.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — now uses `validationResult.getMissingRequired()` in the error message instead of a hardcoded string.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — removed the redundant `!mappingIds.isEmpty()` check.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — both issues addressed: (1) Activation moved to `AnalyzerFieldMappingService.bulkActivateMappings()` in a single `@Transactional` boundary (all-or-nothing). (2) Ownership violations now throw immediately instead of silently skipping.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — uses the standardized response envelope pattern.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — unavailable optional services now return 404 instead of 503.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — returns 413 PAYLOAD_TOO_LARGE. Also moved the size check before form parsing/validation.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — all `LIMSRuntimeException` handlers now use `AnalyzerControllerHelper.mapExceptionToResponse()` which inspects the cause: IllegalArgument→400, StaleObjectState→409, ConstraintViolation→409, other→500.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — applied the `wrapResponse()`/`wrapError()`/`mapExceptionToResponse()` pattern to all methods.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — `mapExceptionToResponse()` now differentiates HTTP status codes by cause across all methods.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerFieldMappingRestController.java
pmanko: Fixed — moved the size check before service availability check and form parsing.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Acknowledged — valid optimization target for a follow-up. The admin endpoint handles <100 analyzers, making in-memory filtering acceptable for now. Higher-priority fixes took precedence in this batch.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Addressed — the null check is now integral to the collected-validation-errors pattern (protects against NPE in `matches()` and `isBlockedAddress()`).

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — `createAnalyzer` now collects all validation errors into a list and returns them together in a single 400 response with a `validationErrors` array.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — all 'unified status' comments rewritten to explain the lifecycle: `SETUP → ACTIVE → INACTIVE → DELETED`.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — default changed from `"ASTM LIS2-A2"` to `"LIS2-A2"` in the entity, controller defaults, and Liquibase column definition (new changeset 026).

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — all string concatenation in logger calls converted to `{}` parameter placeholders across all 3 controllers.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — 'Analyzer configuration not found' now returns 422 UNPROCESSABLE_ENTITY (analyzer exists but isn't configured — precondition failure, not bad request).

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — replaced dead `catch (ObjectNotFoundException)` with `if (analyzer == null)` null check, matching every other method in the controller. The `get()` method returns null, never throws.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — removed the duplicate `DELETE /analyzers/{id}` endpoint. The POST legacy endpoint (`/analyzers/{id}/delete`) with 90-day logic is the canonical one used by the UI.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — `ENQ` and `ACK` are now `private static final byte` constants at class level with Javadoc.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — renamed to `testAstmTcpConnection()` to clarify it's ASTM-specific over TCP/IP.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — added platform note Javadoc explaining *NIX device node dependency (`/dev/ttyS0`, `/dev/ttyUSB0`) and that Windows would require javax.comm or jSerialComm.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — `AnalyzerQueryService.startQuery()` now validates the protocol before creating the background job. Push-only analyzers (FILE, RS-232, SERIAL) get a clear error message. Controller returns 422.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — protocol check now uses `equalsIgnoreCase()`.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — improved Javadoc to document the exact regex (`^[a-zA-Z0-9\-_.]+$`), its purpose (path traversal prevention), and the normalized path verification.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — refactored to use `java.nio.file.Path` API throughout. Shared logic now uses `Path.resolve()`, `Files.readString()`, `Files.list()` with try-with-resources, and `Path.normalize()` for path traversal verification.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — replaced all hardcoded `"/"` separators with `Path.resolve()` which handles platform separators automatically.

pmanko: Fixed — no more mixed usage. Everything is Path-based now.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — converted all `java.io.File` usage to `java.nio.file.Path` in the defaults/template methods.

src/main/java/org/openelisglobal/analyzer/controller/AnalyzerRestController.java
pmanko: Fixed — the `scanTemplates` method now uses `Files.list()` with a single unified `catch (IOException)` clause. The old two-clause pattern (`IOException` vs generic `Exception`) is eliminated.

projects/analyzer-defaults/astm/genexpert-astm.json
pmanko: `analyzer_name` is a product name (not localized). Renamed `test_name` → `test_name_hint` in all defaults — these are developer-facing fallbacks; the system resolves display names from LOINC codes at runtime.

frontend/src/components/analyzers/DeleteAnalyzerModal/DeleteAnalyzerModal.jsx
```diff
@@ -0,0 +1,137 @@
...
+  };
+
+  const analyzerName =
+    analyzer?.name || intl.formatMessage({ id: "analyzer.delete.unknown" });
+
+  return (
+    <ComposedModal
+      open={open}
+      onClose={onClose}
+      danger
```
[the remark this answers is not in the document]
pmanko: Fixed — added `preventCloseOnClickOutside={isDeleting}` to prevent ESC/click-outside closure during active delete. See 64cb7ca.

frontend/src/components/analyzers/FieldMapping/OpenELISFieldSelector.jsx
```diff
@@ -0,0 +1,147 @@
...
+ * Task Reference: T061
+ *
+ * Features:
+ * - Searchable dropdown
+ * - Category filtering (8 entity types)
+ * - Type compatibility filtering
+ */
+
+import React, { useState } from "react";
+import { ComboBox, Tag, Button } from "@carbon/react";
```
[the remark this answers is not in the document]
pmanko: Fixed — removed unused `Tag` import. See 64cb7ca.

frontend/src/components/analyzers/FieldMapping/MappingRetirementModal.jsx
```diff
@@ -0,0 +1,171 @@
...
+import {
+  ComposedModal,
+  ModalHeader,
+  ModalBody,
+  ModalFooter,
+  Button,
+  TextArea,
+  InlineNotification,
+} from "@carbon/react";
+import { WarningAlt } from "@carbon/icons-react";
```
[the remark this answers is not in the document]
pmanko: Fixed — removed unused `WarningAlt` import. See 64cb7ca.

frontend/src/components/analyzers/FieldMapping/MappingActivationModal.jsx
```diff
@@ -0,0 +1,232 @@
...
+  ComposedModal,
+  ModalHeader,
+  ModalBody,
+  ModalFooter,
+  Button,
+  Checkbox,
+  InlineNotification,
+  Link,
+} from "@carbon/react";
+import { WarningAlt, Error } from "@carbon/icons-react";
```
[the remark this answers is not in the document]
pmanko: Fixed — removed unused `Error` import. See 64cb7ca.

frontend/src/components/analyzers/FileImportConfiguration/__tests__/FileImportConfiguration.test.jsx
```diff
@@ -0,0 +1,242 @@
...
+jest.mock("../../../../services/analyzerService", () => ({
+  getAnalyzers: jest.fn(),
+}));
+
+// ========== IMPORTS ==========
+
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import { waitFor } from "@testing-library/dom";
+import "@testing-library/jest-dom/extend-expect";
```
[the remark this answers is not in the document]
pmanko: Fixed — updated to `@testing-library/jest-dom` (removed `/extend-expect` suffix). See 64cb7ca.

frontend/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # pull official base image
-FROM node:20.1.0-alpine
+FROM node:20-alpine
```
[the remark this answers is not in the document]
pmanko: Fixed — pinned to `node:20.20-alpine`. See 64cb7ca.

frontend/Dockerfile.prod
```diff
@@ -1,5 +1,5 @@
 # build env
-FROM node:20.1.0-alpine AS build
+FROM node:20-alpine AS build
```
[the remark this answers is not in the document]
pmanko: Fixed — pinned to `node:20.20-alpine`. See 64cb7ca.

frontend/.prettierignore
```diff
@@ -14,5 +14,10 @@ coverage/
 reports/
 **/*.min.js
 *.log
-
+# Playwright test artifacts (generated)
+playwright-report/
+playwright/.auth/
+test-results/
+# Stray Playwright test output at frontend root only (playwright/tests/*.ts are formatted)
+/sidenav.spec.ts
```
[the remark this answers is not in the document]
pmanko: Fixed — removed stale `/sidenav.spec.ts` entry. See 64cb7ca.

frontend/src/components/analyzers/FieldMapping/ValidationDashboard.jsx
```diff
@@ -0,0 +1,318 @@
...
+            <div className="unmapped-fields-section">
+              <h4>
+                <FormattedMessage id="validation.unmapped.fields.title" />
+              </h4>
+              <div className="unmapped-fields-list">
+                {metrics.unmappedFields.map((field, index) => (
+                  <Tag key={index} type="red" size="sm">
+                    {field}
+                  </Tag>
+                ))}
```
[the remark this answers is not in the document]
pmanko: Fixed — replaced array index keys with stable field/warning string values. See 64cb7ca.

frontend/cypress.config.js
```diff
@@ -1,18 +1,106 @@
...
+  if (!cypressUsername || !cypressPassword) {
+    throw new Error(
+      "In CI, CYPRESS_USERNAME/CYPRESS_PASSWORD or TEST_USER/TEST_PASS must be set for E2E tests.",
+    );
+  }
+} else {
+  cypressUsername =
+    process.env.CYPRESS_USERNAME || process.env.TEST_USER || "admin";
+  cypressPassword =
+    process.env.CYPRESS_PASSWORD || process.env.TEST_PASS || "adminADMIN!";
```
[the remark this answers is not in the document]
pmanko: Intentional — these are standard OpenELIS dev defaults matching docker-compose. CI already requires env vars (throws if missing). Local fallbacks avoid requiring env setup for every developer.

mozzy11 (approved):

mozzy11:
Thanks @pmanko

Issue #2270, opened 2025-10-23
Improve README to guide new contributors


Agaba-derrick:
Add a small guide in  the readme  directing new contributors  to our wiki ( https://uwdigi.atlassian.net/wiki/spaces/OG/pages/240844805/Dev+Environment+Setup+Instructions)

### Contributing

We welcome community contributions to help improve OpenELIS Global!

If you're new and curious , here is a link to our Wiki ( https://uwdigi.atlassian.net/wiki/spaces/OG/pages/240844805/Dev+Environment+Setup+Instructions)

note : This helps first-time contributors understand where to begin and improves onboarding clarity.

Agaba-derrick:
this is a good first time Issue !

mukulpythondev:
Hi @Agaba-derrick 

I’d love to work on this issue! Now that I’m familiar with the project structure and setup, I can add a short guide in the README directing new contributors to the Wiki for environment setup.  

This will make onboarding smoother and help first-time contributors get started quickly.

Agaba-derrick:
@mukulpythondev  can you first solve the comments in  Pr #2263

Agaba-derrick:
@reagan-meant  are we able to add Good first Issues badge on  this for new contrbibutors ?

SumritSingh15:
Hi @Agaba-derrick 
Subject:  Add contributor onboarding guide link to README.
I’ve been exploring the OpenELIS Global repository and noticed that while the setup wiki is well documented, new contributors might not immediately find it from the main README.

I’d like to add a small “Contributing” section (or enhance the existing one) with a short paragraph directing new contributors to the Wiki here:
https://uwdigi.atlassian.net/wiki/spaces/OG/pages/240844805/Dev+Environment+Setup+Instructions

This will make onboarding easier for first-time contributors and improve clarity.

Would you like me to go ahead and open a pull request with this change?

Thanks,
Sumrit Singh

Roshan-Pawar:
assign this to me, I have started with project setup

Roshan-Pawar:
@Agaba-derrick please review

vidishadeswal:
Hey @Agaba-derrick,  I have made a development setup guide for this project and a summary of how everything works, so I'd be glad if you assign this issue to me if its still open

Agaba-derrick:
@vidishadeswal  this issue is assigned to @Roshan-Pawar  and there is an existing Pr for it, kindly pick another issue to work upon

Issue #2770, opened 2026-02-08
Inventory Management: “Add New Lot” → Storage Location Dropdown Not Working and “Add Storage Location” Form Inputs Not Editable


AnuragSingh2101:
### Overview

There is a UI interaction issue in the **Inventory Management** workflow when adding a new inventory lot from the Dashboard.  
The problem occurs in the **“Receive New Inventory Lot”** screen and prevents users from selecting or creating a storage location.

This blocks the entire **Add New Lot** flow.

---

### Navigation Flow

1. Go to **Dashboard**
2. Navigate to **Inventory Management**
3. Click the **“Add New Lot”** button
4. The **“Receive New Inventory Lot”** modal/page opens

---

#### Select Storage Location dropdown is not working

- In the **Receive New Inventory Lot** screen, the **“Select Storage Location”** dropdown:
  - Does not open when clicked
  - Does not allow selecting any existing storage location
  - Appears non-interactive

#### Add Storage Location modal inputs are not editable

- Clicking **“Add Storage Location”** opens the **Add Storage Location** modal
- However:
  - Input fields do not accept any typing
  - Fields remain stuck showing placeholder text
  - NumberInput increment/decrement buttons also do not work
- As a result, a new storage location cannot be created

---

### Observed Behavior

- Storage location dropdown cannot be used
- Add Storage Location modal opens but behaves like it is read-only
- Users are unable to proceed with receiving a new inventory lot

---

### Expected Behavior

- **Select Storage Location** dropdown should open and allow selecting an existing location
- **Add Storage Location** modal should allow typing into all input fields
- Newly created storage locations should be selectable in the dropdown
- Users should be able to successfully complete the **Receive New Inventory Lot** workflow

---

### Additional Notes / Suspected Cause

- Issue occurs when **Add Storage Location** is opened from within **Receive New Inventory Lot**
- Likely related to modal interaction and focus handling
- Nested or overlapping Carbon `Modal` components may be causing focus traps and blocking user input

---

https://github.com/user-attachments/assets/d44ad87b-4d51-4c18-a63d-f28d99260964

AnuragSingh2101:
@pmanko, I found this issue. is it valid?
Currently I'm working on this issue. If I'm able to solve this issue, I will raise a PR. Does it sounds good?

libinfaby:
@AnuragSingh2101 @pmanko 

[...]

PR #2569, opened 2026-01-18
Solved #2568 : Fix crushed sidebar filters on resize

shreyaakalra:
### Description
Fixes #2568

This PR addresses a UI regression where the Notebook Dashboard sidebar and filter inputs would become crushed and unusable when resizing the browser window or viewing on smaller screens.

### Changes Made
1. **Layout Architecture:**
   - Replaced the rigid outer `Grid` with a responsive Flexbox container.
   - Added `min-width: 280px` to the sidebar to prevent it from shrinking beyond usability.
   - Added `min-width: 0` to the main content area to ensure proper shrinking without overflow.
   - Implemented a mobile breakpoint (max-width: 768px) to stack the sidebar vertically on small devices.

2. **Filter Components:**
   - Increased the Carbon Grid column span for filter inputs from `lg={2}` to `lg={4}`. This ensures dropdowns maintain a readable width (approx 250px) and wrap to the next line instead of collapsing into a thin strip.

### Screenshots
**Before (Bug):**
Screenshot 2026-01-19 at 4 57 10 AM

**After (Fix):**
Screenshot 2026-01-19 at 4 54 01 AM

### Testing
- Verified that the sidebar no longer collapses on resize.
- Verified that filter inputs wrap correctly and remain clickable.
- Verified that the layout stacks vertically on mobile view.

PR #2780, opened 2026-02-09
Inventory Management: Nested modal issue blocking storage location creation in Add New Lot

AnuragSingh2101:
### Description

This PR fixes a critical UI issue in the **Inventory Management → Add New Lot** workflow where users were unable to select or create a storage location.

When opening **Add Storage Location** from within **Receive New Inventory Lot**, the modal inputs were not editable and the **Select Storage Location** dropdown was non-interactive. This behavior blocked the entire lot creation process.

The issue was caused by **nested Carbon Modal components**, which resulted in focus traps and blocked user input.  
The fix ensures proper modal handling so that only one modal is active at a time, restoring full interactivity.

### Navigation Flow
1. Go to **Dashboard**
2. Navigate to **Inventory Management**
3. Click **Add New Lot**
4. The **Receive New Inventory Lot** modal opens

### Observed Behavior
- **Select Storage Location** dropdown does not open or allow selection
- Clicking **Add Storage Location** opens a modal where:
  - Text inputs do not accept typing
  - NumberInput increment/decrement buttons do not work
  - Fields remain stuck showing placeholder text
  -  Users cannot proceed with receiving a new inventory lot

### Expected Behavior
- The storage location dropdown should open and allow selection
- Add Storage Location modal should allow typing into all inputs
- Newly created locations should appear in the dropdown
- Users should be able to complete the Add New Lot workflow

### Root Cause
- Nested or overlapping **Carbon Modal** components caused focus trapping
- Input events were blocked when one modal was rendered inside another

---
Fixes **#2270**

https://github.com/user-attachments/assets/1cdb608e-5c52-413a-b923-1801dd5d88c2


pmanko (changes requested):
Hey @AnuragSingh2101 — good call fixing the nested modal focus-trap, the sibling-modal approach is the right direction. A few things need addressing though:

### Must fix

**State loss bug:** The "Add New Lot" button still calls `setLotModalOpen(true)`, so LotEntryModal renders via the legacy path. After creating a storage location, `activeModal` switches to `"LOT"` which mounts a *different* LotEntryModal instance — wiping whatever the user had filled in. The dual system (`lotModalOpen` + `activeModal`) needs to be collapsed into one. Either fully migrate to `activeModal` and drop `lotModalOpen`, or don't unmount LotEntryModal during the storage location flow.

**Unrelated commit:** `f5ae5985` (sidebar filter fix from #2569) is squashed into this branch — please rebase it out.

**Issue link typo:** "Fixes #2270" should be "Fixes #2770" (transposed digits).

**Hardcoded string:** `"Storage location created successfully"` in `handleLocationCreated` needs `intl.formatMessage()`.

### Smaller things

- `handleLocationCreated` sets `activeModal(null)`, then the inline `onSave` callback immediately sets it to `"LOT"` — contradictory, works by accident via React batching. Pick one place to own the transition.
- `LogEvent.logError("InventoryStorageLocation", e, false)` — the first arg is `errorMessage` per the API, not a class name. The rest of the file uses `LogEvent.logError(e)`.

### fhir_uuid gap — worth fixing here?

Your `InventoryStorageLocationServiceImpl` fix is solid. The same bug exists in `InventoryItemServiceImpl` and `InventoryLotServiceImpl` — same `@Column(nullable = false, unique = true)` on `fhirUuid`, no `@PrePersist`, no service-level generation. Would you be up for adding the same pattern to those two services in this PR? Identical change, and it'd close the gap for the whole inventory module.

mozzy11:
Thanks for identifying the nested Carbon Modal focus trap in the Add New Lot flow — that was a real blocker. Closing because current develop already implements the same remedy: LotEntryModal now renders StorageLocationModal as a sibling and suspends itself while the location modal is active, so the workflow is unblocked. If you find any remaining rough edges in the lot creation flow on latest develop, a fresh PR would be very welcome.

PR #2859, opened 2026-02-17
fix(validation): add @NotNull validation for ExternalConnection URI

taniy8:
## Summary
Added @NotNull validation for the ExternalConnection URI field to prevent null values and improve backend data validation.

## Screenshots
N/A – backend validation only

## Related Issue
N/A

## Other
- Backend-only change
- No UI impact
- Manually tested


mozzy11:
Thanks for the contribution. Closing during a triage pass of older PRs: this one-line @NotNull could break saves of existing ExternalConnection rows that legitimately have no URI yet (e.g., connections configured in stages), and without a linked issue or test demonstrating the null-URI failure mode it is hard to verify the constraint is safe. If you have a concrete scenario where a null URI causes a problem, please open an issue describing it and a fresh PR with a test — that would make this easy to accept.

PR #2877, opened 2026-02-19
Initialise address hierachy

reagan-meant:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+    /**
+     * Flush the current batch to the database.
+     */
+    private void flushBatch(List<Organization> batchToInsert, List<OrganizationTypeLinkInfo> typeLinksToCreate,
+            Map<String, Organization> pathToOrgMap) {
+        try {
+            // Insert all organizations in batch
+            for (Organization org : batchToInsert) {
+                try {
+                    String orgId = organizationService.insert(org);
```
ibacher: So, I think this line is actually problematic for effectively batching these. The issue is that `organizationService.insert()` will ultimately call [this insert method on the BaseDaoImpl](https://github.com/reagan-meant/OpenELIS-Global-2/blob/3f0138411df636246770109cbb390cea61a04dc9/src/main/java/org/openelisglobal/common/daoimpl/BaseDAOImpl.java#L511), which does:

```java
entityManager.persist(object);
entityManager.flush();
```

The immediate flush here is fine for normal processing of entities one-by-one, but it destroys any efficiency gain from trying to batch here because every call to `insert()` will basically immediately invoke the `INSERT ...` query. (Batching in Hibernate involves effectively queuing up a set of transformations and then running a single DB transaction that runs all the `INSERT` statements in one go. In order to take advantage of that, objects need to be "persisted" but not "flushed".

Instead of calling `organizationService.insert()` it actually probably makes most sense here to just directly call `entityManager.persist(org)` and then flushing them all at once.

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+                    org.setId(orgId);
```
ibacher: This line seem redundant because the value here [is already part of the object](https://github.com/reagan-meant/OpenELIS-Global-2/blob/3f0138411df636246770109cbb390cea61a04dc9/src/main/java/org/openelisglobal/common/daoimpl/BaseDAOImpl.java#L515).

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+                } catch (Exception e) {
+                    LogEvent.logError(this.getClass().getSimpleName(), "flushBatch",
+                            "Failed to insert organization: " + org.getOrganizationName() + " - " + e.getMessage());
+                    throw e;
+                }
+            }
+
+            // Flush to database
+            entityManager.flush();
```
ibacher: Note that if this was effective (i.e., running queries), we'd likely need a `try... catch` surrounding this as well.

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+            // Get organization type for this level
+            OrganizationType orgType = levelTypeMap.get(level);
+            if (orgType == null) {
+                LogEvent.logWarn(this.getClass().getSimpleName(), "processHierarchyRowBatch",
+                        "No OrganizationType found for level " + level + ". Skipping: " + name);
+                continue;
+            }
+
+            // Check if organization already exists in database
+            Organization existingOrg = findExistingOrganization(name, code, parent);
```
ibacher: I think to get the best performance out of a genuine batch operation, it would make sense to preload the set of existing organizations into memory, i.e. try to pre-populate `pathToOrgMap` with a one-time load of what's in the database. Then we can skip querying the database here. Trading up front memory to reduce database querying is generally a net-positive in batch-oriented operations.

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
ibacher: There are similar considerations with `OrganizationTypeLinkInfo`s below but because that's ultimately a [native query](https://github.com/reagan-meant/OpenELIS-Global-2/blob/3f0138411df636246770109cbb390cea61a04dc9/src/main/java/org/openelisglobal/organization/daoimpl/OrganizationOrganizationTypeDAOImpl.java#L55), we're less able to leverage Hibernate's batching for this. But at the very least the `entityManager.flush()` calls after that loop are probably not doing anything.

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+                            "Failed to link organization " + linkInfo.organization.getOrganizationName() + " to type "
+                                    + linkInfo.orgType.getName() + " - " + e.getMessage());
+                }
+            }
+
+            // Flush type links
+            entityManager.flush();
+
+            // Clear persistence context to free memory
+            entityManager.clear();
```
ibacher: For the reasons noted above, I'd move this after the `flush()` for organizations.

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+
+        // Add level prefix to ensure uniqueness across levels
+        return "L" + level + "_" + baseCode;
+    }
+
+    /**
+     * Ensure an organization is linked to the specified type. Only links if not
+     * already linked.
+     */
+    private void ensureOrganizationTypeLink(Organization org, OrganizationType orgType) {
```
ibacher: Similar to my remarks on `findExistingOrganization()` it may be more performant to pre-cache the existing organization type links (I don't have a good sense how big that set is) so that we can avoid the "SELECT..." (and implicit session flushing) here.

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+                allLines.add(line);
+            }
+        }
+        int totalRows = allLines.size();
+        LogEvent.logInfo(this.getClass().getSimpleName(), "processConfiguration",
+                "Total rows to process: " + totalRows);
+
+        // Track created organizations by their full path key to avoid duplicates
+        // Key format: "level1Name|level2Name|level3Name"
+        Map<String, Organization> pathToOrgMap = new LinkedHashMap<>();
```
ibacher: Rather than storing the full object, it may make sense to only store the primary key. Then instead of:

```java
newOrg.setOrganization(parent);
```

We can do:

```java
newOrg.setOrganization(entityManager.getReference(Organization.class, parentId));
```

This results in a light-weight, non-persistent object rather than the full persistent object which will reduce the memory pressure here and be safer across `entityManager.clear()` calls.

src/main/java/org/openelisglobal/common/util/ConfigurationProperties.java
```diff
@@ -238,6 +238,7 @@ public enum Property {
         GPS_ENABLED("gpsCoordinatesEnabled", "text"), // if true, GPS coordinate fields are displayed in order entry
         GPS_ACCURACY_METERS("gpsRequiredAccuracyMeters", "text"), // maximum acceptable GPS accuracy in meters
         GPS_TIMEOUT_SECONDS("gpsTimeoutSeconds", "text"); // timeout in seconds for GPS location requests
+        USE_NEW_ADDRESS_HIERARCHY("useNewAddressHierarchy", "text"); // if true, use new configurable address hierarchy
```
ibacher:
```suggestion
        GPS_TIMEOUT_SECONDS("gpsTimeoutSeconds", "text"), // timeout in seconds for GPS location requests
        USE_NEW_ADDRESS_HIERARCHY("useNewAddressHierarchy", "text"); // if true, use new configurable address hierarchy
```

volume/configuration/backend/address-hierarchy/haiti-values.lcsv
ibacher, line 1: While the examples make sense, I don't love the idea of having these files here. Surely this are things that are suited for the respective [distribution(https://github.com/DIGI-UW/openelis-indonesia-distro) rather that OE core.

reagan-meant: @ibacher Should I keep here a simple test file with about 10 addresses as I move the rest to the respective distros given that we done have a reference/test distro

reagan-meant:
Added full configs to distro [here](https://github.com/DIGI-UW/openelis-indonesia-distro/commit/432ce6e4d3862ddf48378bdd4f07b8388ff3be82)

reagan-meant:
@ibacher the suggested changes have greatly improved the loading time to about 4-5 minutes when loading the addresses. Thank you.

volume/configuration/backend/address-hierarchy/haiti-values.lcsv
ibacher: Yes, I think having a reference is useful!

src/main/java/org/openelisglobal/patient/service/PatientServiceImpl.java
```diff
@@ -676,6 +676,29 @@ protected void persistIdentityTypes(PatientManagementInfo patientInfo, Patient p
         persistIdentityType(patientInfo.getHealthRegion(), "HEALTH REGION", patientInfo, patient, sysUserId);
         persistIdentityType(patientInfo.getOtherNationality(), "OTHER NATIONALITY", patientInfo, patient, sysUserId);
         persistIdentityType(patientInfo.getGuid(), "GUID", patientInfo, patient, sysUserId);
+
+        // Persist dynamic address hierarchy values (addressHierarchy_0,
+        // addressHierarchy_1, etc.)
+        System.out
```
ibacher: Presumably we want to remove these debugging lines or convert them to proper logging...

docs/configuration/address-hierarchy-configuration.md
```diff
@@ -0,0 +1,277 @@
...
+- The `defaultValue` must match an organization name exactly at that level
+- Defaults cascade: if level 1 has a default, level 2's dropdown is populated
+  with its children
+- Users can always change the default selection
+- Defaults only apply to new patients, not when editing existing patients
+
+If no levels file exists, levels are created automatically from the values CSV
+header.
+
+## REST API
```
ibacher: I'm not sure this REST API stuff belongs with the detailed "How to configure this" stuff? Those seem like two different audiences.

frontend/src/components/patient/AddressSearch.js
```diff
@@ -0,0 +1,250 @@
...
+        // Handle both success and error cases - Utils.js calls callback with undefined on error
+        if (results === undefined) {
+          setSearchResults([]);
+          setIsLoading(false);
+          return;
+        }
+        const resultsArray = Array.isArray(results) ? results : [];
+        setSearchResults(resultsArray);
+        setIsLoading(false);
+        setIsOpen(resultsArray.length > 0);
```
ibacher: Do we really need this `if` here? `Array.isArray(undefined)` is `false`.

frontend/src/components/patient/CreatePatientForm.js
```diff
@@ -221,6 +226,156 @@ function CreatePatientForm(props) {
...
+          },
+        );
+      }
+    };
+
+    // Start from the first level to populate all dropdowns
+    fetchChildrenForLevels(0);
+  };
+
+  const fetchAddressHierarchyLevels = (levels) => {
```
ibacher, line 307: Is this name accurate? It only seems to fetch the top level no?

src/main/java/org/openelisglobal/address/service/AddressHierarchyConfigurationHandler.java
```diff
@@ -0,0 +1,334 @@
...
+     * Get the hierarchy level for an OrganizationType by parsing its description.
+     *
+     * @param orgType the organization type to check
+     * @return the hierarchy level, or -1 if not an address hierarchy type
+     */
+    public static int getHierarchyLevel(OrganizationType orgType) {
+        if (orgType == null || orgType.getDescription() == null) {
+            return -1;
+        }
+        if (orgType.getDescription().startsWith(ADDRESS_HIERARCHY_LEVEL_PREFIX)) {
```
ibacher: This is how we're storing the level associated with an organization? Why isn't this just a db column?

src/main/java/org/openelisglobal/address/service/AddressHierarchyConfigurationHandler.java
```diff
@@ -0,0 +1,334 @@
...
+            }
+        }
+
+        DisplayListService.getInstance().refreshLists();
+
+        LogEvent.logInfo(this.getClass().getSimpleName(), "processConfiguration",
+                "Successfully loaded " + processedTypes.size() + " address hierarchy levels from " + fileName);
+    }
+
+    private String[] parseCsvLine(String line) {
```
ibacher, line 126: General comment: We've got 5 domains based on CSVs... It seems obnoxious that they all need to define "parsing a CSV file".

src/main/java/org/openelisglobal/address/service/AddressHierarchyConfigurationHandler.java
```diff
@@ -0,0 +1,334 @@
...
+
+    @Override
+    public void processConfiguration(InputStream inputStream, String fileName) throws Exception {
+
+        BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
+
+        // Read and validate header
+        String headerLine = reader.readLine();
+        if (headerLine == null) {
+            throw new IllegalArgumentException("Address hierarchy levels file " + fileName + " is empty");
```
ibacher: Is this a stop-the-world condition or a log-an-error-and-move-on condition?

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,685 @@
...
+                "COMPLETED processing " + fileName + " in " + (totalTime / 1000) + " seconds: " + rowsProcessed
+                        + " rows processed, " + pathToOrgIdMap.size() + " unique locations created/updated, " + errors
+                        + " errors");
+    }
+
+    /**
+     * Helper class to track type links that need to be created after organization
+     * insert.
+     */
+    private static class OrganizationTypeLinkInfo {
```
ibacher: I'd probably use a `Record` for this (and the other one).

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,685 @@
...
+    }
+
+    /**
+     * Generate a unique code for a location based on its name and position in
+     * hierarchy.
+     */
+    private String generateCode(String name, int level, String parentPath) {
+        // Create a code from the name: uppercase, no spaces, max 20 chars
+        String baseCode = name.toUpperCase().replaceAll("[^A-Z0-9]", "").substring(0,
+                Math.min(name.replaceAll("[^A-Za-z0-9]", "").length(), 15));
```
ibacher: This is a sign that we should run the replace all, store that in a variable and reference that here.

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,685 @@
...
+
+        return String.join("|", pathParts);
+    }
+
+    /**
+     * Generate a unique code for a location based on its name and position in
+     * hierarchy.
+     */
+    private String generateCode(String name, int level, String parentPath) {
+        // Create a code from the name: uppercase, no spaces, max 20 chars
```
ibacher: This doesn't seem accurate

ibacher: I.e., we don't filter spaces and the max is 15?

src/main/java/org/openelisglobal/address/service/AddressHierarchyValuesConfigurationHandler.java
```diff
@@ -0,0 +1,685 @@
...
+        // Not linked, add the link
+        organizationService.linkOrganizationAndType(org, orgType.getId());
+        // Add to cache so we don't try to create it again
+        existingTypeLinks.add(linkKey);
+    }
+
+    /**
+     * Truncate a string for logging purposes.
+     */
+    private String truncate(String str, int maxLength) {
```
ibacher: These sort of utilities usually make sense to factor out into a utilities class of some sort.

src/main/java/org/openelisglobal/common/rest/provider/PatientSearchPopulateRestController.java
```diff
@@ -178,4 +191,30 @@ private String getPatientType(Patient patient) {
...
+    /**
+     * Get identity value without auto-creating the identity type if it doesn't
+     * exist. Returns empty string if not found.
+     */
+    private String getIdentityValueSafe(List<PatientIdentity> identityList, String typeName) {
+        if (typeName == null || identityList == null) {
+            return "";
+        }
+        try {
+            PatientIdentityTypeService identityTypeService = SpringContext.getBean(PatientIdentityTypeService.class);
```
ibacher: Pulling from `SpringContext` like this is an antipattern. Just autowire it.

src/main/java/org/openelisglobal/configuration/service/ConfigurationInitializationService.java
```diff
@@ -220,6 +220,13 @@ private LoadResult processFiles(DomainConfigurationHandler handler, Map<String,
             InputStreamSource streamSource = entry.getValue();
             filesFound = true;
 
+            // Check if this handler can process this file
+            if (!handler.canProcess(fileName)) {
```
ibacher: Instead of this, why don't we extend `DomainConfigurationHandler` to support `getFileMatcher()` which can return a pattern like those supported by [Spring's AntPathMatcher](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html)? We can then rework things so that if `getExtension()` returns `null`, we check the `getFileMatcher()` and use that with the `PathMatchingResourcePatternResolver` (which supports non-classpath paths). Slightly bigger refactor, but it will wind up being more declarative I think. And that way this will play nicely with the loading logic.

src/main/java/org/openelisglobal/organization/daoimpl/OrganizationDAOImpl.java
```diff
@@ -545,4 +545,74 @@ public Organization getOrganizationByFhirId(String uuid) {
...
+    @Transactional(readOnly = true)
+    public Organization getOrganizationByCode(String code) {
+        if (GenericValidator.isBlankOrNull(code)) {
+            return null;
+        }
+
+        String sql = "from Organization o where o.code = :code";
+
+        try {
+            Query<Organization> query = entityManager.unwrap(Session.class).createQuery(sql, Organization.class);
```
ibacher: `unwrap()` feels unnecessary here. This isn't a SQL query and you should just be able to call `entityManager.createQuery()`.

src/main/java/org/openelisglobal/organization/daoimpl/OrganizationDAOImpl.java
```diff
@@ -545,4 +545,74 @@ public Organization getOrganizationByFhirId(String uuid) {
...
+
+        return null;
+    }
+
+    @Override
+    @Transactional(readOnly = true)
+    public List<Organization> searchOrganizationsWithTypes(String filter) throws LIMSRuntimeException {
+        List<Organization> list;
+        try {
+            // Simple query - we'll eagerly initialize within the transaction
```
ibacher: ???

src/main/java/org/openelisglobal/organization/daoimpl/OrganizationDAOImpl.java
```diff
@@ -545,4 +545,74 @@ public Organization getOrganizationByFhirId(String uuid) {
...
+            list = query.list();
+
+            // Force initialization of all organization types and parent chain within the
+            // transaction
+            // This supports any number of hierarchy levels
+            for (Organization org : list) {
+                initializeOrganizationHierarchy(org);
+            }
+        } catch (RuntimeException e) {
+            LogEvent.logError(e);
```
ibacher: Shouldn't the rest of the context added to the exception be added here?

src/main/java/org/openelisglobal/organization/service/OrganizationServiceImpl.java
```diff
@@ -226,4 +226,16 @@ public void activateOrganizationsAndDeactivateOthers(List<String> organizationNa
     public Organization getOrganizationByFhirId(String uuid) {
         return baseObjectDAO.getOrganizationByFhirId(uuid);
     }
+
+    @Override
+    @Transactional(readOnly = true)
```
ibacher, line 231: The `@Transactional()` annotations here add no value and potentially create problems since the DAOs themselves are transactional. Unless (like e.g., `activateOrganizationsAndDeactivateOthers` we need a single transaction for multiple objects), but this is exactly like `getOrganizationByFhirId()`

volume/configuration/backend/address-hierarchy/indonesia-values.csv
```diff
@@ -0,0 +1,15 @@
...
+DKI JAKARTA,KOTA JAKARTA SELATAN,JAGAKARSA,CIGANJUR
+DKI JAKARTA,KOTA JAKARTA SELATAN,JAGAKARSA,JAGAKARSA
+DKI JAKARTA,KOTA JAKARTA SELATAN,JAGAKARSA,LENTENG AGUNG
+DKI JAKARTA,KOTA JAKARTA SELATAN,JAGAKARSA,TANJUNG BARAT
+MALUKU UTARA,KABUPATEN HALMAHERA TIMUR,WASILE TIMUR,DAKA INO
+MALUKU UTARA,KABUPATEN HALMAHERA TIMUR,WASILE TIMUR,AKE DAGA
+MALUKU UTARA,KABUPATEN HALMAHERA TIMUR,WASILE TIMUR,TOBOINO
+MALUKU UTARA,KABUPATEN HALMAHERA TIMUR,WASILE TIMUR,DODAGA
+MALUKU UTARA,KABUPATEN HALMAHERA TIMUR,WASILE TIMUR,TUTULING JAYA
+MALUKU UTARA,KABUPATEN HALMAHERA TIMUR,WASILE TIMUR,WOKA JAYA
```
ibacher, line 15:
```suggestion
MALUKU UTARA,KABUPATEN HALMAHERA TIMUR,WASILE TIMUR,WOKA JAYA

```

ibacher:
I've got a branch over [here](https://github.com/DIGI-UW/OpenELIS-Global-2/tree/OGC-317-suggestion) built on this one that implements a replacement mechanism for `canProcess()`, essentially by allowing domains to define patterns instead of just extensions and tracking which files have been "claimed" so files processed by one domain aren't re-processed.

frontend/src/components/patient/CreatePatientForm.js
reagan-meant: This is the callback function after fetching address hierarchy levels before attempting to set the defaults

src/main/java/org/openelisglobal/address/service/AddressHierarchyConfigurationHandler.java
reagan-meant: Will wire a follow up commit once this is done to fix that for all domains

PR #2886, opened 2026-02-20
Fix PostgreSQL type mismatch in SampleProjectDAOImpl (#2646)

### LINKED ISSUE #2646: Investigate  PostgreSQL Operator Error in SampleProjectServiceTest
Test failure in `SampleProjectServiceTest` due to PostgreSQL type mismatch error: `operator does not exist: numeric = bytea`

## Environment
- **Test Class**: `org.openelisglobal.sampleproject.SampleProjectServiceTest`
- **Date/Time**: 26 Jan 2026 00:37:47
- **Severity**: ERROR

### Error Message
```
ERROR: operator does not exist: numeric = bytea
Position: 341
```

## Description
The test is failing because PostgreSQL cannot find an operator to compare a `numeric` type with a `bytea` type. This indicates a type mismatch in the SQL query at position 341.

pantha704:
### Description

This PR fixes [Issue #2646](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/2646) where the `SampleProjectServiceTest` was failing on PostgreSQL environments due to a type mismatch error (`operator does not exist: numeric = bytea`).

### Root Cause

The `SampleProjectDAOImpl` was passing string IDs to the `Query.setParameter` method by forcing a cast via `Integer.parseInt(id)` (and similarly `Integer.valueOf(organizationId)`). Because the entities (`SampleProject` and `Organization`) map their IDs using OpenELIS's custom string type mapper (`LIMSStringNumberUserType`), passing an `Integer` caused Hibernate to serialize the parameter as a binary object (`bytea`), resulting in the SQL failure.

### Changes Made

- Removed `Integer.parseInt(id)` for the `sampleId` parameter and replaced it with direct string binding.
- Removed an identical antipattern for `organizationId` (`Integer.valueOf(organizationId)`) later in the same file to proactively ensure related queries execute cleanly on PostgreSQL.
- Did not change any public API or method signatures.

Fixes #2646

### DIFF
--- a/src/main/java/org/openelisglobal/common/daoimpl/BaseDAOImpl.java
+++ b/src/main/java/org/openelisglobal/common/daoimpl/BaseDAOImpl.java
@@ -674,22 +674,26 @@ public abstract class BaseDAOImpl<T extends BaseObject<PK>, PK extends Serializa
             }
             Predicate predicate;
             switch (comparisonOperation.getComparison()) {
-            case EQ:
-                predicate = criteriaBuilder.equal(pathToProperty, propertyValue);
-                break;
-            case LIKE:
-                predicate = criteriaBuilder.like(criteriaBuilder.lower(pathToProperty),
-                        "%" + ((String) propertyValue).toLowerCase() + "%");
-                break;
-            case IN:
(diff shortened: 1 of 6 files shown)


mozzy11:

PR #2914, opened 2026-02-23
Fix All Permissions checkbox to select all available roles

### LINKED ISSUE #2913: Bug Report: "All Permissions" checkbox only selects 4 roles instead of all available roles
The `All Permissions` checkbox in the User Management page `/MasterListsPage/userManagement` is hardcoded to only select 4 specific roles (Reception, Reports, Results, Validation) instead of selecting ALL available roles for a lab unit. This makes the checkbox misleading and breaks the expected `select all` functionality.

**Steps to Reproduce**
  1. Log in as an administrator
  2. Navigate to Admin → User Management
  3. Click "Add" to create a new user OR click "Edit" on an existing user
  4. Scroll down to the "Lab Unit Roles" section
  5. Select any lab unit from the dropdown (e.g.,"Serology", etc.)
  6. Observe all the available role checkboxes that appear (should be 20+ roles including Chemical Analyst, Pharmacist, Data Manager, etc.)
  7. Click the "All Permissions" checkbox
  8. BUG: Only 4 roles get selected (Reception, Reports, Results, Validation)

  **Technical Details**
  **Current hardcoded logic**:
```
  checked={["4", "5", "7", "10"].every(  // Only these 4 roles!
  onChange={() => {
    const numbersToAdd = ["4", "5", "7", "10"];  // Hardcoded array
```

  Should be dynamic: Get all available role IDs from `userDataShow.labUnitRoles.map(section => section.roleId)`

utafrali:
Fixes #2913

The "All Permissions" checkbox in the user role management form was hardcoded to only check for 4 specific role IDs (4, 5, 7, 10). This prevented users from selecting all roles when more than 4 roles existed in the system.

This change replaces the hardcoded role list with a dynamic lookup that uses all available roles from `userDataShow.labUnitRoles`. The checkbox now correctly validates against and selects all roles present in the system, not just a fixed subset.

### DIFF
--- a/frontend/src/components/admin/userManagement/UserAddModify.js
+++ b/frontend/src/components/admin/userManagement/UserAddModify.js
@@ -1317,13 +1317,20 @@ function UserAddModify() {
                         <Checkbox
                           id={`all-permissions-${key}`}
                           labelText={"All Permissions"}
-                          checked={["4", "5", "7", "10"].every(
+                          checked={(
+                            userDataShow?.labUnitRoles?.map(
+                              (section) => section.roleId,
+                            ) ?? []
+                          ).every(
                             (num) =>
                               selectedTestSectionLabUnits[key] &&
                               selectedTestSectionLabUnits[key].includes(num),
                           )}
                           onChange={() => {
-                            const numbersToAdd = ["4", "5", "7", "10"];
+                            const numbersToAdd =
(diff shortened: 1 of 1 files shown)


utafrali:
Closing this PR as it has been open for 1d 1h without maintainer feedback. If the fix is still needed, I'd be happy to reopen or submit an updated version. Thank you for your time!

PR #2917, opened 2026-02-24
feat: add Organization FHIR facade provider with Create, Update, Delete with fhir utils

vsvishalsharma:
## Description

Adds `OrganizationProvider` as a new HAPI FHIR `IResourceProvider` for the **Organization** resource, extending the FHIR facade beyond the existing `PractitionerProvider`.

This implementation follows the same architectural and design pattern used in `PractitionerProvider` to maintain consistency across the FHIR facade layer.

---

## Operations Implemented

| Operation | Annotation | Behavior |
|------------|------------|-----------|
| **Create** | `@Create` | Accepts FHIR `Organization` → transforms to local entity via `transformToOrganization()` → persists locally → syncs to FHIR store |
| **Update** | `@Update` | Looks up existing org by FHIR UUID via `getOrganizationByFhirId()` → merges fields using transform → saves → syncs to FHIR store |
| **Delete** | `@Delete` | Soft-delete: sets `isActive = "N"` → syncs deactivation to FHIR store |

---

## Design Decisions

- Follows the exact same pattern as `PractitionerProvider` for structural and behavioral consistency  
- Imports `org.hl7.fhir.r4.model.Organization` directly; uses fully qualified name (FQN) for the local valueholder to avoid name collision  
- Introduces `syncToFhirStore()` helper to reduce duplication across operations  
- FHIR store sync failures are logged but **do not block** the local operation (non-fatal by design)  
- Auto-discovered by `FhirRestfulServer` — no routing changes required in `InternalFhirApi`  

---

## Reused Infrastructure

- `FhirTransformService.transformToOrganization()` (FHIR → local)  
- `FhirTransformService.transformToFhirOrganization()` (local → FHIR)  
- `OrganizationService.getOrganizationByFhirId()` for UUID lookups  

---

## Related Issue

Part of the FHIR facade expansion effort.

### DIFF
--- a/src/main/java/org/openelisglobal/fhir/providers/FhirProviderUtils.java
+++ b/src/main/java/org/openelisglobal/fhir/providers/FhirProviderUtils.java
@@ -6,6 +6,7 @@ import jakarta.servlet.http.HttpServletRequest;
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.hl7.fhir.r4.model.IdType;
 import org.hl7.fhir.r4.model.OperationOutcome;
+import org.hl7.fhir.r4.model.Resource;
 import org.openelisglobal.common.log.LogEvent;
 import org.openelisglobal.common.util.ControllerUtills;
 import org.openelisglobal.dataexchange.fhir.service.FhirPersistanceService;
@@ -72,7 +73,7 @@ public final class FhirProviderUtils {
      * not propagate — the local operation is considered successful even if the sync
      * fails.
      */
-    public static void syncToFhirStore(FhirPersistanceService fhirPersistenceService, IBaseResource resource,
+    public static void syncToFhirStore(FhirPersistanceService fhirPersistenceService, Resource resource,
             String callerClassName, String callingMethod) {
         try {
             fhirPersistenceService.updateFhirResourceInFhirStore(resource);
--- a/src/main/java/org/openelisglobal/fhir/providers/FhirProviderUtils.java
+++ b/src/main/java/org/openelisglobal/fhir/providers/FhirProviderUtils.java
@@ -1,0 +1,107 @@
+package org.openelisglobal.fhir.providers;
+
+import ca.uhn.fhir.rest.api.MethodOutcome;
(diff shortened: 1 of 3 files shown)


mherman22 (commented):
Hey @vsvishalsharma , nice work here. Just this major comment.

is it possible to have some sort of a utils class to reduce boiler plate around all the provider classes that will be created. could be called `FhirProviderUtils`?

Something to handle some of these as util methods that can simply be called in these fhir provider classes you are creating:

For example below is the update util method, this can be done for `buildDeleteOutcome`, `buildCreateOutcome`, 
```
    public static MethodOutcome buildUpdateOutcome(IBaseResource resource) {
        MethodOutcome outcome = new MethodOutcome();
        outcome.setId(resource.getIdElement());
        outcome.setResource(resource);
        outcome.setCreated(false);
        outcome.setResponseStatusCode(200);
        return outcome;
    }
```

    public static MethodOutcome buildCreateOutcome(IBaseResource resource) {
        MethodOutcome outcome = new MethodOutcome();
        outcome.setId(resource.getIdElement());
        outcome.setResource(resource);
        outcome.setCreated(true);
        outcome.setResponseStatusCode(201);
        return outcome;
    }

Even `syncToFhirStore`, `setSysUserId` should be a util method since it will be used across many classes.

src/main/java/org/openelisglobal/fhir/providers/FhirProviderUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Shared utility methods for FHIR resource providers. Eliminates boilerplate
+ * around {@link MethodOutcome} construction, FHIR store synchronization, and
+ * common validation that is identical across all
+ * {@link ca.uhn.fhir.rest.server.IResourceProvider} implementations.
+ */
+public final class FhirProviderUtils {
+
+    private FhirProviderUtils() {
```
mherman22, line 22: this can be simplified by `@NoArgsConstructor(access = AccessLevel.PRIVATE)`

mherman22 (approved):
LGTM 
cc: @mozzy11

PR #2971, opened 2026-02-28
feat: add UPDATE/DELETE operations for FHIR Observation resource

nigar-08:
## Summary
This PR extends the native FHIR Observation facade to support full CRUD operations:
Read: fetch lab results by UUID (GET /fhir/Observation/{uuid})
Update: modify result value/type (PUT /fhir/Observation/{uuid})
Delete: soft-delete (marks FHIR status as CANCELLED) (DELETE /fhir/Observation/{uuid})
Search: forwards to HAPI store with full FHIR parameter set
## Screenshots

## Related Issue

## Other


nigar-08:
Hi @mozzy11 —
This PR adds @Update and @Delete for FHIR Observation

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +121,85 @@ public Observation read(@IdParam IdType id) {
         }
     }
 
+    @Update
+    public MethodOutcome update(@IdParam IdType theId, @ResourceParam Observation fhirObservation,
```
mherman22, line 117:
```suggestion
    public MethodOutcome update(@NonNull @IdParam IdType theId, @ResourceParam Observation fhirObservation,
```

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -15,19 +19,26 @@
...
+import org.openelisglobal.analysis.service.AnalysisService;
 import org.openelisglobal.common.log.LogEvent;
+import org.openelisglobal.common.services.IStatusService;
 import org.openelisglobal.dataexchange.fhir.FhirUtil;
+import org.openelisglobal.dataexchange.fhir.service.FhirPersistanceService;
 import org.openelisglobal.dataexchange.fhir.service.FhirTransformService;
 import org.openelisglobal.result.service.ResultService;
+import org.openelisglobal.result.valueholder.Result;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
```
mherman22, line 42:
```suggestion
import lombok.NonNull;
```

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +121,85 @@ public Observation read(@IdParam IdType id) {
...
+            throw e;
+        } catch (Exception e) {
+            LogEvent.logError(this.getClass().getSimpleName(), method,
+                    "Unexpected error updating observation: " + e.getMessage());
+            throw new InternalErrorException("Unexpected server error updating Observation");
+        }
+    }
+
+    @Delete
+    public MethodOutcome delete(@IdParam IdType theId, HttpServletRequest request) {
```
mherman22, line 161:
```suggestion
    public MethodOutcome delete(@NonNull @IdParam IdType theId, HttpServletRequest request) {
```

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +121,85 @@ public Observation read(@IdParam IdType id) {
         }
     }
 
+    @Update
+    public MethodOutcome update(@IdParam IdType theId, @ResourceParam Observation fhirObservation,
+            HttpServletRequest request) {
+        String method = "update";
+        LogEvent.logDebug(this.getClass().getSimpleName(), method,
+                "Received FHIR UPDATE request for Observation ID: " + (theId != null ? theId.getIdPart() : "null"));
```
mherman22, line 121:
```suggestion
                "Received FHIR UPDATE request for Observation ID: " + theId.getIdPart());
```

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +121,85 @@ public Observation read(@IdParam IdType id) {
...
+                    "Unexpected error updating observation: " + e.getMessage());
+            throw new InternalErrorException("Unexpected server error updating Observation");
+        }
+    }
+
+    @Delete
+    public MethodOutcome delete(@IdParam IdType theId, HttpServletRequest request) {
+        String method = "delete";
+        LogEvent.logDebug(this.getClass().getSimpleName(), method,
+                "Received FHIR DELETE request for Observation ID: " + (theId != null ? theId.getIdPart() : "null"));
```
mherman22, line 164:
```suggestion
                "Received FHIR DELETE request for Observation ID: " + theId.getIdPart());
```

mherman22 (changes requested):
Some small changes here.

Ref: https://projectlombok.org/features/NonNull

nigar-08:
> Some small changes here.
> 
> Ref: https://projectlombok.org/features/NonNull

Thanks @mherman22, applied all suggestions!

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +122,85 @@ public Observation read(@IdParam IdType id) {
...
+        try {
+            FhirProviderUtils.validateIdParam(theId, "Observation", this.getClass().getSimpleName(), method);
+
+            Result result = resultService.getResultByFhirUuid(theId.getIdPart());
+            if (result == null) {
+                throw new ResourceNotFoundException("Observation/" + theId.getIdPart());
+            }
+
+            result.setSysUserId(FhirProviderUtils.getSysUserId(request));
+            resultService.save(result);
```
vsvishalsharma, line 174: Saving the Result without updating any status/active flag means OpenELIS state is unchanged. DELETE becomes a no-op on the OE DB. Please apply a soft-delete field or status update before saving (see OrganizationProvider.delete() pattern)

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +122,85 @@ public Observation read(@IdParam IdType id) {
         }
     }
 
+    @Update
+    public MethodOutcome update(@NonNull @IdParam IdType theId, @ResourceParam Observation fhirObservation,
```
vsvishalsharma: Using Lombok @NonNull here produces a bare NPE → HTTP 500. HAPI expects InvalidRequestException via validateIdParam. Please remove @NonNull and the Lombok import.

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -53,9 +72,18 @@ public class ObservationProvider implements IResourceProvider {
...
     private FhirTransformService fhirTransformService;
 
+    @Autowired
+    private FhirPersistanceService fhirPersistenceService;
+
     @Autowired
     private ResultService resultService;
 
+    @Autowired
+    private AnalysisService analysisService;
```
vsvishalsharma: Not being used anywhere

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -53,9 +72,18 @@ public class ObservationProvider implements IResourceProvider {
...
+
+    @Autowired
+    private IStatusService statusService;
```
vsvishalsharma: same here as well

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +122,85 @@ public Observation read(@IdParam IdType id) {
...
+            if (fhirObservation.hasValueQuantity()) {
+                existingResult.setValue(fhirObservation.getValueQuantity().getValue().toPlainString());
+                existingResult.setResultType("N");
+            } else if (fhirObservation.hasValueStringType()) {
+                existingResult.setValue(fhirObservation.getValueStringType().getValue());
+                existingResult.setResultType("T");
+            }
+
+            existingResult.setSysUserId(FhirProviderUtils.getSysUserId(request));
+            Result updatedResult = resultService.save(existingResult);
```
vsvishalsharma, line 139: Direct resultService.save() bypasses OE analysis/result workflow (validation/release steps). Please consider using the same save logic used by manual results entry to maintain audit/state integrity.

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -94,6 +122,85 @@ public Observation read(@IdParam IdType id) {
...
+
+            Observation resultObservation = fhirTransformService.transformResultToObservation(updatedResult);
```
vsvishalsharma, line 141: Other providers call resource.setId(theId) before returning the updated resource. Without this, the returned Observation may have an inconsistent ID. Please align with Practitioner/Organization providers

vsvishalsharma (changes requested):

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
nigar-08: Hey @vsvishalsharma  — thanks for the suggestion but Result doesn't have an isActive equivalent. I checked Result.java and ResultDAOImpl — the only candidate is isReportable (line 39), but that field is used in SQL filters for report generation (ResultDAOImpl line 233: where r.isReportable = YES). Setting it to "N" for deletion would silently drop results from clinical reports — that's a data integrity risk, not a soft-delete.
OrganizationProvider can do setIsActive(NO) because Organization has that field. Result doesn't. If you know of a proper lifecycle field on Result I missed, happy to use it. Otherwise I've added a TODO to track adding one in a follow-up PR.

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
nigar-08: You're right that LogbookResultsPersistService would be the proper path for workflow integrity. I checked it — its only method is persistDataSet(ResultsUpdateDataSet, List<IResultUpdate>, String). ResultsUpdateDataSet holds lists of TestResultItem, Analysis, ReferralSet, Note — it's a bulk UI workflow object built for lab technician entry, not single FHIR resource updates.
Wrapping a single Result in it safely requires understanding IResultUpdate implementations and ResultSet construction — rushing that risks breaking the lab workflow on clinical data. I'd rather do that investigation properly in a dedicated PR.
Also worth noting — OrganizationProvider uses direct organizationService.save() with no workflow service either. Added a TODO to investigate whether single-result FHIR updates can safely integrate with the workflow — the bulk design of ResultsUpdateDataSet exists for atomic integrity reasons (Analysis status, referrals, notes saved together). A single-result path may not even be the right solution — a FHIR Bundle transaction might be more appropriate.

mozzy11:
Thanks @nigar-08

PR #3008, opened 2026-03-05
feat: add @Create for Observation 

nigar-08:
## Summary

Added @Create to ObservationProvider so lab results can be submitted via POST /fhir/Observation.
The endpoint resolves the Analysis from the basedOn ServiceRequest reference and persists the result through ResultSaveService + LogbookResultsPersistService — the same pipeline the logbook UI uses internally, so no workflow steps are bypassed.
To support this, getAnalysisByFhirUuid() was added across the DAO/Service layers since no existing method looked up Analysis by FHIR UUID. transformResultToObservation() was also exposed on the FhirTransformService interface since it existed in the impl but wasn't accessible via the interface.
The endpoint validates that the Analysis exists, is not in a terminal status (Finalized, Canceled, SampleRejected), and has no existing result before persisting. Full URL basedOn references are handled correctly alongside relative ones.


nigar-08:
Hey @mozzy11 , I've implemented the Observation @Create endpoint. Can you review when you get a chance?

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -113,6 +137,106 @@ public Observation read(@IdParam IdType id) {
         }
     }
 
+    @Create
+    public MethodOutcome create(@ResourceParam Observation fhirObservation, HttpServletRequest request) {
```
vsvishalsharma, line 141: check if this shoudl be transcational 

IMO :: method persists data via logbookResultsPersistService.persistDataSet(...) but has no @Transactional boundary. If the FHIR store sync or observation transform fails after DB persistence, you're left with a dangerously inconsistent state. At minimum the DB persistence should be wrapped in a transaction.

src/main/java/org/openelisglobal/fhir/providers/ObservationProvider.java
```diff
@@ -113,6 +137,106 @@ public Observation read(@IdParam IdType id) {
...
+            } else {
+                throw new InvalidRequestException("Observation must have either valueQuantity or valueString");
+            }
+
+            String sysUserId = FhirProviderUtils.getSysUserId(request);
+            if (sysUserId == null) {
+                throw new InvalidRequestException("Authentication required to submit a result");
+            }
+
+            ResultSaveService resultSaveService = new ResultSaveService(analysis, sysUserId);
```
vsvishalsharma, line 200: This bypasses Spring's container entirely no AOP, no proxying, no consistent lifecycle. This should be injected as a @Autowired bean (or a factory/builder pattern if it requires per-request state).

mozzy11:

We are doing a cleanup of the PR queue: this PR is now more than 3 months old and has unresolved merge conflicts with develop. The codebase has changed substantially in that time (the frontend moved to Vite and is being migrated to TypeScript, the test infrastructure was rewritten, and several modules — test catalog, analyzers, EQA, storage — were reworked), so a mechanical rebase is unlikely to succeed.

PR #3010, opened 2026-03-05
OGC-349: Make metadata translatable to any language

reagan-meant:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


src/main/resources/hibernate/hbm/LocalizationValue.hbm.xml
ibacher, line 1: For new tables, we should prefer Hibernate / JPA annotations to hbm.xml files. For the existing table, it would be good if we could take this opportunity to migrate it.

src/main/resources/liquibase/3.5.x.x/002-add-dictionary-localization.xml
ibacher, line 1: I think it's better practice to combine all of these changelogs into a single changelog for this feature. Changelogs are already read and processed sequentially.

src/main/resources/liquibase/3.5.x.x/base.xml
```diff
@@ -0,0 +1,18 @@
...
+
+  <!-- Generic Localization System - Enables unlimited language support -->
+
+  <!-- Phase 1: Create new tables and migrate data -->
+  <include relativeToChangelogFile="true" file="001-add-localization-value-table.xml"/>
+
+  <!-- Phase 2: Add localization to dictionary entries -->
+  <include relativeToChangelogFile="true" file="002-add-dictionary-localization.xml"/>
+
+  <!-- Phase 4: Drop legacy columns -->
```
ibacher: Phase description here seems to be missing one?

src/main/java/org/openelisglobal/localization/service/SupportedLocaleConfigurationHandler.java
```diff
@@ -0,0 +1,244 @@
...
+        } else {
+            locale.setActive(true); // Default to active
+        }
+
+        String isFallback = getValueOrEmpty(values, isFallbackIndex);
+        if (!isFallback.isEmpty()) {
+            boolean shouldBeFallback = "Y".equalsIgnoreCase(isFallback) || "true".equalsIgnoreCase(isFallback);
+            if (shouldBeFallback) {
+                // Clear any existing fallback first
+                clearExistingFallback();
```
ibacher: It seems like we should have some kind of sanity check that only at most one locale is indicated as the fallback. It's one thing to clear this once per startup, but additional fallbacks should probably fail as a configuration error.

src/main/java/org/openelisglobal/localization/service/LocalizationValueServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    @Transactional(readOnly = true)
+    public Optional<LocalizationValue> getByLocalizationIdAndLocale(String localizationId, String locale) {
+        return localizationValueDAO.getByLocalizationIdAndLocale(localizationId, locale);
+    }
+
+    @Override
+    @Transactional(readOnly = true)
+    public String getLocalizedValue(String localizationId, String locale) {
+        Optional<LocalizationValue> value = getByLocalizationIdAndLocale(localizationId, locale);
```
ibacher: Couple of points here:

1. The intended API for `Optional` is something more like:
```java
getByLocalizationIdAndLocale().orElseGet().orElseGet()
```
2. It might be worth thinking about supporting full BCP-47 locales, e.g., `fr-CI` and supporting those falling back, e.g, `fr-CI` -> `fr`. The Java [Locale](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html) machinery may be some help in implementing this.
3. Instead of doing 3 DB hits, since we have a `localizationId` (and these methods should bail when neither is provided! It probably makes more sense to just load all of the values for that key and then solve for the correct locale in memory. This is slightly wasteful, but it's likely that there's a small enough set of localizations that it makes sense.

src/main/java/org/openelisglobal/localization/controller/rest/LocalizationRestController.java
```diff
@@ -0,0 +1,581 @@
...
+
+    /**
+     * Get localizations with missing translations for a specific locale.
+     *
+     * @param locale the locale code to check
+     * @return list of localizations missing translations for that locale
+     */
+    @GetMapping("/missing/{locale}")
+    public ResponseEntity<List<LocalizationDTO>> getMissingTranslations(@PathVariable String locale) {
+        List<Localization> allLocalizations = localizationService.getAll();
```
ibacher: I think instead of loading everything, we can do this:

```hql
SELECT l.id, l.description
FROM Localization l
WHERE NOT EXISTS (
  SELECT 1
  FROM LocalizationValue lv
  WHERE lv.localization_id = l.id
    AND lv.locale = :locale
    AND lv.value IS NOT NULL
    AND lv.value != ''
);
```

src/main/java/org/openelisglobal/localization/controller/rest/LocalizationRestController.java
```diff
@@ -0,0 +1,581 @@
...
+        return ResponseEntity.ok(missing);
+    }
+
+    /**
+     * Get translation statistics.
+     *
+     * @return statistics about translations per locale
+     */
+    @GetMapping("/stats")
+    public ResponseEntity<TranslationStatsDTO> getTranslationStats() {
```
ibacher, line 115: This is replaceable by a single query (in SQL):

```sql
SELECT sl.locale_code,
  sl.display_name,
  COUNT(lv.id) AS translated,
  (SELECT COUNT(*) FROM clinlims.localization) - COUNT(lv.id) AS missing
FROM clinlims.supported_locale sl
LEFT JOIN clinlims.localization_value lv
  ON lv.locale = sl.locale_code
 AND lv.value IS NOT NULL
 AND lv.value != ''
WHERE sl.is_active = true
GROUP BY sl.locale_code, sl.display_name, sl.sort_order
ORDER BY sl.sort_order;
```

src/main/java/org/openelisglobal/localization/controller/rest/SupportedLocaleRestController.java
```diff
@@ -0,0 +1,363 @@
...
+                SupportedLocale old = currentFallback.get();
+                old.setFallback(false);
+                old.setSysUserId(getSysUserId(request));
+                supportedLocaleService.update(old);
+            }
+
+            // Set new fallback
+            newFallback.setFallback(true);
+            newFallback.setSysUserId(getSysUserId(request));
+            supportedLocaleService.update(newFallback);
```
ibacher: The logic here needs to be pushed into a service call so that it's performed transactionally. As is, with two calls to `supportedLocaleService.update()` and no db constraint enforcing the "one fallback" rule this can potentially leave the database in an inconsistent state.

Generally speaking, REST controllers should have as little logic as possible as mostly defer to services to provide the implemetnation.

src/main/java/org/openelisglobal/localization/controller/rest/LocalizationRestController.java
```diff
@@ -0,0 +1,581 @@
...
+                return ResponseEntity.notFound().build();
+            }
+
+            String sysUserId = getSysUserId(request);
+
+            for (Map.Entry<String, String> entry : translations.entrySet()) {
+                String locale = entry.getKey();
+                String value = entry.getValue();
+
+                LocalizationValue lv = localizationValueService.setTranslation(id, locale, value);
```
ibacher: It feels like `sysUserId` should be passed in here so that it's part of the record on insert or update rather than adding it to the Hibernate session to be committed at an unknown latter point.

src/main/java/org/openelisglobal/localization/controller/rest/LocalizationRestController.java
```diff
@@ -0,0 +1,581 @@
...
+            List<String> errors = new ArrayList<>();
+
+            for (LocalizationImportDTO item : importData) {
+                try {
+                    Localization localization = localizationService.get(item.getId());
+                    if (localization != null) {
+                        for (Map.Entry<String, String> entry : item.getTranslations().entrySet()) {
+                            LocalizationValue lv = localizationValueService.setTranslation(item.getId(), entry.getKey(),
+                                    entry.getValue());
+                            lv.setSysUserId(sysUserId);
```
ibacher: Same remark here as with the parallel call above.

src/main/java/org/openelisglobal/localization/valueholder/Localization.java
```diff
@@ -33,8 +33,24 @@ public class Localization extends BaseObject<String> {
...
+
+    /**
+     * New normalized translation values stored in localization_value table. Key is
+     * locale code (e.g., "en", "fr"), value is the LocalizationValue entity.
+     */
+    private Map<String, LocalizationValue> values = new HashMap<>();
+
+    /**
+     * Legacy locale values map used by the old Hibernate mapping. This will be
+     * populated from the english/french columns during the transition period.
```
ibacher: What is this "transition period"?

ibacher: More exactly, this seems to be retained to also retain `getLocaleValues()` which is called in [exactly one place ](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/b640457305fe0240f7b5030d8f52f44033969c20/src/main/java/org/openelisglobal/common/util/DefaultConfigurationProperties.java#L638) in existing code. It seems easier to just rip this out and update that one callsite to use the new API.

ibacher (commented):
Thanks @reagan-meant. This is exactly what we discussed and I think the implementation basically looks right except for the things I've noted in comments.

One thing we'll need to look into in the future is enabling some degree of caching. I sort of figured that OE would be using at least [Hibernate's cache](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/b640457305fe0240f7b5030d8f52f44033969c20/src/main/resources/hibernate/hibernate.cfg.xml#L26) or Spring's [caching](https://docs.spring.io/spring-framework/reference/6.2/integration/cache.html). We should queue up enabling those as future work (that can probably be a community task cc: @pmanko @caseyi). Once some level of caching is available we should heavily leverage it here as translation strings are definitely read many more times than they are written. FWIW, OpenMRS recently moved to [Infinispan](https://infinispan.org/) (from EhCache), which is the closest thing to "Java Redis", but OE could even benefit from something like [Caffeine](https://github.com/ben-manes/caffeine), which defaults to entirely in-memory (I actually think Infinispan uses Caffeine as its first level of caching).

src/main/java/org/openelisglobal/localization/controller/rest/SupportedLocaleRestController.java
```diff
@@ -0,0 +1,363 @@
...
+            Optional<SupportedLocale> existing = supportedLocaleService.getByLocaleCode(dto.getLocaleCode());
+            if (existing.isPresent()) {
+                return ResponseEntity.badRequest()
+                        .body(new ErrorResponse("Locale code already exists: " + dto.getLocaleCode()));
+            }
+
+            SupportedLocale locale = fromDTO(dto);
+            locale.setSysUserId(getSysUserId(request));
+            String id = supportedLocaleService.insert(locale);
+            locale.setId(id);
```
ibacher, line 119: These seems redundant no? Hibernate should set this on inert.

ibacher (approved):
Basically LGTM with one remark.

ibacher:
@mozzy11 Even if I've approved something can we hold off on merging it in? There was a pending minor comment, but now that needs to be done in a follow-up PR.

reagan-meant:
Follow up commit [here](https://github.com/DIGI-UW/OpenELIS-Global-2/pull/3057)

PR #3011, opened 2026-03-05
Fix: return 400 Bad Request for non-numeric path IDs in storage endpo…

### LINKED ISSUE #3009: Bug: Non-numeric room ID returns HTTP 500 instead of 400 in storage REST endpoint
## Description

The storage REST endpoint for retrieving a room by ID does not properly validate the `id` path parameter. When a non-numeric value is provided (e.g., `"abc"`), the backend attempts to parse the value using `Integer.parseInt(id)`. This causes a `NumberFormatException`, which is caught by a generic exception handler and returned as an **HTTP 500 Internal Server Error**.

Since this is caused by invalid client input, the endpoint should return **HTTP 400 Bad Request** instead of **500 Internal Server Error**.

Returning a 500 status misrepresents the issue as a server failure rather than a client input error.

---

## Endpoint

```
GET /api/OpenELIS-Global/rest/storage/rooms/{id}
```

---

## Steps to Reproduce

1. Log in to the application.
2. Send the following request:

```
GET https://localhost/api/OpenELIS-Global/rest/storage/rooms/abc
```

3. Observe the server response.

---

## Actual Result

```
HTTP 500 Internal Server Error
```

---

## Expected Result

```
HTTP 400 Bad Request
```

The API should detect invalid input and return a client error instead of a server error.

---

## Root Cause

The controller attempts to parse the path parameter directly:

```java
Integer idInt = Integer.parseInt(id);
```

When the value is not numeric, a `NumberFormatException` occurs and is caught by a generic exception block:

```java
catch (Exception e) {
    return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
}
```

Ayushdk:
## Pull Requests Requirements

* [x] The PR title includes a brief description of the work done, including the Issue number if applicable.
* [x] The PR title follows conventional commit label standards.
* [x] The changes conform to the OpenELIS Global Styleguide.
* [x] The changes are validated manually through API testing.
* [x] I have read and agree to the Contributing Guidelines of this project.

## Summary

This PR fixes an issue in the storage REST API where invalid numeric path parameters
(e.g. `/rest/storage/rooms/abc`) caused a `500 Internal Server Error`.

The endpoint previously parsed the `id` path variable using:

```
Integer.parseInt(id)
```

If a non-numeric value was provided, a `NumberFormatException` was thrown and
caught by a generic `Exception` handler, resulting in a `500` response.

This behavior is incorrect because the error is caused by invalid client input.

### Fix implemented

* Added explicit handling for `NumberFormatException`.
* Invalid numeric IDs now return **400 Bad Request**.
* Existing behavior for valid requests and server errors remains unchanged.

### Example

Before:

```
GET /rest/storage/rooms/abc
→ 500 Internal Server Error
```

After:

```
GET /rest/storage/rooms/abc
→ 400 Bad Request
```

## Related Issue

Fixes: #3009 

## Other

The fix was verified locally using Postman:

```
GET https://localhost/api/OpenELIS-Global/rest/storage/rooms/abc
```

Result:

```
400 Bad Request
```


vsvishalsharma (approved):
LGTM

Ayushdk:
@mozzy11 could you review it , please ?

Ayushdk:
@mozzy11 @pmanko could you review it , please ?

mozzy11 (commented):
Still a valid fix — on current develop, StorageLocationRestController still parses path IDs with raw Integer.parseInt and the generic catch (Exception) blocks turn a non-numeric ID into a 500 rather than a 400. The controller has grown quite a few endpoints since your branch (it's now ~600 lines with rooms/devices/shelves/racks), so please rebase and make sure the 400 handling covers the newer endpoints too, then we'll re-run CI. Your original run was fully green, so this should be straightforward to bring home.

Agaba-derrick:
@Ayushdk are you able to adress the above reviews for a merge ?

Ayushdk:
Hi @Agaba-derrick, thanks for the follow-up. I haven't had a chance to address the requested changes yet, but I'm planning to rebase the branch onto the latest `develop` and update the fix to cover the newer storage endpoints this week. I'll push the changes and request another review once they're ready. Thanks for your patience!

PR #3080, opened 2026-03-16
test: Add unit tests for Patient valueholder entity

manroop-2005:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Adds 

PatientTest.java
 — a comprehensive set of pure unit tests for the 

Patient
 valueholder entity (

patient/valueholder/Patient.java
), which previously had zero test coverage.

These tests run entirely in-memory with no Spring context and no Docker required, making them fast and CI-friendly.

## Related Issue
[ https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3255 ]

### DIFF
--- a/src/test/java/org/openelisglobal/patient/valueholder/PatientTest.java
+++ b/src/test/java/org/openelisglobal/patient/valueholder/PatientTest.java
@@ -90,8 +90,7 @@ public class PatientTest {
         assertEquals("Original ambiguous display value should be preserved", "xx/xx/2024",
                 patient.getBirthDateForDisplay());
         assertEquals("Ambiguous month and day should normalize to configured placeholder values",
-                LocalDate.of(2024, 1, 1),
-                patient.getBirthDate().toLocalDateTime().toLocalDate());
+                LocalDate.of(2024, 1, 1), patient.getBirthDate().toLocalDateTime().toLocalDate());
     }
 
     @Test
@@ -109,7 +108,8 @@ public class PatientTest {
         patient.setBirthTimeForDisplay("03/05/2024");
 
         assertEquals("Display value should be preserved", "03/05/2024", patient.getBirthTimeForDisplay());
-        assertEquals("Parsed SQL date should match the display date", Date.valueOf("2024-03-05"), patient.getBirthTime());
+        assertEquals("Parsed SQL date should match the display date", Date.valueOf("2024-03-05"),
+                patient.getBirthTime());
     }
 
     @Test
@@ -127,7 +127,8 @@ public class PatientTest {
         patient.setDeathDateForDisplay("04/06/2024");
 
         assertEquals("Display value should be preserved", "04/06/2024", patient.getDeathDateForDisplay());
-        assertEquals("Parsed SQL date should match the display date", Date.valueOf("2024-04-06"), patient.getDeathDate());
+        assertEquals("Parsed SQL date should match the display date", Date.valueOf("2024-04-06"),
+                patient.getDeathDate());
     }
 
     @Test
--- a/src/test/java/org/openelisglobal/patient/valueholder/PatientTest.java
+++ b/src/test/java/org/openelisglobal/patient/valueholder/PatientTest.java
@@ -16,7 +16,8 @@ import org.openelisglobal.person.valueholder.Person;
 /**
  * Pure unit tests for the Patient valueholder.
  *
- * <p>These tests require NO Spring context and NO Docker — they validate the
+ * <p>
+ * These tests require NO Spring context and NO Docker — they validate the
  * entity's field behaviour, date-display synchronisation, FHIR UUID handling,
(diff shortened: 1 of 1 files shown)


Rohanbhalerao875 (changes requested):
Thanks for adding unit test coverage for the Patient valueholder — this is a good step forward, especially since it previously had no tests.

However, I have a few concerns before approving:

The tests rely heavily on reflection to set private fields (e.g., birthDate, birthTime, deathDate). This bypasses the actual setter logic and makes the tests fragile and less meaningful. Ideally, tests should validate behavior through public APIs rather than internal implementation details.

Some tests intentionally avoid calling methods that use DateUtil. While I understand the intent to keep tests lightweight, this means we're not actually validating the real behavior of those methods.

Many tests focus on simple getter/setter verification, which adds limited value. It would be more useful to focus on behavior or logic (e.g., transformations, constraints, or derived values).

Overall, the direction is good, but I’d recommend reducing reflection usage and focusing more on testing real behavior rather than internal state.

mozzy11 (commented):
Thanks for adding coverage to the Patient valueholder — it had none, and the ambiguous-date test in particular is valuable. I checked develop and this test class doesn't clash with anything; all the Patient APIs you exercise still exist, so this is still relevant.

Two things before we can merge:
1. the unit-test CI job is failing on this PR — please rebase onto the latest develop (CI has been reworked since) and get it green;
2. the `@BeforeClass` sets global static state (SpringContext application context and MessageUtil message source) with **no teardown**, which can leak mocks into other tests running in the same JVM and cause order-dependent failures. Please capture and restore (or reset) those statics in an `@AfterClass`.

With those fixed I'd be happy to merge this.

src/test/java/org/openelisglobal/patient/valueholder/PatientTest.java
```diff
@@ -0,0 +1,170 @@
...
+        when(configuration.getPropertyValue(Property.DEFAULT_LANG_LOCALE)).thenReturn("en-US");
+        when(configuration.getPropertyValue(Property.StringContext)).thenReturn(null);
+
+        AutowireCapableBeanFactory beanFactory = mock(AutowireCapableBeanFactory.class);
+        when(beanFactory.getBean(DefaultConfigurationProperties.class)).thenReturn(configuration);
+
+        ApplicationContext applicationContext = mock(ApplicationContext.class);
+        when(applicationContext.getAutowireCapableBeanFactory()).thenReturn(beanFactory);
+
+        new SpringContext().setApplicationContext(applicationContext);
```
mozzy11, line 46: This swaps the process-global static `SpringContext` (and `MessageUtil` below) for mocks with no `@AfterClass` restore — every test class that runs afterward in the same surefire JVM sees a context that only resolves `DefaultConfigurationProperties`. It's also fragile in reverse: `DateUtil` caches `AMBIGUOUS_DATE_CHAR` and the `ConfigurationProperties` singleton statically, so if another class loads them first these mocks never take effect. Capture and restore the previous context/messageSource in `@AfterClass`, or move the date tests into a `DateUtil`-focused test that manages this shared fixture the way the existing barcode tests do.

src/test/java/org/openelisglobal/patient/valueholder/PatientTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        patient.setFhirUuid(uuid);
+
+        assertSame("FHIR UUID getter should return the same UUID instance", uuid, patient.getFhirUuid());
+        assertEquals("FHIR UUID string should use canonical format", "550e8400-e29b-41d4-a716-446655440000",
+                patient.getFhirUuidAsString());
+    }
+
+    @Test
+    public void setPerson_ShouldReplaceCurrentPersonAssociation() {
```
mozzy11, line 146: `setPerson_ShouldReplaceCurrentPersonAssociation` (and `mergeFields_ShouldRetainMergeStateMetadata`, plus the `assertSame` half of the FHIR UUID test) only verify that a plain field assignment sticks — setter-then-getter tautologies with no logic under test (Test Quality Invariants V.6). Consider dropping them and keeping the genuinely behavioral date-conversion and `getFhirUuidAsString` tests.

mozzy11 (changes requested):
Thanks @manroop-2005 — the date-handling tests here have real value: the display↔timestamp round-trips and ambiguous-date normalization genuinely exercise `DateUtil` logic and would fail on regression.

Two blockers: (1) **CI is failing** on `checkFormat-build-unitTest-and-run` and the E2E job (logs have expired, so please push a rebase to get fresh runs). (2) The `@BeforeClass` swaps the process-global `SpringContext` and `MessageUtil` for mocks and never restores them — leaking a mocked context into every test class that runs later in the same surefire JVM. That's the classic order-dependent contamination this repo has been bitten by, and is the most likely cause of the CI failure. Details inline.

PR #3083, opened 2026-03-17
fix: patient search uses wrong parameter for external ID

alikhere:
## PR DESCRIPTION:
Was digging into the patient search code and found that external ID search
was silently using nationalID as the bind parameter in both query methods
inside LuceneSearchResultsDAOImpl. So searching by external ID was actually
filtering by national ID the whole time. Changed both occurrences to bind
externalID instead.

One file changed, no new logic added.

## Related Issue
Closes #3082


mozzy11 (commented):
Great catch — this bug is still live on current develop: both query methods in LuceneSearchResultsDAOImpl bind nationalID to EXTERNAL_ID_PARAM (lines 76 and 162), so external-ID patient search has been silently filtering by national ID all along, and your two-line fix applies cleanly. No CI ever ran on this branch, so we'll approve a run; if you can, adding a small regression test that searches by external ID would make this bulletproof. We'd like to get this merged promptly.

Issue #3113, opened 2026-03-20
Missing password confirmation validation and password handling concerns in ChangePasswordLoginForm


Harishvishwa01:
**Description:**

The `ChangePasswordLoginForm` currently validates individual password fields using the `@Password` annotation, but it does not enforce that `newPassword` and `confirmPassword` match.

Additionally, password handling lacks clarity regarding leading/trailing spaces, which may lead to inconsistent user experience or authentication issues.

---

**Issues Identified:**

1. **Missing Password Match Validation (Critical)**

   * There is no validation ensuring that `newPassword` and `confirmPassword` are identical.
   * This may allow inconsistent or unintended password updates if not validated elsewhere.

2. **Missing Required Field Validation**

   * Password fields (`password`, `newPassword`, `confirmPassword`) do not explicitly enforce non-empty input.
   * Depending on the implementation of `@Password`, empty values may pass validation.

3. **Password Handling (Whitespace Concern)**

   * Passwords are currently accepted without checks for leading/trailing spaces.
   * Trimming passwords is not recommended, as passwords should be treated as exact user input.
   * However, lack of validation may lead to accidental whitespace causing login issues.

---

**Expected Behavior:**

* `newPassword` and `confirmPassword` must match before processing password change.
* Password fields should not accept blank values.
* Password handling should be explicit:

  * Either reject leading/trailing spaces
  * Or clearly allow them without modification

---

**Suggested Fix:**

* Extend the existing `ChangePasswordLoginFormValidator` to:

  * Validate that `newPassword` and `confirmPassword` match
  * Add required field validation using `ValidationUtils`
  * Add validation to handle leading/trailing whitespace in passwords

---

**Optional Improvements:**

* Normalize input fields like `loginName` using trimming
* Consider documenting password policy for better clarity and maintainability

---

**Impact:**

* Prevents inconsistent password updates
* Improves validation robustness
* Enhances security and user experience

vaalemax:
Hi, I would like to personally look into the issue. I’ll get back to you with a PR as soon as possible.

vaalemax:
Hi, I've opened a PR addressing this issue. 
The fix extends ChangePasswordLoginFormValidator with blank field validation, whitespace handling and an early return to prevent null errors. 
Happy to receive feedback.

PR #3114, opened 2026-03-20
 issue-3113: Improve ChangePasswordLoginForm validation

Harishvishwa01:
---

**Related Issue:** #3113 

---

**Description:**

This PR improves validation logic in `ChangePasswordLoginFormValidator` by adding missing checks and improving robustness.

---

**Changes Made:**

* Added required field validation using `ValidationUtils`
* Added validation to ensure new password and confirm password match
* Prevented reuse of old password
* Added validation to reject leading/trailing whitespace in passwords
* Improved null safety

---

**Why This Change:**

* Prevents invalid or inconsistent password updates
* Improves user experience and validation reliability
* Aligns with existing project validation patterns

---

**Testing:**

* Verified empty fields are rejected
* Verified mismatched passwords are rejected
* Verified same old/new password is rejected
* Verified whitespace in passwords is rejected

---


src/main/java/org/openelisglobal/login/validator/ChangePasswordLoginFormValidator.java
```diff
@@ -17,11 +18,36 @@ public boolean supports(Class<?> clazz) {
...
         }
-        if (!form.getNewPassword().equals(form.getConfirmPassword())) {
+
+        // 3. Confirm password match
+        if (!newPassword.equals(confirmPassword)) {
             errors.reject("login.error.password.notmatch");
         }
+
+        //  4. Prevent leading/trailing spaces (without trimming)
+        if (!password.equals(password.trim()) ||
```
sagarkhandagre998, line 45: do we need to check for password field here ?

src/main/java/org/openelisglobal/login/validator/ChangePasswordLoginFormValidator.java
```diff
@@ -17,11 +18,36 @@ public boolean supports(Class<?> clazz) {
     public void validate(Object target, Errors errors) {
         ChangePasswordLoginForm form = (ChangePasswordLoginForm) target;
 
-        if (form.getPassword().equals(form.getNewPassword())) {
-            errors.reject("login.error.newpassword.required", "New password cannot match old password");
+        //  1. Required field validation (consistent with project)
+        ValidationUtils.rejectIfEmptyOrWhitespace(errors, "password", "login.password.required");
```
sagarkhandagre998, line 22: do we required validation for password field?

sagarkhandagre998 (changes requested):

Agaba-derrick:
can you confirm that this issue was not solved earlier !

mozzy11 (commented):
The validator gap you're addressing is real — ChangePasswordLoginFormValidator on develop still only checks old-vs-new and confirm-match, so the required-field and whitespace validations are a genuine improvement. Three things are needed to merge: fix the failing 'Build + Test' check (the file is missing its trailing newline, so run mvn spotless:apply); add the new message keys (login.password.required, login.newpassword.required, login.confirmpassword.required, login.error.password.whitespace) to message_en.properties since they don't exist yet and would render as raw codes; and prefer errors.rejectValue(...) for the field-level checks. Also note your #3119 currently carries this same validator diff — please drop it there so this PR is the single home for the change.

Agaba-derrick:
@sagarkhandagre998 Are you able to address the reviews above before the Pr is marked  as stale which would automatically  close this Pr

PR #3102, opened 2026-03-19
test(PatientDAO): add DAO-layer integration tests for issue #3092

noel-mugisha:
## Summary

Closes #3092

Adds `PatientDAOTest.java` — integration tests for the `PatientDAO` layer 
using the project's existing DBUnit + `BaseWebContextSensitiveTest` pattern.

## What is tested

**CRUD & lookups (18 tests total):**
- `getAllPatients` — verifies all dataset records are returned
- `getData(String)` — valid ID returns correct patient, non-existent ID returns null
- `readPatient` — patient loads with person eagerly
- `externalIDExists` — true for known ID, false for unknown
- `getPatientByNationalId` — correct patient returned, null for unknown
- `getPatientsByNationalId` — correct list returned, empty list for unknown
- `getPatientByExternalId` — correct patient returned, null for unknown
- `getPatientByPerson` — patient correctly linked to person
- `getPageOfPatients` — first page is non-empty
- `getAllMissingFhirUuid` — all 4 returned when none assigned; count drops to 3 after one UUID is set

**Edge cases (2 tests):**
- Special characters in names (`José`, `O'Brien`) — stored and retrieved without corruption
- UUID-formatted external ID that doesn't exist — returns null (FHIR integration scenario)

18/18 tests pass.

PR #3122, opened 2026-03-21
test(PatientService): moved unique DAO edge-case tests into PatientServiceTest

noel-mugisha:
Per review feedback on #3102 — PatientDAOTest duplicated tests already
covered by PatientServiceTest. Removed PatientDAOTest.java and moved 
edge-case tests directly into PatientServiceTest:

- getData returns null for non-existent ID
- externalIDExists returns false for unknown external ID
- getPatientByNationalId returns null for unknown national ID
- getPatientsByNationalId returns empty list for unknown ID
- getAllMissingFhirUuid returns all patients when none have UUID assigned
- special characters in names (José, O'Brien) persist without corruption
- UUID-formatted external ID lookup returns null when not found

Closes issue#3092"


noel-mugisha:
@Agaba-derrick The missing tests were added, check it out

Rohanbhalerao875 (changes requested):
The PR description says "Removed PatientDAOTest.java" but the diff only shows 1 file changed with 62 additions. Since PR #3102 was closed without merging, PatientDAOTest.java never existed on develop. Please update the PR description to accurately reflect what this PR does: it adds 7 new edge-case/negative-path tests to PatientServiceTest.

noel-mugisha:
PatientDAOTest.java  existed on PR #3102, I was asked to remove it and add the 7 new edge cases.

Rohanbhalerao875:
got it

Rohanbhalerao875:
Good coverage of edge cases and null/empty scenarios

Quick question: for assertions like expecting exactly 4 patients (getAllMissingFhirUuid), is that guaranteed by the test setup/dataset? Just wondering if this could become brittle if the dataset changes.

Maybe worth making the setup explicit or relaxing the assertion if needed.

Overall, solid tests.

noel-mugisha:
Thanks for the feedback! 

Yes, the assertion of exactly 4 patients in is fully guaranteed by the test setup — the **@Before** method calls executeDataSetWithStateManagement("testdata/patient.xml") which resets the database to a clean, known state before every test run. That dataset defines exactly 4 patients, so the count is deterministic and will never be affected by other tests or leftover data. 

The existing **getAllMissingFhirUuid_shouldGetAllPatientsMissingFhirUuid** test in this same file follows the same pattern and relies on the same dataset count, so this is consistent with the established testing approach across the suite. If the dataset ever changes, both tests would need updating together — which is intentional, since the tests are meant to reflect the known data state.

noel-mugisha:
Hey @pmanko, this PR has been ready for a few weeks with all checks passing. Can you take a look and approve when you get a chance? Happy to make any changes if needed. Thank you!

mozzy11 (commented):
This closes the loop on the #3102 review feedback nicely — PatientDAOTest is gone from develop and PatientServiceTest still lacks the negative edge cases (unknown external ID, unknown national ID, getData on a non-existent ID), so these additions are real coverage rather than duplication. The PR is mergeable and the last CI run was fully green across all 8 checks. A quick rebase onto current develop and a fresh CI run (the test base class was hardened since April) and this should be ready to merge.

noel-mugisha:
@mozzy11 , rebase done

PR #3153, opened 2026-03-24
Ahri/inventory biorepository fixes

dawit-Tegegnwork:
## Summary
This PR includes 3 AHRI fixes:

1. Inventory report generation/export fix
2. Biorepository manifest sample type fix
   - accepts broader AHRI specimen types
   - creates new valid sample types with warnings
   - keeps duplicate barcode protection
3. Continuous box numbering for biorepository storage
   - numeric boxes now show 1, 2, 3... instead of 1-1, 1-2...

## Validation
- Tested locally on the Docker dev stack
- Verified inventory report generation works
- Verified biorepository manifest validation/import accepts DNA, RNA, and other new sample types
- Verified duplicate barcodes are still blocked
- Verified numeric storage boxes render continuous numbering

### DIFF
--- a/frontend/src/components/storage/StorageDashboard.jsx
+++ b/frontend/src/components/storage/StorageDashboard.jsx
@@ -2906,6 +2906,11 @@ const StorageDashboard = () => {
       Object.keys(occupiedCoordinates).some((coordinate) =>
         /^\d+-\d+$/.test(coordinate),
       );
+    const hasLegacyContinuousCoordinates =
+      hint === "number-number" &&
+      Object.keys(occupiedCoordinates).some((coordinate) =>
+        /^\d+$/.test(coordinate),
+      );
 
     // Generate coordinate labels based on position schema hint
     const getCoordinateLabel = (rowIdx, colIdx) => {
@@ -2915,13 +2920,21 @@ const StorageDashboard = () => {
         return `${letter}${colIdx + 1}`;
       }
 
-      if (hasLegacyNumberCoordinates) {
-        // Backward compatibility for boxes that already stored positions as 1-1, 1-2, ...
+      if (hint === "number-number" && !hasLegacyContinuousCoordinates) {
+        // Existing number-number schema uses row-column coordinates like 1-1, 1-2, ...
         return `${rowIdx + 1}-${colIdx + 1}`;
       }
 
-      // Continuous numbering for AHRI biorepository boxes: 1, 2, 3, ... in row-major order.
-      return String(rowIdx * (selectedBox.columns || 0) + colIdx + 1);
+      if (hint === "continuous" || hasLegacyContinuousCoordinates) {
+        // Backward compatibility for boxes created before the explicit continuous schema type existed.
+        return String(rowIdx * (selectedBox.columns || 0) + colIdx + 1);
+      }
+
+      if (hasLegacyNumberCoordinates) {
+        return `${rowIdx + 1}-${colIdx + 1}`;
+      }
+
+      return `${rowIdx + 1}-${colIdx + 1}`;
     };
 
     return (
--- a/frontend/src/components/storage/StorageDashboard.jsx
+++ b/frontend/src/components/storage/StorageDashboard.jsx
@@ -2900,17 +2900,28 @@ const StorageDashboard = () => {
     // occupiedCoordinates is now an object: { "A1": { externalId: "...", sampleItemId: "..." }, ... }
     const occupiedCoordinates = selectedBox.occupiedCoordinates || {};
 
+    const hint = selectedBox.positionSchemaHint || "letter-number";
+    const hasLegacyNumberCoordinates =
+      hint === "number-number" &&
+      Object.keys(occupiedCoordinates).some((coordinate) =>
+        /^\d+-\d+$/.test(coordinate),
+      );
+
     // Generate coordinate labels based on position schema hint
     const getCoordinateLabel = (rowIdx, colIdx) => {
-      const hint = selectedBox.positionSchemaHint || "letter-number";
       if (hint === "letter-number") {
         // A1, A2, ..., B1, B2, etc.
         const letter = String.fromCharCode(65 + rowIdx); // A=65
         return `${letter}${colIdx + 1}`;
-      } else {
-        // 1-1, 1-2, etc.
+      }
+
+      if (hasLegacyNumberCoordinates) {
+        // Backward compatibility for boxes that already stored positions as 1-1, 1-2, ...
         return `${rowIdx + 1}-${colIdx + 1}`;
       }
+
+      // Continuous numbering for AHRI biorepository boxes: 1, 2, 3, ... in row-major order.
+      return String(rowIdx * (selectedBox.columns || 0) + colIdx + 1);
     };
 
     return (
--- a/frontend/src/components/storage/StorageDashboard.jsx
+++ b/frontend/src/components/storage/StorageDashboard.jsx
@@ -2901,11 +2901,6 @@ const StorageDashboard = () => {
     const occupiedCoordinates = selectedBox.occupiedCoordinates || {};
 
     const hint = selectedBox.positionSchemaHint || "letter-number";
-    const hasLegacyNumberCoordinates =
-      hint === "number-number" &&
-      Object.keys(occupiedCoordinates).some((coordinate) =>
-        /^\d+-\d+$/.test(coordinate),
-      );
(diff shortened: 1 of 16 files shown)


HerbertYiga:
hi @mherman22 @samuelmale how does this look to you?

src/main/java/org/openelisglobal/inventory/controller/rest/InventoryReportRestController.java
```diff
@@ -0,0 +1,48 @@
...
+import org.openelisglobal.inventory.service.InventoryReportService.GeneratedReport;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/rest/inventory/reports")
```
samuelmale, line 16: @dawit-Tegegnwork Where in the frontend are you consuming this new endpoint?

dawit-Tegegnwork: @samuelmale  The new endpoint is consumed through the existing inventory frontend service in frontend/src/components/inventory/InventoryService.js, where ReportsAPI.generate() calls /rest/inventory/reports/generate.
It is used by the existing inventory report UIs here:
frontend/src/components/inventory/InventoryReports.jsx
frontend/src/components/inventory/InventoryReportsModal.jsx
So this PR does not introduce a brand new UI entry point. It fixes the backend report generation path and updates the existing frontend report flow to download the generated file correctly.

Specifically, both components call ReportsAPI.generate(reportParams) and then download the returned blob response.

frontend/src/components/storage/StorageDashboard.jsx
```diff
@@ -2900,17 +2900,28 @@ const StorageDashboard = () => {
...
-        // 1-1, 1-2, etc.
+      }
+
+      if (hasLegacyNumberCoordinates) {
+        // Backward compatibility for boxes that already stored positions as 1-1, 1-2, ...
         return `${rowIdx + 1}-${colIdx + 1}`;
       }
+
+      // Continuous numbering for AHRI biorepository boxes: 1, 2, 3, ... in row-major order.
+      return String(rowIdx * (selectedBox.columns || 0) + colIdx + 1);
```
samuelmale: @dawit-Tegegnwork This logic makes the storage system aware of "AHRI" making it implementation specific. It also overloads the "number-number" schema hit. Let's use a less invasive approach by adding a "continuous" type to the [existing schema items](https://github.com/dawit-Tegegnwork/OpenELIS-Global-2/blob/develop/frontend/src/components/storage/LocationManagement/EditBoxModal.jsx#L93-L111). In the future, we need to use a more configurable solution.

src/main/java/org/openelisglobal/biorepository/valueholder/BiorepositoryApprovedSampleType.java
```diff
@@ -27,6 +28,7 @@
  */
 @Entity
 @Table(name = "biorepository_approved_sample_type", schema = "clinlims")
+@AttributeOverride(name = "lastupdated", column = @Column(name = "lastupdated"))
```
samuelmale: @dawit-Tegegnwork Instead of overriding the column name, let's make sure the associated tables or DAO layers are using the correct column name `last_updated`.

src/main/java/org/openelisglobal/inventory/service/InventoryReportServiceImpl.java
```diff
@@ -0,0 +1,566 @@
...
+import org.openelisglobal.common.log.LogEvent;
+import org.openelisglobal.inventory.valueholder.InventoryItem;
+import org.openelisglobal.inventory.valueholder.InventoryLot;
+import org.openelisglobal.inventory.valueholder.InventoryTransaction;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+public class InventoryReportServiceImpl implements InventoryReportService {
```
samuelmale, line 49: Is it possible to add some tests for this service?

frontend/src/components/storage/StorageDashboard.jsx
dawit-Tegegnwork: That makes sense. I agree I’ll update it to introduce a separate continuous schema type instead

src/main/java/org/openelisglobal/biorepository/valueholder/BiorepositoryApprovedSampleType.java
dawit-Tegegnwork: Understood. I used the override as the smallest immediate fix for the runtime issue, but I agree the better fix is to align the underlying mapping with the actual column name. I’ll trace the associated table/DAO usage and correct it there instead of keeping the entity override.

src/main/java/org/openelisglobal/inventory/service/InventoryReportServiceImpl.java
dawit-Tegegnwork: Yes, I can add tests for this service.

dawit-Tegegnwork:
I pushed an update addressing the review comments:
- added an explicit continuous storage schema type instead of overloading number-number
- replaced the entity override with a Liquibase column rename for biorepository_approved_sample_type.last_updated
- added focused tests for InventoryReportServiceImpl

Please take another look when you have time.

frontend/src/components/storage/StorageDashboard.jsx
```diff
@@ -2906,6 +2906,11 @@ const StorageDashboard = () => {
       Object.keys(occupiedCoordinates).some((coordinate) =>
         /^\d+-\d+$/.test(coordinate),
       );
+    const hasLegacyContinuousCoordinates =
```
samuelmale: Do you still need this?

dawit-Tegegnwork: OK,The explicit `continuous` schema should stay. I only added this for backward compatibility with earlier local boxes, but I agree that should be handled by migration instead of runtime logic. I’ll remove this fallback.

dawit-Tegegnwork: Removed the fallback and pushed the update. The grid now only treats `continuous` as continuous.

frontend/src/components/storage/StorageDashboard.jsx
```diff
@@ -2900,17 +2900,34 @@ const StorageDashboard = () => {
     // occupiedCoordinates is now an object: { "A1": { externalId: "...", sampleItemId: "..." }, ... }
     const occupiedCoordinates = selectedBox.occupiedCoordinates || {};
 
+    const hint = selectedBox.positionSchemaHint || "letter-number";
+    const hasLegacyNumberCoordinates =
```
samuelmale: I think we don't need this.

frontend/src/components/storage/StorageDashboard.jsx
```diff
@@ -2900,17 +2900,34 @@ const StorageDashboard = () => {
...
+      if (hint === "number-number") {
+        // Existing number-number schema uses row-column coordinates like 1-1, 1-2, ...
+        return `${rowIdx + 1}-${colIdx + 1}`;
+      }
+
+      if (hint === "continuous") {
+        return String(rowIdx * (selectedBox.columns || 0) + colIdx + 1);
+      }
+
+      if (hasLegacyNumberCoordinates) {
```
samuelmale: Ditto

frontend/src/components/storage/StorageDashboard.jsx
dawit-Tegegnwork: Understood. I’ll remove this too.

frontend/src/components/storage/StorageDashboard.jsx
dawit-Tegegnwork: Understood. I’ll remove it

dawit-Tegegnwork: Removed this as well and pushed the update.

frontend/src/components/storage/StorageDashboard.jsx
dawit-Tegegnwork: Removed this and pushed the update.

PR #3161, opened 2026-03-24
fix(ci): checkout PR code in analyzer harness E2E + mount bridge volume

pmanko:
## Summary
- Pass `head_sha` to the reusable analyzer harness workflow so it checks out **the PR's code** (compose files, test scripts, fixtures) instead of develop. The `workflow_run` trigger defaults to the default branch checkout, which means PRs that modify CI compose config never see their changes in the cascaded E2E.
- Mount `analyzer-imports` volume on the bridge container in CI compose. The bridge needs this to detect files dropped by E2E tests. Was only mounted on OE; bridge registration failed with HTTP 500.

## Test plan
- [ ] Existing E2E tests continue to pass (no behavioral change for PRs that don't modify CI files)
- [ ] PRs that modify `.github/ci/ci.analyzer-harness.yml` will now use their own compose config in the harness E2E


.github/workflows/e2e-playwright-analyzer-harness-reusable.yml
```diff
@@ -40,6 +45,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
+          ref: ${{ inputs.head_sha || github.sha }}
```
pmanko, line 48: Copilot's concern about image/checkout SHA mismatch is addressed: `head_sha` comes from the setup job's build context — the same SHA the shared build used to build Docker images. Checkout and images are consistent.

The suggested `ref: ${{ github.sha }}` would revert to the default branch checkout (develop) — exactly the bug this PR fixes.

This aligns with the Playwright core pattern at line 271 of e2e-tests.yml, which uses `workflow_run.head_sha` for non-fork PRs.

.github/workflows/e2e-tests.yml
```diff
@@ -421,6 +421,7 @@ jobs:
     with:
       build_run_id: ${{ needs.setup.outputs.build_run_id }}
       is_fork: ${{ needs.setup.outputs.is_fork }}
+      head_sha: ${{ needs.setup.outputs.head_sha }}
```
pmanko, line 424: Same answer as above — `head_sha` is resolved from the build context artifact in the setup job. It's the same SHA the shared build step checked out to build images. No mismatch.

Passing `workflow_run.head_sha` directly (as suggested) would also work, but `head_sha` from setup handles edge cases (e.g. push events where head_sha falls back to `workflow_run.head_sha` — see line 73-74).

PR #3187, opened 2026-03-25
fix(ci): pre-create incoming dirs before bridge starts

pmanko:
## Summary
Pre-create `analyzer-imports/{slug}/incoming` directories before `docker compose up`, eliminating the async race where the bridge creates them after seed registration.

The 30s timeout was failing ~75% of E2E runs. This fix pre-creates the dirs and increases the safety-net timeout to 120s.

## Test plan
- [ ] E2E harness Demo shards pass reliably


pmanko:
Closing — the dir pre-creation fix is already in #3035, and #3188 covers the chmod restore + submodule pointers.

PR #2935, opened 2026-02-25
fix build warnings at the start of the build

Agaba-derrick:
# Pull Requests Requirements

- [ ] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary
Correct invalid PostgreSQL dependency scope and replace dynamic project version expression in pom.xml to eliminate Maven build warnings.
## Screenshots

[Add relevant screenshots here if applicable]

## Related Issue

Closes #2934 

## Other

[Add any additional information or notes here]

PR #3225, opened 2026-03-29
fix: resolve Maven build warnings in pom.xml and Java source files (#2934)

### LINKED ISSUE #2934: Fix Maven build warnings in pom.xml
Running `mvn clean install` produces Maven build warnings caused by an invalid dependency scope and a version defined using expressions in pom.xml.  

These warnings do not break the build but may affect compatibility with future Maven versions.we can fix PostgreSQL dependency scope and stabilizing the codebase version definition to remove the warnings.

Piyush-churhe:
## Issue
Fixes #2934 — Fix Maven build warnings in pom.xml

## Description
Resolved multiple Maven build warnings that appear during `mvn clean install`:

1. **pom.xml**: Fixed invalid dependency scope `scope` → `runtime` for 
   postgresql dependency (line 648)
2. **StringUtil.java**: Replaced deprecated `new Double(String)` constructor 
   with `Double.valueOf()` (line 627)
3. **PanelSortOrderComparator.java**: Replaced deprecated `new Integer(int)` 
   constructors with `Integer.valueOf()` (line 33)
4. **SiteBrandingService.java**: Added missing `@Deprecated` annotation on 
   `validateColor()` method (line 43)

## How to Test
1. Run `mvn clean install -DskipTests 2>&1 | grep -i "warning"`
2. Verify the following warnings no longer appear:
   - `'dependencies.dependency.scope' for org.postgresql:postgresql must be one of [provided, compile, runtime, test, system]`
   - `Double(java.lang.String) has been deprecated and marked for removal`
   - `Integer(int) has been deprecated and marked for removal`
   - `deprecated item is not annotated with @Deprecated`

## Files Changed
- `pom.xml`
- `src/main/java/org/openelisglobal/common/util/StringUtil.java`
- `src/main/java/org/openelisglobal/panel/valueholder/PanelSortOrderComparator.java`
- `src/main/java/org/openelisglobal/sitebranding/service/SiteBrandingService.java`

## Screenshots
N/A (build warning fixes, no UI changes)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -645,7 +645,7 @@
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
             <version>${postgresql.version}</version>
-            <scope>scope</scope>
+            <scope>runtime</scope>
         </dependency>
         <dependency>
             <groupId>org.quartz-scheduler</groupId>
--- a/src/main/java/org/openelisglobal/common/util/StringUtil.java
+++ b/src/main/java/org/openelisglobal/common/util/StringUtil.java
@@ -3,15 +3,19 @@
  * you may not use this file except in compliance with the License. You may obtain a copy of the
  * License at http://www.mozilla.org/MPL/
  *
- * <p>Software distributed under the License is distributed on an "AS IS" basis, WITHOUT WARRANTY OF
- * ANY KIND, either express or implied. See the License for the specific language governing rights
- * and limitations under the License.
+ * <p>
+ * Software distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
+ * the specific language governing rights and limitations under the License.
  *
- * <p>The Original Code is OpenELIS code.
+ * <p>
+ * The Original Code is OpenELIS code.
  *
- * <p>Copyright (C) The Minnesota Department of Health. All Rights Reserved.
+ * <p>
+ * Copyright (C) The Minnesota Department of Health. All Rights Reserved.
  *
- * <p>Contributor(s): CIRG, University of Washington, Seattle WA.
+ * <p>
+ * Contributor(s): CIRG, University of Washington, Seattle WA.
  */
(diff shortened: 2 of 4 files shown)


mozzy11 (commented):
Thanks @Piyush-churhe — I verified all four warnings are still present on develop (the postgresql <scope>scope</scope> typo, the deprecated Double/Integer constructors, and the missing @Deprecated annotation), so this is still valuable. Two things before review: first, the branch is stale enough that no CI checks ever ran — please rebase onto latest develop and run mvn spotless:apply before pushing so the formatting gate passes. Second, heads-up that #2935 also fixes the pom scope typo; whichever lands first, the other will need a quick rebase, so keep an eye on it. Once CI is green this should merge easily.

pom.xml
```diff
@@ -645,7 +645,7 @@
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
             <version>${postgresql.version}</version>
-            <scope>scope</scope>
+            <scope>runtime</scope>
```
mozzy11, line 648: `runtime` removes the driver from the compile classpath, but main sources compile against it — `JsonBinaryType.java` and `JsonbObjectType.java` both `import org.postgresql.util.PGobject`. Today's invalid `scope` value falls back to compile, which is why the build works. Use `compile` (or drop the `<scope>` element entirely) instead of `runtime`.

mozzy11 (changes requested):
Re-verified on current develop: all four warnings are still present (the `<scope>scope</scope>` typo — now at pom.xml:698 — the deprecated `new Double`/`new Integer` constructors, and the missing `@Deprecated` on `SiteBrandingService.validateColor`), and #2935 has not merged, so this PR is still valuable. The Java-side changes look good.

One real defect blocks it (inline): `runtime` scope for the PostgreSQL driver will break compilation. Beyond that, the two earlier asks still stand — the branch has never had a CI run, so please rebase onto latest develop and run `mvn spotless:apply` before pushing so the formatting gate and the backend suite can execute.

PR #3231, opened 2026-03-29
test(address): added unit tests for address service layer

preetihonawad:
## Summary
Adds 4–5 new JUnit unit tests to each of the three address service test 
classes in the `org.openelisglobal.address` package. All existing tests 
are untouched and continue to pass.

### Files Changed
- `AddressPartServiceTest.java` — 4 new tests added
  - Insert persists and is retrievable by ID
  - Insert increases total count
  - DisplayOrder update persists correctly
  - Two separate inserts produce distinct IDs

- `OrganizationAddressServiceTest.java` — 3 new tests added
  - getAll returns all preloaded organization addresses
  - Update persists new value on reload
  - Unknown organization ID returns empty list (not null)

- `PersonAddressServiceTest.java` — 3 new tests added  
  - Unknown person ID returns empty list (not null)
  - Non-existent person/part combination returns null
  - Updating type field persists correctly

### Why these tests?
These tests improve coverage of edge cases that were previously untested:
- Null/empty returns for unknown IDs (guards against NPE in callers)
- Field-level persistence verification via re-fetch
- Insert contract validation (ID generation, count increase)

## Screenshots
No UI changes — this PR only adds backend test coverage.

## Checklist
- [x] PR title follows conventional commit label standards (`test(address):`)
- [x] Changes include new tests covering the address service layer
- [x] All existing tests are untouched and passing
- [x] Tests run: 23, Failures: 0, Errors: 0, Skipped: 0
- [x] I have read and agree to the Contributing Guidelines of this project

## Other
- No database schema changes
- No UI changes  
- No breaking changes to existing functionality
- Tests use the same pattern as existing OpenELIS tests 
  (`BaseWebContextSensitiveTest`, `@Before` dataset reset, JUnit 4)

PR #3234, opened 2026-03-29
test(menu): add unit tests for MenuRestController

preetihonawad:
## Summary
Adds 4 new JUnit integration tests to `MenuRestControllerTest.java` for the
`MenuController` REST endpoints. All existing 4 tests are untouched and 
continue to pass.

### Endpoints Tested
- `GET /rest/menu` — returns full menu tree
- `GET /rest/menu/{elementId}` — returns single menu item by ID
- `POST /rest/menu` — saves a list of menu items

### New Tests Added

| Test | What It Covers |
|------|---------------|
| `getMenuTree_shouldReturn200WithJsonContentType` | Endpoint returns HTTP 200 with correct `application/json` Content-Type |
| `postMenuTree_withEmptyList_shouldReturnEmptyList` | Posting an empty `[]` returns `[]` without error |
| `postMenuTree_withInactiveMenu_shouldPersistInactiveStatus` | `isActive=false` is not overridden during save |
| `postMenuTree_shouldReturnItemsInSameOrderAsSubmitted` | 3 items come back in exact same order with correct `presentationOrder` |

### Why These Tests?
These tests cover edge cases that were previously untested:
- Content-Type contract validation on the main GET endpoint
- Empty list handling on POST
- Inactive flag persistence — critical for correct menu rendering
- Order preservation — wrong order causes wrong UI display

## Screenshots
No UI changes — this PR only adds backend test coverage.

## Checklist
- [x] PR title follows conventional commit label standards (`test(menu):`)
  (`BaseWebContextSensitiveTest`, MockMvc, JUnit 4)
- [x] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0

## Other
- Tests use `MockMvc` via `BaseWebContextSensitiveTest` — same pattern
  as all other REST controller tests in the project
- Test data loaded from existing `testdata/menu.xml` dataset
- No new dependencies added

### DIFF
--- a/src/test/java/org/openelisglobal/address/AddressPartServiceTest.java
+++ b/src/test/java/org/openelisglobal/address/AddressPartServiceTest.java
@@ -72,4 +72,69 @@ public class AddressPartServiceTest extends BaseWebContextSensitiveTest {
         assertEquals("Village", part.getPartName());
         assertEquals("1", part.getDisplayOrder());
     }
+
+    @Test
+    public void insertAddressPart_shouldPersistAndBeRetrievableById() {
+        AddressPart part = new AddressPart();
+        part.setPartName("Pathway"); // ← change "Street" to "Pathway"
+        part.setDisplayOrder("10");
+        part.setSysUserId("1");
+
+        String insertedId = partService.insert(part);
+
+        assertNotNull("Inserted ID should not be null", insertedId);
+        AddressPart retrieved = partService.get(insertedId);
+        assertNotNull("Retrieved part should not be null", retrieved);
+        assertEquals("Pathway", retrieved.getPartName());
+        assertEquals("10", retrieved.getDisplayOrder());
+    }
+
+    @Test
+    public void insertAddressPart_shouldIncreaseTotalCount() {
+        int countBefore = partService.getAll().size();
+
+        AddressPart part = new AddressPart();
+        part.setPartName("Alley"); // short, under 20 chars
+        part.setDisplayOrder("5");
+        part.setSysUserId("1");
+        partService.insert(part);
+
+        assertEquals("Count should increase by 1", countBefore + 1, partService.getAll().size());
+    }
+
+    @Test
+    public void updateDisplayOrder_shouldPersistNewDisplayOrder() {
+        AddressPart part = new AddressPart();
+        part.setPartName("Avenue"); // short, under 20 chars
+        part.setDisplayOrder("30");
+        part.setSysUserId("1");
+        String id = partService.insert(part);
+
+        AddressPart saved = partService.get(id);
+        saved.setDisplayOrder("100");
+        partService.save(saved);
+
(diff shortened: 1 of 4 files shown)


Rohanbhalerao875 (commented):
Nice coverage across services and controller flows  especially the persistence + reload validations and menu API tests.

Quick thoughts:
- Some tests modify state (insert/update/delete) — do we rely on rollback/cleanup to avoid affecting other tests?
- A few assertions depend on counts or specific IDs — just checking if the dataset guarantees make this stable.

Overall looks solid to me.

preetihonawad:
> Nice coverage across services and controller flows especially the persistence + reload validations and menu API tests.
> 
> Quick thoughts:
> 
> * Some tests modify state (insert/update/delete) — do we rely on rollback/cleanup to avoid affecting other tests?
> * A few assertions depend on counts or specific IDs — just checking if the dataset guarantees make this stable.
> 
> Overall looks solid to me.

Thanks for the detailed review and feedback!

Regarding state isolation — good point. The tests that modify data (insert/update/delete) use the `@Before setUp()` method, which calls `executeDataSetWithStateManagement()` before each test. This resets the database to a clean state every time, so changes from one test don’t affect others. For the POST tests, I’ve also used unique `elementIds` (like "elementTest20", "elementOrder40") to avoid conflicts with existing fixture data.

For the count and ID assertions — these are based on the `menu.xml` fixture, which contains a fixed dataset. So values like the number of active menus or specific IDs remain consistent as long as the fixture doesn’t change. If the fixture is updated in the future, these assertions may need to be updated as well. I can add a comment in the test file to make this clear for future contributors if needed.

Please let me know if you’d like any changes, and thanks again for your time and feedback!

Rohanbhalerao875:
Ok menu.xml fixture is used here ...  alright then looks good to me. 
one thing though please add a short comment for menu.xml fixture being used here as the change in the dataset would affect the test cases in feature thanks.

mozzy11 (commented):
Thanks @preetihonawad — the four new MenuRestController tests cover angles the existing five don't (content-type assertions, empty-list POST handling), so we'd like to take this. One structural issue: this branch also carries the three address test files from your #3231, so the diff isn't independent. Please rebase this branch so it contains only the MenuRestControllerTest change (we're planning to merge #3231 separately), then push to refresh the stale E2E check. After that it should be a fast review.

mozzy11 (commented):
Checking back in on this one. The branch still carries the three address service test files from #3231 (last code commit here is Apr 1), so the diff is still not independent and the E2E checkpoint has gone stale. I re-verified develop: MenuRestControllerTest there still has only the original five tests, so the four additions here remain non-overlapping and worth landing. @preetihonawad please rebase this branch onto latest develop so it contains only the MenuRestControllerTest change (the address files will land via #3231) and push - that will re-trigger CI and this can then be reviewed quickly.

PR #3238, opened 2026-03-30
test: add missing edge case tests for RoleService

CodeWithJatinSh:
### Summary
The existing RoleServiceTest.java had good coverage for happy path scenarios but was missing edge case tests for null/non-existent inputs and the getReferencingRoles() method was completely untested.

###  Changes
Added **getRoleByName_shouldReturnNullForNonExistentName()** — verifies null is returned for non-existent role name
Added **getRoleById_shouldReturnNullForNonExistentId()** — verifies null is returned for non-existent role ID
Added **getReferencingRoles_shouldReturnEmptyListForRoleWithNoReferences()** — verifies method returns non-null list

###  Related
Contributing to **GSoC 2026 project: "Improve Integration Tests Coverage"**


Rohanbhalerao875 (changes requested):
Nice addition of edge case tests

Quick question: in getReferencingRoles_shouldReturnEmptyListForRoleWithNoReferences, we only assert that the result is non-null ... should we also verify that the list is actually empty to match the test intent?

I don’t see this being asserted elsewhere in the test class, so just wanted to confirm.

CodeWithJatinSh:
Great point! You're absolutely right — the test name says 'shouldReturnEmptyList' but we only assert non-null. I'll update the assertion to also include assertTrue(referencingRoles.isEmpty()) to match the test intent. Thanks for the thorough review!

Agaba-derrick:
ca you work on the CI failures please

mozzy11 (commented):
Thanks @CodeWithJatinSh — confirmed that getReferencingRoles has zero coverage on develop and the null-path cases for getRoleByName/getRoleById are untested, so these three additive tests are worthwhile. Two asks: strengthen getReferencingRoles_shouldReturnEmptyListForRoleWithNoReferences to assert the list is actually empty (not just non-null) so it can't pass vacuously, and rebase onto latest develop to trigger the full CI pipeline, which never ran on this branch. With those two tweaks this is an easy approve.

src/test/java/org/openelisglobal/role/RoleServiceTest.java
```diff
@@ -63,4 +63,25 @@ public void getRoleById_shouldReturnRoleById() {
         Role role = roleService.getRoleById("4");
         Assert.assertEquals("enter and review results.", role.getDescription());
     }
+
+    @Test
+    public void getRoleByName_shouldReturnNullForNonExistentName() {
+        Role role = roleService.getRoleByName("NonExistentRole");
+        Assert.assertNull(role);
```
mozzy11, line 70: This now fails on develop: since #3288 (2ccbb8ec), `getRoleByName` returns a stub Role with id "-1" for unknown names instead of null. Assert the stub contract instead (`assertNotNull(role)` + `assertEquals("-1", role.getId())`) or remove this test.

mozzy11 (changes requested):
develop has moved underneath this PR: #3288 (commit 2ccbb8ec, merged Apr 8) changed `RoleServiceImpl.getRoleByName` to return a stub Role (id "-1") instead of null when the name doesn't exist, so `getRoleByName_shouldReturnNullForNonExistentName` now fails against current develop. Please either assert the new stub contract or drop that test.

The other two tests remain valid — `getRoleById` does return null via `uniqueResult()`, and the `isEmpty()` assertion requested earlier is in place, thanks for that.

Two smaller items: the blank lines between the new tests carry trailing whitespace (run `mvn spotless:apply`), and the full CI pipeline has never run on this branch (only the lightweight Run check) — please rebase onto latest develop and push to trigger it.

PR #3241, opened 2026-03-30
test(organization): add integration tests for Organization service layer

preetihonawad:
## Summary
Adds new integration tests to the Organization service test classes.
All existing tests are untouched and continue to pass.

### Files Changed

**OrganizationServiceTest.java** — 2 new tests
- `delete_shouldSoftDeleteBySettingIsActiveToN` — verifies that
  `delete()` performs a soft delete by setting `isActive="N"` while
  keeping the record in the database
- `getOrganizationByCode_shouldReturnOrganizationMatchingCode` — verifies
  that an organization can be looked up by its code, used by the address
  hierarchy REST controller

**OrganizationContactServiceTest.java** — 2 new tests
- `getAll_shouldReturnAllPreloadedContacts` — verifies `getAll()` returns
  a non-null non-empty list (the existing `testDataBaseData()` had no
  `@Test` annotation so was never running)
- `getListForOrganizationId_withUnknownOrgId_shouldReturnEmptyList` —
  verifies unknown org ID returns empty list not null, guarding callers
  against NullPointerException

**OrganizationTypeServiceTest.java** — 3 new tests
- `insert_shouldPersistNewOrganizationType` — verifies a new type is
  persisted and retrievable by name, used by address hierarchy CSV loader
- `getOrganizationIdsForType_shouldReturnLinkedOrganizationIds` — verifies
  org IDs are correctly returned for a known type
- `getOrganizationIdsForType_withUnknownTypeId_shouldReturnEmptyList` —
  verifies unknown type ID returns empty list not null

## Screenshots
No UI changes — this PR adds backend integration test coverage only.

## Checklist
- [x] PR title follows conventional commit label standards (`test(organization):`)
  (`BaseWebContextSensitiveTest`, `@Before` dataset reset, JUnit 4)
- [x] Tests run: 32, Failures: 0, Errors: 0, Skipped: 0
- [x] `./mvnw spotless:apply` run before committing

## Other
- Test data loaded from existing `testdata/organization.xml` fixture
- No new dependencies added
- Soft delete behaviour confirmed — `delete()` sets `isActive="N"`,
  record remains in DB


Rohanbhalerao875 (changes requested):
Nice coverage overall

Quick question: in getOrganizationByCode_shouldReturnOrganizationMatchingCode, we return early if the code is null/empty — does that mean the test could pass without actually validating anything?

Should we instead assert or ensure the test data guarantees a non-null code?

Also, for the delete test, do we rely on rollback/cleanup to avoid affecting other tests?

Overall looks good otherwise.

preetihonawad:
> Nice coverage overall
> 
> Quick question: in getOrganizationByCode_shouldReturnOrganizationMatchingCode, we return early if the code is null/empty — does that mean the test could pass without actually validating anything?
> 
> Should we instead assert or ensure the test data guarantees a non-null code?
> 
> Also, for the delete test, do we rely on rollback/cleanup to avoid affecting other tests?
> 
> Overall looks good otherwise.

Thanks for the review — really appreciate the detailed feedback!

For the `getOrganizationByCode_shouldReturnOrganizationMatchingCode` test, you’re absolutely right. The early return when the code is null could let the test pass without actually verifying anything. I’ll update this by adding an `Assert.assertNotNull` so the test fails clearly if the fixture data is not as expected.

For the delete test, the state is reset before each test using the `@Before setUp()` method, which calls `executeDataSetWithStateManagement()`. This ensures every test runs with a clean database, so changes from one test don’t affect others.

I’ll make these updates and push them. Let me know if there’s anything else you’d like me to adjust!

pmanko (approved):

mozzy11 (commented):
Thanks @preetihonawad — this one is in great shape: purely additive tests on the existing Organization service test classes, the soft-delete assertion (isActive flips to 'N' while the row remains) and the lookup-by-code test both verify real behavior the suite didn't cover, and every CI check is green including backend and E2E. No changes requested from my side; flagging it for maintainer review as an immediate merge candidate. Nice work keeping the scope tight and the existing tests untouched.

mozzy11 (commented):
Follow-up on the review state: @Rohanbhalerao875's Mar 30 objection (silent pass when the fixture code is null) was addressed in 9356c4f — the test now fails loudly via `assertNotNull`/`assertFalse` preconditions — and the delete-test isolation question was answered (per-test dataset reset in `@Before`). With all CI checks green and @pmanko's approval, the only thing blocking merge is the stale CHANGES_REQUESTED review. @Rohanbhalerao875, could you re-review? Otherwise a maintainer can dismiss it as addressed.

Two optional tidy-ups if another push happens anyway: replace the wildcard static import in OrganizationTypeServiceTest with explicit imports, and add the missing newline at the end of OrganizationServiceTest.

PR #3242, opened 2026-03-30
fix(eqa): use schema-aware localization inserts for EQA menu migrations

HerbertYiga:
The EQA migrations (eqa-007-05, eqa-010-13) were added to the 3.3.x.x changelog after the 3.5.x.x generic localization migration had already dropped the legacy 'english' and 'french' columns from the localization table. On existing databases where 3.5.x.x already ran, these changesets fail with "column 'english' of relation 'localization' does not exist".

Replace static SQL INSERTs with PL/pgSQL blocks that detect the schema at runtime: if legacy columns exist (fresh DB), use them; otherwise insert into localization + localization_value (existing DB).


pmanko:
Thanks for chasing this! These these EQA localization changes need to be safe for existing systems, especially databases that have already run the generic localization migration.

One concern with the current approach is that it rewrites the body of older 3.3 changeSets that may already be recorded in `DATABASECHANGELOG`. That can create checksum/validation issues on environments that already applied those EQA migrations.

I think the safer direction is:
- avoid changing previously shipped changeSets unless we also handle checksum compatibility explicitly
- make the compatibility logic work for upgrade states, not just fresh-vs-new-schema installs
- in particular, consider the intermediate case where legacy columns still exist but `localization_value` is already present / partially in use

A migration that is meant to protect existing systems probably wants to either:
1. use a new follow-up changeSet that repairs/backfills the missing localization rows for upgraded databases, or
2. if these historical changeSets must be updated, account for prior checksums and make sure new rows are written in a way that stays consistent with both schemas during transition.

I’d feel better if we validate this against at least three paths:
- fresh database
- existing database already past 3.5 generic localization
- partially upgraded database where `localization_value` exists but legacy columns have not yet been dropped

The underlying issue seems real, but I want to make sure the fix doesn’t introduce a new migration problem for already-upgraded systems.

mozzy11 (commented):
Verified this against the current liquibase tree on develop:
- Both compat changesets are net-new IDs (eqa-007-05-compat, eqa-010-13-compat); no already-shipped changeset is modified, so there is no checksum breakage risk.
- Placement before eqa-007-05 / eqa-010-13 is load-bearing and correct: on upgraded DBs (english/french already dropped by 3.5.x.x 001-generic-localization-system) the compat block inserts via localization + localization_value first, so the originals' count-based preconditions MARK_RAN instead of failing on the missing column.
- Fresh installs are unaffected: 3.3.x.x runs before the 3.5.x.x drop, so the columnExists guard MARK_RANs the compat sets and the originals run as before.
- Table/column/sequence names all match the new schema (localization_value(id, localization_id, locale, value, last_updated), localization_seq / localization_value_seq).

The approach is sound. Remaining ask: the branch is ~3 months stale and the E2E gate shows a single Playwright Core shard failure from the April run - please rebase onto latest develop and push to re-run the pipeline. The failure looks unrelated to these migrations (fresh CI DBs take the legacy path where the compat sets are MARK_RAN), but we need a green gate before merging. Ready to approve once CI is green.

PR #3206, opened 2026-03-27
Feat/patient service integration tests

yusuf7861:
# Pull Requests Requirements

- [x] The PR title includes a brief description of the work done, including the Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done. (N/A - backend test-only change)
- [x] The PR title follows conventional commit label standards.
- [x] The changes confirm to the OpenElis Global x3 Styleguide and Design documentation.
- [x] The changes include tests or are validated by existing tests.
- [x] I have read and agree to the Contributing Guidelines of this project.

## Summary

This PR expands backend integration coverage for `PatientService` and stabilizes dataset-driven test execution for project/status-based patient identity lookup.

### Included across all commits in this PR
- Added/updated tests in `src/test/java/org/openelisglobal/patient/service/PatientServiceTest.java`
- Added fixture dataset `src/test/resources/testdata/patient-identity-query.xml`

### New test coverage added
- `getPatientIdentityBySampleStatusIdAndProject_shouldReturnIdentities`
- `getPatientIdentityBySampleStatusIdAndProject_shouldReturnEmptyListWhenNoMatchingData`
- `persistPatientData_shouldCreateNewPatientAndContact`
- `insertNewPatientAddressInfo_shouldInsertAddressInfo`
- `insert_shouldGenerateFhirUuidIfNull`

### Why this change
- Ensures `getPatientIdentityBySampleStatusIdAndProject(...)` has a complete, FK-valid DBUnit dataset path (`sample -> sample_projects -> project -> sample_human -> patient`) and avoids empty-result/fixture errors.
- Improves integration confidence for persistence and patient-related workflows in `PatientService`.

## Screenshots

- No UI changes.
- (Optional) Added coverage screenshot for `PatientServiceTest` as supporting evidence.
- Screenshot From 2026-03-27 23-11-08

## Related Issue

- No GitHub Issue is resolved in these changes.

## Other

Validation run used:
`mvn -q -Dtest=org.openelisglobal.patient.service.PatientServiceTest#getPatientIdentityBySampleStatusIdAndProject_shouldReturnIdentities test`

Branch diff summary:
- `src/test/java/org/openelisglobal/patient/service/PatientServiceTest.java`
- `src/test/resources/testdata/patient-identity-query.xml`
- Total: 164 insertions, 0 deletions (test-only scope).

PR #3249, opened 2026-03-30
test(patient): add integration tests for Patient service layer

preetihonawad:
## Summary
Adds new integration tests to the Patient service test classes.
All existing tests are untouched and continue to pass.

### Files Changed

**PatientServiceTest.java** — 2 new tests added
- `externalIDExists_withNonExistentId_shouldReturnFalse` — existing test
  only covered the `true` case; the `false` path guards against duplicate
  patient registration in the FHIR import workflow
- `getPatientByNationalId_withUnknownId_shouldReturnNull` — unknown
  national ID must return null not throw an exception; guards the patient
  lookup flow in lab order workflows

**PatientTypeServiceTest.java** — 3 new tests added 
- `getPatientTypes_withNonMatchingFilter_shouldReturnEmptyList` — non-null
  empty list contract; callers iterate result directly without null check
- `getTotalPatientTypeCount_shouldMatchGetAllSize` — pagination correctness;
  a count mismatch causes wrong page numbers in the UI
- `update_shouldPersistUpdatedType` — `update()` has its own duplicate
  check in the service impl; the success path was untested

**PatientContactServiceTest.java** — 3 new tests added
- `get_withValidId_shouldReturnCorrectPatientContact` — basic fetch-by-ID
  path used before every update/delete operation; completely missing from
  the test suite
- `get_withUnknownId_shouldThrowException` — documents that the base DAO
  throws `ObjectNotFoundException` for unknown IDs so callers know to use
  try-catch rather than null checks
- `getAll_afterDelete_shouldReflectReducedCount` — confirms `delete()`
  actually reduces the total count; basic data integrity check

## Screenshots
No UI changes — this PR adds backend integration test coverage only.

## Checklist
- [x] PR title follows conventional commit label standards (`test(patient):`)
  (`BaseWebContextSensitiveTest`, `@Before` dataset reset, JUnit 4)
- [x] Tests run: 87, Failures: 0, Errors: 0, Skipped: 0
- [x] `./mvnw spotless:apply` run before committing

## Other
- Test data loaded from existing `testdata/patient.xml` and
  `testdata/patient_contact.xml` fixtures
- No new dependencies added
- Real UI bug documented via test — name fields accept numeric input at
  the service layer with no validation

### DIFF
--- a/src/test/java/org/openelisglobal/patient/service/PatientTypeServiceTest.java
+++ b/src/test/java/org/openelisglobal/patient/service/PatientTypeServiceTest.java
@@ -129,4 +129,37 @@ public class PatientTypeServiceTest extends BaseWebContextSensitiveTest {
 
         Assert.assertEquals("D", savedPatientType.getType());
     }
+
+    @Test
+    public void getPatientTypes_withNonMatchingFilter_shouldReturnEmptyList() throws Exception {
+        List<PatientType> result = typeService.getPatientTypes("NonExistentType999");
+
+        Assert.assertNotNull("Result must not be null for non-matching filter", result);
+        Assert.assertTrue("Should return empty list when no types match filter", result.isEmpty());
+    }
+
+    @Test
+    public void update_shouldPersistUpdatedType() throws Exception {
+        PatientType existing = typeService.get("6");
+        Assert.assertNotNull("Precondition: type must exist", existing);
+
+        existing.setType("UPDATED_D");
+        typeService.update(existing);
+
+        PatientType reloaded = typeService.get("6");
+        Assert.assertEquals("Updated type value should be persisted", "UPDATED_D", reloaded.getType());
+    }
+
+    @Test
+    public void getTotalPatientTypeCount_shouldMatchGetAllSize() throws Exception {
+        PatientType newType = new PatientType();
+        newType.setDescription("Outpatient Type");
+        newType.setType("OP");
+        typeService.insert(newType);
+
+        int totalCount = typeService.getTotalPatientTypeCount();
+        int allCount = typeService.getAllPatientTypes().size();
+
+        Assert.assertEquals("getTotalPatientTypeCount must match getAllPatientTypes size", allCount, totalCount);
+    }
 }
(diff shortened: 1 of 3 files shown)


mozzy11 (commented):
Thanks @preetihonawad — I checked develop and the existing PatientServiceTest indeed only covers the positive paths for externalIDExists and getPatientByNationalId, so the false/null-path tests here close a real gap (and the FHIR-import duplicate-registration rationale in the description is exactly right). The E2E check on this branch has gone stale, so please rebase onto latest develop and push to re-trigger CI. One heads-up: #3206 also touches PatientServiceTest with different tests, so whichever merges second will need a trivial rebase. Green checks and this is ready for review.

src/test/java/org/openelisglobal/patient/service/PatientTypeServiceTest.java
```diff
@@ -129,4 +129,37 @@ public void getallPatientTypeByName_shouldReturnPatientType() throws Exception {
...
+
+    @Test
+    public void update_shouldPersistUpdatedType() throws Exception {
+        PatientType existing = typeService.get("6");
+        Assert.assertNotNull("Precondition: type must exist", existing);
+
+        existing.setType("UPDATED_D");
+        typeService.update(existing);
+
+        PatientType reloaded = typeService.get("6");
```
mozzy11, line 149: `get("6")` likely returns the same instance from the Hibernate first-level cache, so this assertion can pass even if `update()` never hits the DB. Consider flushing/clearing the session (or re-querying) before reloading so the test actually exercises persistence.

mozzy11 (commented):
Re-verified against current develop: the negative paths (`externalIDExists` false, `getPatientByNationalId` null) are still untested there, `BaseObjectServiceImpl.get()` does throw `org.hibernate.ObjectNotFoundException` as the new contact test expects, and the delete test follows the existing pattern in this file — the content is solid and remains purely additive. One test-strength note inline on the PatientType update test.

The blocker is unchanged since the last review: no commits since Apr 1, the Build + Test checkpoint never ran on the current head, and E2E is stuck pending. Please rebase onto latest develop and push to trigger a full CI run.

PR #3251, opened 2026-03-31
fix: encode AjaxXMLServlet validation response fields (#3250) (GsoC 26)

### LINKED ISSUE #3250: Security: XML injection risk in Ajax XML validation response serializer
## Summary

The sendData method currently concatenates field and message directly into XML:

`<formfield>` tag receives the field value as-is
 `<message>` tag receives the message value as-is

Because the values were not XML-encoded, special characters such as <, >, &, quotes used within form could be used to alter the document structure or inject unexpected content.

**How we could Fix/Patch the vunerability**

In AjaxXMLServlet.java:

- XML-encode field and message before appending into the `<formfield>` and `<message>` tags

---

## Type of Issue

- [x] Security

biga-codes:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
 This PR patches an XML injection vunerability in the validation AjaxXMLServlet response builder.
User-controlled values in formfield and message are now XML-encoded before being inserted into the XML payload, which help patch an  malformed XML and injection-style payloads while preserving the existing response contract.

The issue was that the AjaxXMLServlet was taking user controlled input like field and message values and dropping them raw into an XML response without any sanitization or encoding.

This could lead to issues where an attacker could craft input containing characters like <, >, &, or full tag fragments to break the XML structure or inject arbitrary XML content into the response.

I've wrapped those values with org.owasp.encoder.Encode.forXml() before inserting them, so special characters are safely escaped and can't alter the document structure.

## Screenshots

[Not Applicable ]

## Related Issue

Closes [#3250](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3250)

## Other

I've confirmed the XML message/formfield response structure remains unchanged

### DIFF
--- a/src/main/java/org/openelisglobal/common/servlet/validation/AjaxXMLServlet.java
+++ b/src/main/java/org/openelisglobal/common/servlet/validation/AjaxXMLServlet.java
@@ -25,6 +25,7 @@ import org.openelisglobal.common.util.StringUtil;
 import org.openelisglobal.internationalization.MessageUtil;
 import org.openelisglobal.login.dao.UserModuleService;
 import org.openelisglobal.spring.util.SpringContext;
+import org.owasp.encoder.Encode;
 import org.springframework.security.web.csrf.CsrfToken;
 import org.springframework.security.web.csrf.HttpSessionCsrfTokenRepository;
 
@@ -36,9 +37,10 @@ public class AjaxXMLServlet extends AjaxServlet {
     public void sendData(String field, String message, HttpServletRequest request, HttpServletResponse response)
             throws IOException, ServletException {
         if (!StringUtil.isNullorNill(field)) {
-            StringBuilder sb = new StringBuilder().append("<fieldmessage>").append("<formfield>").append(field)
-                    .append("</formfield>").append("<message>").append(message).append("</message>")
-                    .append("</fieldmessage>");
+            StringBuilder sb = new StringBuilder().append("<fieldmessage>").append("<formfield>")
+                .append(Encode.forXmlContent(String.valueOf(field))).append("</formfield>")
+                .append("<message>").append(Encode.forXmlContent(String.valueOf(message)))
+                .append("</message>").append("</fieldmessage>");
             if ("true".equals(request.getParameter("asJSON"))) {
                 response.setContentType("application/json");
                 response.setHeader("Cache-Control", "no-cache");
--- a/src/main/java/org/openelisglobal/common/servlet/validation/AjaxXMLServlet.java
+++ b/src/main/java/org/openelisglobal/common/servlet/validation/AjaxXMLServlet.java
@@ -38,7 +38,7 @@ public class AjaxXMLServlet extends AjaxServlet {
             throws IOException, ServletException {
         if (!StringUtil.isNullorNill(field)) {
             StringBuilder sb = new StringBuilder().append("<fieldmessage>").append("<formfield>")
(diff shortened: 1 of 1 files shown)


mozzy11 (commented):
Thanks @biga-codes — the unencoded XML response in AjaxXMLServlet is still present on develop and the OWASP encoder is already a project dependency, so this fix is squarely applicable. One gap: the diff encodes message but leaves field unencoded (String.valueOf doesn't sanitize), so the formfield element is still injectable — please apply Encode.forXmlContent there too. Also note the asJSON branch reuses this same string, so double-check the encoded output is still valid for the JSON conversion path. Finally, rebase onto latest develop so CI actually runs; no checks ever executed on this branch. Please also fold the identical fix from your #3253 (AjaxQueryXMLServlet) into this PR so both servlets are fixed and tested together.

src/main/java/org/openelisglobal/common/servlet/validation/AjaxXMLServlet.java
```diff
@@ -36,9 +37,10 @@ public class AjaxXMLServlet extends AjaxServlet {
     public void sendData(String field, String message, HttpServletRequest request, HttpServletResponse response)
             throws IOException, ServletException {
         if (!StringUtil.isNullorNill(field)) {
-            StringBuilder sb = new StringBuilder().append("<fieldmessage>").append("<formfield>").append(field)
-                    .append("</formfield>").append("<message>").append(message).append("</message>")
-                    .append("</fieldmessage>");
+            StringBuilder sb = new StringBuilder().append("<fieldmessage>").append("<formfield>")
+                .append(String.valueOf(field)).append("</formfield>")
```
mozzy11, line 41: `String.valueOf(field)` is a no-op for sanitization — the `formfield` element is still injectable. Apply `Encode.forXmlContent(field)` here too (and de-duplicate the call-site encoding; see review body).

src/main/java/org/openelisglobal/common/servlet/validation/AjaxXMLServlet.java
```diff
@@ -36,9 +37,10 @@ public class AjaxXMLServlet extends AjaxServlet {
...
+                .append("<message>").append(Encode.forXmlContent(String.valueOf(message)))
```
mozzy11, line 42: Several callers already pass a pre-encoded message (e.g. `QuickEntrySampleTypeValidationProvider:54` passes `Encode.forXmlContent(result)`), so this double-encodes their output. Centralize encoding here and remove it at the call sites.

mozzy11 (changes requested):
Confirmed the unencoded response is still present on develop, so this fix is still needed — but two defects block it:

1. `field` is still written unencoded (`String.valueOf` doesn't sanitize), so the `formfield` element remains injectable. This was flagged in the previous review and is still unaddressed.
2. Double-encoding regression: some callers already encode the message at the call site — e.g. `QuickEntrySampleTypeValidationProvider` line 54 passes `Encode.forXmlContent(result)` — so encoding again inside `sendData` corrupts their output (`&` renders as `&amp;`). Please pick one layer: encode both args centrally in `sendData` and remove the call-site `Encode.forXmlContent(...)` wrappers after auditing all `sendData` callers.

Also: the new indentation doesn't match the project formatter (run `mvn spotless:apply`), and CI has never run on this branch — please rebase onto latest develop and push.

biga-codes:
@mozzy11 Thank you so much for the review,  will get to work on these changes and provide updates as soon as possible

Issue #3252, opened 2026-03-31
Security: XML output encoding vulnerability in AjaxQueryXMLServlet sendData


biga-codes:
**Summary**
The sendData method in AjaxQueryXMLServlet concatenates field and message directly into XML without encoding:

1) formfield tag receives the field value as is
2) message tag receives the message value as is

Because these values aren't XML encoded special characters like <, >, &, and quotes can alter the XML structure or inject unexpected content into the response.

**Affected Component**
AjaxQueryXMLServlet.java
Class: AjaxQueryXMLServlet
Method: sendData(String field, String message, HttpServletRequest request, HttpServletResponse response)

The current AjaxQueryXMLServlet.java is prone to vulnerabilities in the following way:

The message parameter in AjaxQueryXMLServlet.sendData() was directly concatenated into XML output without encoding 

**Why this is an issue**

An attacker could craft a malicious message containing XML special characters:

```
Input message: </message><injected>attack</injected><message>
```

Without encoding the XML output becomes:

```xml
<fieldmessage>
  <formfield>someField</formfield>
  <message></message><injected>attack</injected><message></message>
</fieldmessage>
```

The attacker successfully breaks out of the message tag and injects arbitrary XML.

With encoding it becomes safe:

```xml
<fieldmessage>
  <formfield>someField</formfield>
  <message>&lt;/message&gt;&lt;injected&gt;attack&lt;/injected&gt;&lt;message&gt;</message>
</fieldmessage>
```

The patch ensures special characters are escaped so the injection fails. JavaScript consumers will still parse correctly because .nodeValue automatically decodes entities

**Risk Level**
The risk level is high and exploitable via any QueryProvider that passes user controlled data as the message parameter, it could lead to XML structure corruption and potential information disclosure so we should avoid those circumstances

**How we could fix or patch the vulnerability**
In AjaxQueryXMLServlet.java:

1) XML encode field and message before appending into the formfield and message tags
2) Keep endpoint contract and response structure unchanged

**Type of Issue**
[ X ] Security

PR #3253, opened 2026-03-31
fix: encode message in AjaxQueryXMLServlet response (#3252)

biga-codes:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

Summary
This PR fixes an XML injection vulnerability in AjaxQueryXMLServlet sendData where message content was inserted into XML without output encoding and so it includes a security fix for the vulnerability in AjaxQueryXMLServlet.sendData(). 

The message parameter is now XML encoded using org.owasp.encoder.Encode.forXmlContent() to prevent XML injection attacks.

The patch does the following-

Encodes the message field (free form user/system text, higher injection risk).

Does not encode the field parameter (since it's already pre encoded by some QueryProvider callers - which should avoid any instances of double escaping

Without encoding the XML output could lead to this: 

An attacker crafting a malicious message value like:
</message><injected>attack</injected><message>
 without encoding would break out of the <message> tag and inject arbitrary XML into the response structure, 

allowing XML structure corruption and response tampering this would mean that consumers parsing the responseXML would receive unexpected nodes and could lead to information disclosure depending on what gets injected and how the client handles it

With encoding it becomes safe:

xml<fieldmessage>
  <formfield>someField</formfield>
  <message>&lt;/message&gt;&lt;injected&gt;attack&lt;/injected&gt;&lt;message&gt;</message>
</fieldmessage>

any special characters in input will be escaped so the injection fails.

This prevents: 

Screenshots
Not applicable : backend security fix

Related Issue
Refs [#3252](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3252)

## Other
I noticed formfield encoding is currently mixed across callers. Some providers already pass pre encoded values while others pass raw values. If we Centralize the 'formfield' encoding in sendData right now this could cause double encoding issues, for example entity text becoming double escaped. For this reason, I've ensured this PR encodes only message as an fix as of now. 

A future cleanup should be able to standardize the encoding across providers and servlet responses globally

### DIFF
--- a/src/main/java/org/openelisglobal/common/servlet/query/AjaxQueryXMLServlet.java
+++ b/src/main/java/org/openelisglobal/common/servlet/query/AjaxQueryXMLServlet.java
@@ -5,6 +5,7 @@ import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import org.json.XML;
+import org.owasp.encoder.Encode;
 import org.openelisglobal.common.exception.LIMSRuntimeException;
 import org.openelisglobal.common.provider.query.BaseQueryProvider;
 import org.openelisglobal.common.provider.query.QueryProviderFactory;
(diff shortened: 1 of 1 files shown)


mozzy11:
Thanks @biga-codes — this fix is valid (AjaxQueryXMLServlet still has the unencoded response on develop), but it's a one-line twin of your #3251 fixing the same vulnerability class in a sibling servlet. To keep the queue manageable I'm closing this one: please fold this change into #3251 so both servlets are fixed, reviewed, and CI-tested together in a single PR. Same note applies here — the field value should be encoded as well, not just message.

PR #3291, opened 2026-04-03
fix: use Objects.equals() for category string comparison in updateReport

Junaidarmaan:
## Problem
`updateReport()` was using `!=` to compare two strings, which compares object references not values. The menu reassignment logic would silently fail even when the report category actually changed.

## Fix
Replaced `!=` with `Objects.equals()` which compares string values and handles null on either side.


mozzy11 (commented):
Thanks @Junaidarmaan — confirmed the bug on current develop: ReportServiceImpl line 63 compares two Strings with !=, so the category-change/menu-reassignment branch only fires when the JVM happens not to intern the values. Objects.equals (negated) is the right minimal fix and also handles nulls. Two small asks: rebase onto latest develop so the full CI pipeline runs (only the trigger job executed on this branch), and if feasible add a focused test asserting that updateReport reassigns the menu when the category changes — that would lock the fix in. Happy to review as soon as CI is green.

Junaidarmaan:
Hi @mozzy11, thanks for the feedback.

I've rebased the branch onto the latest develop and added focused regression tests covering both the category-change and unchanged-category scenarios for updateReport.

The branch has been force-pushed and is ready for review. It looks like the remaining CI workflows are currently awaiting maintainer approval.

If there are any additional changes, improvements, or tests you'd like me to add, I'd be happy to work on them. Thanks again for taking the time to review this PR!

src/test/java/org/openelisglobal/reportconfiguration/service/ReportServiceImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+        formReport.setIsVisible(true);
+        form.setCurrentReport(formReport);
+
+        form.setIdOrder(REPORT_ID);
+
+        ReportCategory reportCategory = new ReportCategory(NEW_CATEGORY_ID, "New Category",
+                NEW_CATEGORY_MENU_ELEMENT_ID, "1", null);
+        form.setReportCategoryList(Arrays.asList(reportCategory));
+
+        return form;
```
mozzy11, line 89: With interned literals, `OLD_CATEGORY_ID != OLD_CATEGORY_ID` is already false, so this test passes even with the old buggy `!=` code. Use a distinct String instance to expose the reference-comparison bug, e.g. `buildForm(new String(OLD_CATEGORY_ID))`: under `!=` the branch then wrongly fires, the `reportCategory` lookup returns null, the NPE is swallowed and `updateReport` returns false — failing this test exactly as a regression test should.

mozzy11 (changes requested):
Thanks for the rebase and the regression tests. The production fix is correct and still needed — develop's `ReportServiceImpl` line 63 still compares the two Strings with `!=`.

However, the new tests don't actually guard the fix: both category IDs are compile-time String literals, which the JVM interns, so `categoryId != previousCategory` evaluates identically to `!Objects.equals(categoryId, previousCategory)` in both scenarios. I checked both tests against the pre-fix code path — they pass with the old `!=` comparison too, so they'd never catch a regression. Details inline on how to make the unchanged-category test discriminate.

Also: the test file is missing a trailing newline (`mvn spotless:apply` will fix), and prefer `assertSame`/`assertNull` over `assertTrue(x == y)` at lines 81-82 and 97. With the inversion fixed, this is ready — we'll approve the full CI run on the next push.

Junaidarmaan:
Hi @mozzy11, thanks for the detailed review.

I've addressed the requested changes:
- Updated the unchanged-category regression test to use `new String(OLD_CATEGORY_ID)` so it exercises the reference-comparison bug and would fail with the old `!=` implementation.
- Replaced the assertions with `assertSame` and `assertNull` as suggested.
- Cleaned up the test file (Spotless/checks pass locally) and verified the relevant tests pass.

The branch has been updated. Thanks again for the helpful feedback!

PR #3324, opened 2026-04-05
feat(fhir): add DiagnosticReport read endpoint

vsvishalsharma:


pmanko (approved):

pmanko (dismissed):
Thanks for this — the `@Read` implementation is clean and follows the existing provider pattern correctly: Analysis lookup from OE DB by fhirUuid, on-the-fly transform via `FhirTransformService`.

A few notes for follow-up (not blocking merge):

**`@Search`**: When this gets added, it should delegate to the co-habitant HAPI FHIR store via `FhirUtil.forwardSearchToFhirStore(request)`, matching the existing providers. At minimum: `subject`, `status`, `date` parameters.

**Test coverage gaps**: Two cases worth adding in a follow-up:
- Duplicate-UUID path (`matches.size() > 1` → 500)
- `FhirTransformationException` scenario

**Architectural context** — this PR highlights a broader issue we need to address:

OpenELIS currently has two overlapping FHIR surfaces: the OE providers (this code) and the co-habitant HAPI FHIR store (`external-fhir-api`). They serve the same resource types but from different sources — `@Read` pulls from the OE database and transforms on-the-fly, while `@Search` forwards to the HAPI store. The sync between them is fire-and-forget (`FhirProviderUtils.syncToFhirStore` catches and logs all failures), so the two can silently diverge.

The intended direction:
1. **The co-habitant HAPI store should be the external-facing FHIR endpoint for OpenELIS.** It already has the naming (`external-fhir-api`), the search indexing, and the conformance infrastructure.
2. **The OE FHIR providers are a transitional facade** — useful for domain-validated writes and for bootstrapping new resource types (like this PR), but not the long-term external surface.
3. **Before we can rely on HAPI for external reads, the sync must be hardened** — retry with backoff, a reconciliation job to detect drift, and monitoring for sync failures. Today's fire-and-forget is a known gap.
4. **Write routing (how external systems submit orders/results with domain validation) is a separate design question** that can be addressed once the read/search path is solid.

I've updated #3328 to track this broader scope. This PR is a good incremental step — merging as-is.

pmanko:
**Analyzer bridge connection**: The analyzer bridge (`POST /analyzer/fhir`) receives FHIR bundles from instruments that contain DiagnosticReport, Observation, and Specimen resources. The DiagnosticReport in the incoming bundle is correctly discarded — it represents raw analyzer output, not a validated lab report. The validated DiagnosticReport is created later during result finalization, after tech/bio review.

But the **ingress path itself** is relevant to the architecture direction above. The analyzer bridge is a hardwired FHIR consumer that manually extracts Observations/Specimens and maps them to staging rows, completely bypassing the HAPI store. If we're standardizing around HAPI as the external FHIR surface, the ingress path should be part of that conversation — not just egress (how external systems read results) but also how external systems submit data into OE through a standardized FHIR interface.

Both directions tracked in #3328.

vsvishalsharma:
Addressed review follow-ups in this PR:

- Added duplicate UUID guard in DiagnosticReport @Read (`matches.size() > 1` -> 500)
- Added explicit `FhirTransformationException` handling in provider
- Added test coverage for:
  - duplicate UUID path -> 500 (`DiagnosticReportProviderTest`)
  - transform exception path -> 500 (`DiagnosticReportProviderTest`)
  - invalid UUID -> 400 (`DiagnosticReportFacadeTest`)

pmanko (approved):

PR #3342, opened 2026-04-06
fix(equipment): use native date inputs in usage log

dawit-Tegegnwork:
## Summary
- replace the equipment usage date field with a native date input
- replace the approval date field with a native date input
- avoid the repeated character/date-entry issue in the usage log table

## Testing
- tested locally

### DIFF
--- a/frontend/src/components/equipmentUsage/EquipmentUsage.css
+++ b/frontend/src/components/equipmentUsage/EquipmentUsage.css
@@ -281,6 +281,27 @@
 }
 
 /* Date Picker Input in Table */
+.tableDatePicker {
+  width: 100%;
+  min-width: 140px;
+}
+
+.tableDatePicker .cds--date-picker,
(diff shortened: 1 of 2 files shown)


dawit-Tegegnwork:
Screenshot from 2026-04-06 09-03-37-1
This screenshot shows the repeated date-entry issue in the Equipment Usage Log before the fix.

samuelmale:
@dawit-Tegegnwork use: https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/frontend/src/components/common/CustomDatePicker.js. Here is an example: https://github.com/DIGI-UW/OpenELIS-Global-2/commit/abd135aa02103a6af397dabf3a6e38411566b40a

dawit-Tegegnwork:
Updated it to use CustomDatePicker and pushed the change.

PR #3344, opened 2026-04-06
inventory: add next maintenance date for equipment

dawit-Tegegnwork:
Adds an optional Next Maintenance Date field for equipment inventory items and shows it in the equipment dashboard view.

This also updates the inventory UI so users see Equipment instead of Analyzer Cartridge, while keeping the backend item type unchanged.

Tested locally.

### DIFF
--- a/frontend/src/components/inventory/InventoryCatalog.jsx
+++ b/frontend/src/components/inventory/InventoryCatalog.jsx
@@ -166,7 +166,7 @@ const InventoryCatalog = () => {
     const labels = {
       REAGENT: "Reagent",
       RDT: "RDT (Rapid Diagnostic Test)",
-      CARTRIDGE: "Analyzer Cartridge",
+      CARTRIDGE: "Equipment",
       HIV_KIT: "HIV Test Kit",
       SYPHILIS_KIT: "Syphilis Test Kit",
       ENZYME: "Enzyme",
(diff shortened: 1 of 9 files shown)


frontend/src/components/inventory/InventoryCatalog.jsx
```diff
@@ -166,7 +166,7 @@ const InventoryCatalog = () => {
     const labels = {
       REAGENT: "Reagent",
       RDT: "RDT (Rapid Diagnostic Test)",
-      CARTRIDGE: "Analyzer Cartridge",
+      CARTRIDGE: "Equipment",
```
samuelmale, line 169: Was this intended?

frontend/src/components/inventory/InventoryItemForm.jsx
```diff
@@ -790,6 +798,18 @@ const InventoryItemForm = ({ open, onClose, onSave, item = null }) => {
                   }
                 />
               </DatePicker>
+
+              <DatePicker datePickerType="single">
+                <DatePickerInput
+                  id="nextMaintenanceDate"
+                  placeholder="mm/dd/yyyy"
+                  labelText="Next Maintenance Date"
```
samuelmale: Can we localize the label?

frontend/src/components/inventory/InventoryCatalog.jsx
dawit-Tegegnwork: Yes, that was intentional. The backend item type remains CARTRIDGE, but for this workflow we want the user-facing label to be Equipment because that is what users are actually managing.

frontend/src/components/inventory/InventoryItemForm.jsx
dawit-Tegegnwork: Localized the Next Maintenance Date label and pushed the update.

samuelmale (approved):

PR #3356, opened 2026-04-07
(feat) Introduce a Daemon/System user and a centralised UserContext

samuelmale:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Previously, a lot of code that ran outside a user request (scheduled jobs, system tasks, etc.) would just hardcode "1" (the admin user ID) for audit trails. That's brittle, misleading, and a pain to track down.

This PR fixes that by:

- **Creating a dedicated "daemon" system user in the database** — a real user row that can't log in but represents "the system did this" in audit logs.
- **Adding a central place to get the current user (UserContextHolder)** — whether it's a human logged in via the browser or the daemon running a background job, there's one consistent way to ask "who's doing this?"
- **Auto-stamping audit fields in the service layer** — instead of every service method manually calling setSysUserId(...), the base auditable service (AuditableBaseObjectServiceImpl) fills it in automatically from the current security context on every insert/update/delete. If there's no context at all, it fails loudly instead of silently falling back to admin.
- **Wrapping scheduled tasks in a daemon security context** — so background jobs automatically run as the daemon user. No more manual setup per job.

## Screenshots

## Related Issue

## Other


pmanko:
## Review: UserContext & Daemon System User

Nice work @samuelmale — this is a well-structured approach to a long-standing pain point. Some observations and context from a codebase-wide audit:

### What this PR solves well

1. **Eliminates ~20 hardcoded `setSysUserId("1")` calls** across background services, scheduled jobs, plugin registration, and config handlers — replacing them with `userContextHolder.requireSysUserId()`.
2. **Centralizes user resolution** in `UserContextHolder` with a `ConcurrentHashMap` cache — no more repeated DB lookups per request.
3. **Daemon context for background threads** — `DaemonAuthenticationToken` + `DaemonContextExecutor` + `UserContextPropagatingTaskDecorator` gives scheduled tasks and `@Async` methods a proper identity instead of silent fallback to user ID 1.
4. **AOP safety net** — `AuditContextAdvice` catches mutations with no user context before they hit the DB, preventing the unhandled `NumberFormatException` in `LIMSStringNumberUserType.nullSafeSet()` that would otherwise crash the transaction.
5. **Good test coverage** — `UserContextHolderTest`, `DaemonContextExecutorTest`, `DaemonContextAssertTest`, `AuditContextAdviceTest`, plus a `@WithDaemonUser` annotation for test convenience.

### Context: the broader sysUserId landscape

We did a full codebase audit and filed [OGC-549](https://uwdigi.atlassian.net/browse/OGC-549) to track the String→Long type refactor for `sysUserId`:

| Metric | Count |
|--------|-------|
| Files calling `setSysUserId()` | **283** |
| Files calling `getSysUserId()` | **223** |
| String→int/Long parsing sites | **27** |
| Parsing sites with `NumberFormatException` handling | **2 of 27** |

The `sysUserId` field is `String` in `BaseObject` but `numeric(10,0)` in the database. The critical bottleneck is `LIMSStringNumberUserType.nullSafeSet()` which does `Integer.parseInt()` with zero error handling — every audit trail insert flows through it. There's also inconsistent parsing (some `Integer.parseInt`, some `Long.parseLong`).

**This PR and OGC-549 are complementary:**

| Concern | This PR | OGC-549 |
|---------|---------|---------|
| Eliminate hardcoded "1" | ✅ | ✅ (side effect) |
| Centralize user resolution | ✅ | ❌ |
| Daemon context for background jobs | ✅ | ❌ |
| AOP safety net for mutations | ✅ | ❌ |
| Fix String→Long type mismatch | ❌ | ✅ |
| Remove `LIMSStringNumberUserType` | ❌ | ✅ |
| Eliminate 27 parsing sites | ❌ | ✅ |

OGC-549 should land **after** this PR merges to avoid conflicts, since both touch many of the same files. Once `UserContextHolder` is in place, the type refactor becomes cleaner — `getCurrentSysUserId()` can return `Long` directly instead of `String`.

### A few questions / minor items

1. **`DaemonUserConfig` fallback to ID `"1"`** — The fallback `SystemUser` with `id="1"` and `loginName="admin"` is a reasonable safety net for unmigrated environments, but worth a log-level bump? Currently it's `logWarn` which seems right, just want to confirm this is expected to be transient (i.e., environments should run the Liquibase migration and never hit the fallback).

2. **`UserContextHolder.getCurrentSysUserId()` returns `String`** — Since this is the new canonical way to get the user ID, would it make sense to return `Long` here from the start? The underlying `SystemUser.getId()` returns `String`, but the DB column is `numeric(10,0)` and every downstream consumer parses it to a number anyway. This would set up OGC-549 nicely. (Not a blocker — can be done in the follow-up.)

3. **Cache invalidation** — `principalToUserCache` is a `ConcurrentHashMap` with no eviction. For a typical deployment with a small number of users this is fine, but if a system user record is updated (e.g., deactivated), the cache would serve stale data until app restart. Worth documenting or adding a `clearCache()` method for admin operations?

4. **`AuditContextAdvice` + `AuditableBaseObjectServiceImpl.fillSysUserIdIfMissing()`** — Both the AOP advice and the service method now fill `sysUserId` if missing. The belt-and-suspenders approach is safe, but is the intent that the AOP advice is the primary mechanism and `fillSysUserIdIfMissing()` is the fallback? Or vice versa? A comment clarifying the layering would help future contributors.

Overall this is a solid foundation that the rest of the audit infrastructure can build on. 👍

[OGC-549]: https://uwdigi.atlassian.net/browse/OGC-549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

samuelmale:
@pmanko @mozzy11 any chance we can have this reviewed?

samuelmale:
@mozzy11 any chance we can have this reviewed?

src/main/java/org/openelisglobal/scheduler/SchedulerConfig.java
```diff
@@ -71,7 +77,9 @@ public long getResultsResendTimeMillis() {
 
     @Bean(destroyMethod = "shutdown")
     public Executor taskExecutor() {
-        return Executors.newScheduledThreadPool(10);
+        ScheduledExecutorService raw = Executors.newScheduledThreadPool(10);
```
mozzy11, line 80: **Blocking — startup crash on real deployments.** The daemonized `taskExecutor()` correctly covers all 13 `@Scheduled` methods (verified `configureTasks` → `taskRegistrar.setScheduler(taskExecutor())`), but `configureTasks` itself still runs on the **bootstrap thread with no SecurityContext**, and its catch-up branch calls:

```java
IImmediateJobRunner runner = SpringContext.getBean(IImmediateJobRunner.class);
runner.runNow(reloadableScheduler, jobName);   // @Async
```

`ImmediateJobRunner.runNow` is `@Async`, so the new `UserContextPropagatingTaskDecorator` runs **synchronously on the submitting thread** and throws `IllegalStateException("Cannot propagate SecurityContext…")`. The surrounding catches are `NumberFormatException` / `SchedulerException | ParseException` — neither catches it, so it propagates out of context refresh and **the app fails to boot**. Trigger: any active `cron_scheduler` row (`sendSiteIndicators`, `sendMalariaSurviellanceReport`) whose cron time is earlier than boot time — i.e., exactly the aggregate/malaria-reporting installs, on any restart after the daily send time. CI can't see this: `AppTestConfig` excludes both `SchedulerConfig` and `org.openelisglobal.config.*` from the test profile.

Fix: wrap the catch-up dispatch (`daemonContextExecutor.executeAsDaemon(() -> runner.runNow(...))`) or run all of `configureTasks` under the daemon context.

src/main/java/org/openelisglobal/analyzer/service/PluginRegistryService.java
```diff
@@ -94,6 +98,10 @@ public class PluginRegistryService {
     @PostConstruct
     @Transactional
     public void registerLoadedPlugins() {
+        daemonContextExecutor.executeAsDaemon(() -> doRegisterLoadedPlugins());
```
mozzy11, line 101: **Blocking — two sibling startup writers were not wrapped**, and both now fail silently (their throws are swallowed by broad catches):

1. **`PluginMenuService.onApplicationReady`** (`@EventListener(ContextRefreshedEvent)`) → `registerAnalyzerMenuAndPermission` → `PluginPermissionService.getOrCreateSystemModule(...)` / `bindRoleToModule(...)` → `roleService.insert` / `moduleService.insert` / `roleModuleService.insert` — all `AuditableBaseObjectServiceImpl` subclasses, and this PR removed the `setSysUserId("1")` calls in `PluginPermissionService`. On the context-refresh thread there's no SecurityContext → `fillSysUserIdIfMissing` throws → swallowed by `catch (Exception)` in `initializeAnalyzerMenus()` → on a fresh DB or when a new analyzer is added, **analyzer result menus and permissions are silently never created** (and the loop aborts at the first failure).

2. **`PluginLoader.load()`** (`@PostConstruct`) → legacy `plugin.connect()` → `PluginAnalyzerService.findOrCreateAnalyzerForType` (this PR removed its `setSysUserId("1")`) → `analyzerService.insert(...)` (Auditable) plus `persistTestMappings` (both `setSysUserId("1")` removed) → throws with no context → swallowed → **legacy analyzer plugins (Sysmex, Cobas, …) never get their Analyzer row or test mappings**. Worse: if a deployed plugin's `connect()` reaches `bindRoleToModule` directly, `LIMSRuntimeException` isn't caught by `loadFromXML` (`catch (IOException | DocumentException | LIMSException)` — `LIMSRuntimeException` extends `RuntimeException`, not `LIMSException`) and **escapes the @PostConstruct, aborting startup**.

Both need the same treatment you gave `registerLoadedPlugins()` here: wrap in `daemonContextExecutor.executeAsDaemon(...)`. Recommend a follow-up sweep for any other `ContextRefreshedEvent`/`@PostConstruct` writer (our sweep found these two as the only remaining ones).

src/main/java/org/openelisglobal/config/UserContextPropagatingTaskDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+ * (including daemon context) from their caller.
+ */
+public class UserContextPropagatingTaskDecorator implements TaskDecorator {
+
+    @Override
+    public Runnable decorate(Runnable runnable) {
+        // Guard: reject @Async submissions with no SecurityContext — this is a bug,
+        // not something to silently propagate.
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        if (auth == null || !auth.isAuthenticated()) {
```
mozzy11, line 20: Design note on throw-on-unauthenticated: making `decorate()` throw on the **submitting** thread turns every unguarded `@Async` submission from a context-less thread into an exception at the call site — that's how the `configureTasks` startup crash arises. The guard's intent (surface bugs instead of propagating nothing) is good, but consider: (a) after wrapping the known offenders, keep the throw, or (b) fall back to the daemon context with a loud `logError` for one release, then tighten. Either way, please document the contract on `AsyncConfig` so future `@Async` additions know they must run under an authenticated or daemon context.

src/main/java/org/openelisglobal/common/util/UserContextHolder.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Returns the daemon {@link SystemUser}.
+     */
+    public SystemUser getDaemonSysUser() {
+        return daemonSystemUser;
+    }
+
+    private SystemUser resolveSystemUser(String principalName) {
```
mozzy11, line 100: Performance note (non-blocking, but worth a follow-up): `resolveSystemUser` → `SystemUserService.getDataForLoginUser` is an **uncached HQL query** (`from SystemUser where login_Name = :name`, no `@Cacheable`, bypasses the L2 entity cache), and `fillSysUserIdIfMissing` invokes it on **every insert/update/delete whose entity arrives unstamped** in an HTTP context. Bulk operations (`insertAll`, CSV config reloads, result batches) now pay one extra DB round-trip per row. The earlier revision's `principalToUserCache` was removed (which resolved @pmanko's staleness question) — a small per-request cache (e.g. request-scoped memo or a short-TTL cache keyed by principal) would recover the cost without the staleness problem.

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -234,6 +258,28 @@ protected void executeDataSetWithStateManagement(String datasetFileName) throws
...
+     * Ensure a {@code system_user} row with {@code login_name='admin'} exists so
+     * the principal set by {@link #setDefaultTestAuthentication()} resolves via
+     * {@link org.openelisglobal.common.util.UserContextHolder}. Idempotent: checks
+     * by {@code login_name} (not by PK) to avoid creating a second admin when
+     * another fixture already populated the row with a different id.
+     *
+     * <p>
+     * Tests whose fixtures truncate {@code system_user} and load their own users
+     * (e.g. {@code testUser}, {@code alice}) wipe this row; those tests should set
+     * up a daemon SecurityContext in their own {@code @Before} to avoid the admin
```
mozzy11, line 271: Robustness: this helper is keyed on `login_name='admin'` while the existing `ensureAuditSystemUser()` (invoked after every fixture load) is keyed on `id=1`. For the ~19 fixtures that seed `system_user` with id=1 occupied by a non-admin login plus an `admin` at some other id (e.g. `analysis-qa-event-action.xml`: `501:admin`), the two helpers jointly produce **two `login_name='admin'` rows**, and `getDataForLoginUser` (`query.list()`, no ORDER BY) then resolves nondeterministically. Harmless today (audit service is mocked in tests), but one `uniqueResult`-style refactor away from flaky `NonUniqueResultException`s. Suggest keying this insert on `login_name='admin' OR id=1`, or reconciling the two helpers into one.

src/main/resources/liquibase/3.5.x.x/base.xml
```diff
@@ -29,6 +29,9 @@
   <!-- LocationPicker refactor (Phase 5c): rename /Storage/samples → /Storage/sample-items -->
   <include relativeToChangelogFile="true" file="006-storage-sidenav-rename.xml"/>
 
+  <!-- Daemon system user for background/scheduled operations -->
+  <include relativeToChangelogFile="true" file="003-daemon-system-user.xml"/>
```
mozzy11, line 33: Minor: the include is inserted between `006-*` and `004-*`, and `3.5.x.x` already has three other `003-*` files — both are established (if unlovely) precedent here and Liquibase keys changesets on id/author/path, so this is cosmetic. Verified the substantive bits: `${now}` resolves (property defined in earlier-included changelogs; `002-eqa-enabled-config.xml` uses the same pattern), unqualified `valueSequenceNext="system_user_seq"` matches `014-sample-collection-workflow-menus.xml` precedent and resolves via the `clinlims` search_path, and `@DependsOn("liquibase")` matches the actual `@Bean("liquibase")` in `LiquibaseConfig`. One real gap: **no `<rollback>`** — an explicit `DELETE ... WHERE login_name='daemon'` rollback would keep tag-based rollbacks working across this file.

src/main/java/org/openelisglobal/dataexchange/resultreporting/ResultReportingTransfer.java
```diff
@@ -135,10 +136,16 @@ private void persistMessage(String msg, boolean success) {
             messageOutService.insert(messageOut);
         }
 
+        private String resolveSysUserId() {
+            UserContextHolder uch = SpringContext.getBean(UserContextHolder.class);
+            String id = uch.getCurrentSysUserId();
+            return id != null ? id : uch.getDaemonSysUserId();
+        }
```
mozzy11, line 143: Consistency nit: this handler (and `MalariaReportingTransfer`) uses the lenient `getCurrentSysUserId() != null ? … : getDaemonSysUserId()` fallback, while the job classes use strict `requireSysUserId()`. The lenient form silently attributes to the daemon in contexts the strict form would flag as bugs. Worth picking one policy — strict inside daemon-wrapped code (where context is guaranteed), lenient only where a genuinely mixed HTTP/background call graph exists (which is true here for `sendHL7Report`) — and saying so in a comment, so future readers don't copy the wrong one.

mozzy11 (dismissed):
## Review: deep regression audit (requested by @samuelmale — sorry for the wait!)

I did a full-depth pass: every hunk of the 97-file diff, all prior review feedback (Copilot ×6, @pmanko's audit comment), and a codebase-wide sweep of every non-HTTP execution path that writes through the service layer, verified against the current develop base and the PR head.

### Was this change necessary? Yes.

The problem is real and this is the right shape of solution: ~20+ call sites hardcoded `sysUserId="1"`, which makes audit trails actively misleading — "admin did X" is indistinguishable from "a cron job did X", which matters in a LIMS. A real, non-loginable `daemon` user + one `UserContextHolder` + auto-stamping with fail-loud semantics is a solid foundation, and it sets up OGC-549 cleanly. The concern is not the idea — it's the blast radius of "fail loudly" on paths that weren't wrapped, and that's where the three blocking findings below live.

### Verified safe (so nobody re-litigates these)

- **Compiles.** Copilot's "missing import" comment is a false positive — `ControllerUtills` and `UserContextHolder` share `org.openelisglobal.common.util` (Shared Build passes). The no-arg `getSysUserId()` resolves via `BaseRestController → ControllerUtills` static inheritance.
- **All 13 `@Scheduled` methods are correctly daemonized** — `configureTasks` registers `taskExecutor()`, now a `DelegatingSecurityContextScheduledExecutorService` with the daemon context. Both Quartz jobs on develop are wrapped. `ReportTransmission` handlers run synchronously on the calling (daemon) thread for all job callers; the one async path (`sendHL7Report`) has an explicit daemon fallback.
- **Analyzer ASTM/HL7 ingestion is safe** — enters via Basic-auth REST (`/analyzer/astm`, `/analyzer/hl7`, `/importAnalyzer`) with explicit sysUserId; no raw-socket ingestion exists in the webapp.
- **`IndicatorAggregationReportingServlet`** — behind the security filter chain (`anyRequest().authenticated()`, not in OPEN_PAGES), so the removed `setSysUserId("1")` is unreachable without a principal.
- **Login flows** — failure handler performs zero DB writes on develop; success handler writes only a non-auditable `Notification`; the change-password-before-login fix in this PR is correct.
- **Liquibase specifics** — `${now}`, unqualified `valueSequenceNext`, and `@DependsOn("liquibase")` all check out against repo precedent (details inline).
- **Backend Build+Test is green (4,700+ tests)** and our test-blast-radius analysis agrees: the ~15 patched test classes are exactly the set that needed patching; projected new failures ≈ 0.

### Blocking (3) — all invisible to CI by construction

`AppTestConfig` excludes `SchedulerConfig` and `org.openelisglobal.config.*` from the test profile, so none of these can fail a CI run:

1. **App fails to boot** on installs with active reporting schedules whose cron time already passed: `configureTasks`' catch-up `runNow(...)` is `@Async`, and the new task decorator **throws on the unauthenticated bootstrap thread**, escaping context refresh. (Inline on `SchedulerConfig`.)
2. **Analyzer menus/permissions silently never created** on fresh DBs / new analyzers: `PluginMenuService.onApplicationReady` writes through Auditable services with no context; the throw is swallowed. (Inline on `PluginRegistryService`.)
3. **Legacy analyzer plugin registration silently broken** (Analyzer rows + test mappings never persisted), with a plausible startup-abort variant if a plugin's `connect()` reaches `bindRoleToModule` uncaught. (Same inline.)

All three are fixable with the PR's own tooling — wrap the two remaining startup writers in `executeAsDaemon(...)` and daemonize (or guard) the `configureTasks` catch-up dispatch. Please also add a boot smoke test to the E2E flow with an in-the-past cron row, since this class of bug will keep evading unit CI.

### Prior feedback reconciliation

- **Copilot:** #1 false positive (see above); #2 Javadoc — already fixed in the current head (thread still open — can be resolved); #3/#5/#6 target `AuditContextAdvice`/its tests, which were **removed** in the redesign — moot; #4 conditional-assert — fixed (now a hard `assertNotNull` prerequisite).
- **@pmanko's four questions** are all answered by the current head: the DaemonUserConfig fallback-to-"1" is gone (fail-fast now), the `principalToUserCache` is gone (staleness moot — but see my inline perf note: resolution is now an uncached per-write DB query), and the AOP-vs-service layering question resolved itself when `AuditContextAdvice` was deleted — `fillSysUserIdIfMissing` is the single enforcement point. Returning `Long` remains deferred to OGC-549, which still makes sense.

### Non-blocking notes (inline)

Per-write DB lookup cost; duplicate-`admin` hazard in the test base class; missing Liquibase `<rollback>`; strict-vs-lenient fallback inconsistency between handlers.

Happy to re-review quickly once the three startup paths are wrapped — the core of this PR is genuinely good and overdue.

mozzy11:
Pushed `40ed6af8d` fixing the three blocking findings from my review — all with the PR's own `executeAsDaemon` pattern:

1. **SchedulerConfig**: the overdue-cron catch-up dispatch (`runNow`, @Async) now runs under the daemon context, so the task decorator no longer aborts context refresh on installs whose reporting time already passed.
2. **PluginMenuService.onApplicationReady**: analyzer menu/permission registration is daemonized — role/module/role-module inserts resolve to the daemon on the context-refresh thread.
3. **PluginLoader.load()**: legacy plugin `connect()` (Analyzer rows, test mappings, menu/permission bindings) is daemonized in @PostConstruct.

`spotless:apply` + compile verified locally; CI rerunning now. @samuelmale please sanity-check I haven't stepped on anything you had in flight.

The non-blocking notes from my review (per-write uncached user lookup, dual-keyed admin helpers in the test base, missing Liquibase `<rollback>`, strict-vs-lenient fallback inconsistency) are left as follow-ups — none should hold up the merge. Dismissing my change-request since the blockers are resolved; happy for a fresh review round on the updated head.

src/main/java/org/openelisglobal/security/DaemonAuthenticationToken.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+
+/**
+ * Authentication token representing the daemon/system user. This token is
+ * always considered authenticated and carries the daemon system user ID as its
+ * principal. It is placed into the SecurityContext for scheduled tasks, async
+ * operations, and system-initiated actions where no human user is present.
+ */
```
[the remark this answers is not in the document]
mozzy11: Addressed on the current head — the Javadoc now states the principal is the constant "daemon" and the sys-user id is carried separately via getDaemonSysUserId().

src/main/java/org/openelisglobal/common/aop/AuditContextAdvice.java
```diff
@@ -0,0 +1,59 @@
...
+            + "execution(public * org.openelisglobal.common.service.AuditableBaseObjectServiceImpl+.deleteAll(..))")
+    public void enforceAndStampAuditContext(JoinPoint jp) {
+        DaemonContextAssert.assertAuthenticatedContext();
+
+        String userId = userContextHolder.getCurrentSysUserId();
+        for (Object arg : jp.getArgs()) {
+            if (arg instanceof BaseObject<?> bo) {
+                if (bo.getSysUserId() == null || bo.getSysUserId().isEmpty()) {
+                    bo.setSysUserId(userId);
+                }
```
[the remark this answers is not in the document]
mozzy11: Moot — AuditContextAdvice was removed in the redesign; the single enforcement point is now AuditableBaseObjectServiceImpl.fillSysUserIdIfMissing (404 on the PR head).

src/test/java/org/openelisglobal/security/UserContextHolderTest.java
```diff
@@ -0,0 +1,170 @@
...
+            SecurityContext ctx = SecurityContextHolder.createEmptyContext();
+            ctx.setAuthentication(new UsernamePasswordAuthenticationToken("admin", "pass"));
+            SecurityContextHolder.setContext(ctx);
+
+            SystemUser user = holder.getCurrentSysUser();
+            assertNotNull(user);
+            assertEquals(String.valueOf(adminLogin.getSystemUserId()), user.getId());
+            assertNotNull(user.getLoginName());
+        }
+    }
```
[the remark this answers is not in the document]
mozzy11: Addressed — the current test asserts the prerequisite hard (assertNotNull with an explanatory message) instead of conditionally guarding.

src/main/java/org/openelisglobal/scheduler/SchedulerConfig.java
mozzy11: Fixed in 40ed6af8d (catch-up dispatch wrapped in executeAsDaemon) and pinned by SchedulerConfigDaemonDispatchTest in a7fcc0fe8 (inversion-verified).

src/main/java/org/openelisglobal/analyzer/service/PluginRegistryService.java
mozzy11: Fixed in 40ed6af8d (PluginMenuService.onApplicationReady and PluginLoader.load both daemonized) and pinned by PluginMenuServiceDaemonContextTest + PluginLoaderDaemonContextTest in a7fcc0fe8.

src/main/java/org/openelisglobal/config/UserContextPropagatingTaskDecorator.java
mozzy11: Documented as a follow-up: keep the strict throw now that the known offenders are wrapped; contract note on AsyncConfig can come separately.

src/main/java/org/openelisglobal/common/util/UserContextHolder.java
mozzy11: Tracked as a follow-up (short-TTL/request-scoped principal cache) — non-blocking for this PR; pairs naturally with the OGC-549 type refactor.

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
mozzy11: Tracked as a follow-up: key the baseline insert on login_name='admin' OR id=1 (or merge the two helpers). Non-blocking — resolution is duplicate-tolerant today and the full suite is green.

src/main/resources/liquibase/3.5.x.x/base.xml
mozzy11: Non-blocking; numbering/order matches existing 3.5.x.x precedent. The missing <rollback> can ride any later changelog touch-up.

src/main/java/org/openelisglobal/dataexchange/resultreporting/ResultReportingTransfer.java
mozzy11: Non-blocking consistency note — fine to unify on a policy in the OGC-549 follow-up.

mozzy11 (approved):
Approving after the full verification cycle:

- **Deep review** of all 97 files + a codebase-wide sweep of every non-HTTP write path (findings posted earlier).
- **Three blocking startup regressions fixed** directly on the branch (`40ed6af8d`): daemonized the overdue-cron catch-up dispatch, PluginMenuService's ContextRefreshedEvent writer, and PluginLoader's legacy-plugin @PostConstruct path.
- **Regression tests added** (`a7fcc0fe8`) pinning the daemon-context contract on all three paths — inversion-verified (removing any wrap fails its test).
- **Full backend suite green locally** (4,703 tests; the only error is a pre-existing environment-specific failure that reproduces identically on clean develop and passes on Linux CI for both).
- **All review conversations resolved** with resolution notes; non-blocking items (principal-resolution caching, dual-keyed test admin helpers, Liquibase rollback, fallback-policy consistency) documented as follow-ups, several pairing naturally with OGC-549.

This lands a genuinely overdue foundation — thanks @samuelmale for driving it. Suggest a fresh look from @pmanko or @caseyi before merge since I've now authored commits on the branch.

PR #3373, opened 2026-04-08
Add create and update to service request provider

Isabirye1515:


src/test/java/org/openelisglobal/fhir/ServiceRequestFacadeTest.java
```diff
@@ -102,4 +103,144 @@ public void searchServiceRequest_endpointExists_shouldNotReturn404() throws Exce
...
+    public void createServiceRequest_shouldReturnCreatedServiceRequest() throws Exception {
+        cleanRowsInCurrentConnection(new String[] { "analysis" });
+        String subjectUUID = "b479ab79-5f53-4d1f-bc9b-10f19ce04635";
+        String specimenUUID = "68438220-5cef-44c4-9e6f-9f88e6b93270";
+        String practitionerUUID = "550e8400-e29b-41d4-a716-446655441004";
+        String locationUUID = "68438220-5cef-44c4-9e6f-9f88e6b93287";
+        String loinc = "123456";
+        String loincDisplay = "Blood Test";
+        MockHttpServletRequest request = buildFhirRequest("POST", "/ServiceRequest");
+        String jsonBody = """
```
mherman22: why not use an external json file here? to sort of decongest this method?

src/main/java/org/openelisglobal/sample/controller/rest/SamplePatientEntryRestController.java
```diff
@@ -315,6 +315,7 @@ public ResponseEntity<SamplePatientEntryForm> samplePatientEntrySave(HttpServlet
 
         PatientManagementUpdate patientUpdate = SpringContext.getBean(PatientManagementUpdate.class);
         patientUpdate.setSysUserIdFromRequest(request);
+        SampleUtil.testAndInitializePatientForSaving(request, patientInfo, patientUpdate, updateData);
```
mozzy11: Thanks @Isabirye1515 .
Why did you have to change the order of this method ??
it could break the EQA feature. 
Can you put this back after the EQA sample check ??

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirPersistanceServiceImpl.java
```diff
@@ -155,7 +155,6 @@ public Bundle createUpdateFhirResourcesInFhirStore(List<FhirOperations> fhirOper
                     + fhirContext.newJsonParser().encodeResourceToString(transactionResponseBundle));
         } catch (Exception e) {
             LogEvent.logError(e);
-            throw new FhirLocalPersistingException(e);
```
mozzy11, line 158: Any reason you removed throwing this ??

PR #3391, opened 2026-04-10
fix(workflow): tighten reagent deductions and MNTD handoff

dawit-Tegegnwork:
## Summary
- require reagent selection and per-sample quantity entry on the updated workflow pages
- align pharmaceutical testing with the shared reagent usage and deduction flow
- hide active reagents and instruments that do not have available lots
- fix MNTD temporary storage completion so samples move to processing instead of skipping ahead
- clean up reagent usage display so total deduction is shown clearly

## Testing
- built frontend and backend locally
- verified localhost startup
- tested reagent usage flow in MNTD, Immunology, and Pharmaceuticals
- verified MNTD temporary storage now advances fresh samples to Sample Processing Preparation
- verified invalid no-lot reagents are filtered from active workflow selection

### DIFF
--- a/frontend/src/components/notebook/pages/immunology/ImmunologyAdditionalAssaysPage.js
+++ b/frontend/src/components/notebook/pages/immunology/ImmunologyAdditionalAssaysPage.js
@@ -449,6 +449,10 @@ function ImmunologyAdditionalAssaysPage({
     const selectedReagentItems = reagents.filter((reagent) =>
       assayValues.selectedReagents.includes(reagent.id),
     );
+    if (reagents.length > 0 && selectedReagentItems.length === 0) {
+      setError("Select at least one reagent before applying assay data.");
+      return;
+    }
     const invalidReagentItems = getInvalidReagentUsageItems(
       selectedReagentItems,
       assayValues.reagentQuantities,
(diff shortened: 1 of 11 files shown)


frontend/src/components/notebook/pages/immunology/ImmunologyAdditionalAssaysPage.js
```diff
@@ -439,7 +446,29 @@ function ImmunologyAdditionalAssaysPage({
       return;
     }
 
+    const selectedReagentItems = reagents.filter((reagent) =>
+      assayValues.selectedReagents.includes(reagent.id),
+    );
+    if (reagents.length > 0 && selectedReagentItems.length === 0) {
+      setError("Select at least one reagent before applying assay data.");
```
mherman22: can we use the existing NotificationContext for this? Here is an example of where it was used https://github.com/DIGI-UW/OpenELIS-Global-2/blob/4725a09b5b99245b3499232782a81c50e87a383d/frontend/src/components/nonconform/common/NceDashboard.jsx#L251-L275

mherman22: do for all seterror instances in your changes

dawit-Tegegnwork: Updated. I switched the reagent/kit validation errors introduced in this PR to use NotificationContext across the affected workflow pages.

dawit-Tegegnwork: Hi @mherman22  I addressed the review feedback and updated the PR. It should be ready for re-review now when you have a chance. Thanks.

PR #3392, opened 2026-04-10
feat(ctd): support independent sample intake and order linking

dawit-Tegegnwork:
## Summary
- support independent sample intake through MedLab manifest import
- allow imported samples to be linked to orders and move into the collection queue
- keep bulk and individual collection handoff aligned on the sample collection page
- allow anonymous single-order creation from the link-to-order flow when no participant is linked
- capture collection-page order tracking updates for shared and independent linking flows

## Testing
- verified `mvn -q -DskipTests package`
- tested locally on localhost for MedLab sample import, order linking, and collection handoff
- verified imported samples can be linked and then collected from page 2


src/main/java/org/openelisglobal/medlab/service/MedLabPatientOrderServiceImpl.java
```diff
@@ -640,19 +741,96 @@ public Map<String, Object> createBulkIndependentOrders(List<Integer> sampleIds,
...
+        }
+
+        while (true) {
+            long nextNumber = accessionService.getNextNumberIncrement(normalizedPrefix,
+                    org.openelisglobal.common.provider.validation.AccessionNumberValidatorFactory.AccessionFormat.GENERAL);
+            String labNo = normalizedPrefix + String.format("%03d", nextNumber);
+            if (!labNumberExists(labNo)) {
+                return labNo;
+            }
+        }
```
samuelmale: If a persistent DB inconsistency makes every generated lab number "exist," this loops forever, holding a transaction open, eventually consuming the sequence.

Suggestion: Cap attempts (e.g., 1000), throw on exhaustion.

src/main/java/org/openelisglobal/medlab/service/MedLabPatientOrderServiceImpl.java
```diff
@@ -640,19 +741,96 @@ public Map<String, Object> createBulkIndependentOrders(List<Integer> sampleIds,
...
+    private boolean labNumberExists(String labNo) {
+        List<ElectronicOrder> existingOrders = electronicOrderService.getElectronicOrdersByExternalId(labNo);
+        if (existingOrders != null && !existingOrders.isEmpty()) {
+            return true;
+        }
+
+        return sampleService.getSampleByAccessionNumber(labNo) != null;
+    }
+
+    private void syncAccessionSequenceForLabNo(String labNo) {
```
samuelmale: After creating a single order with a user-supplied `labNo`, the code forcibly fast-forwards the accession sequence via `accessionService.setCurVal(prefix, …, suffixValue)` whenever the supplied suffix exceeds the current counter. A user entering "LAB-9999" on a prefix whose next value is 42 will permanently skip 9,957 numbers and create silent gaps in audit records.

@dawit-Tegegnwork is it possible not to compose lab numbers client side?

dawit-Tegegnwork:
@samuelmale  Addressed the latest review feedback and rebased onto the current demo/ethiopia branch.\n\nChanges included:\n- capped server-side lab number reservation attempts to avoid an unbounded loop\n- removed accession sequence fast-forwarding for manually supplied lab numbers so single-order creation no longer advances the counter and creates silent gaps\n- resolved the en.json merge conflict during rebase\n\nI

PR #3404, opened 2026-04-11
feat(biorepository): support intake-only legacy manifest imports

dawit-Tegegnwork:
## Summary
- support AHRI legacy workbook-style manifest intake in the biorepository upload flow
- keep manifest imports scoped to Sample Intake & Registration and skip rows without a sample ID/barcode
- harden bulk registration by using safe accession generation and stricter intake manifest validation
- raise intake sample fetch limits so large manifest imports remain visible after registration
- refresh the sample manifest test-data files to match the current intake manifest structure

## Testing
- locally tested

### DIFF
--- a/frontend/src/components/notebook/pages/biorepository/BiorepositoryIntakePage.js
+++ b/frontend/src/components/notebook/pages/biorepository/BiorepositoryIntakePage.js
@@ -41,6 +41,9 @@ import ManifestUploadModal from "./ManifestUploadModal";
 import SampleTransferTab from "./SampleTransferTab";
 import RetentionPolicySection from "./RetentionPolicySection";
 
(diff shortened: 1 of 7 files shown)


dev.docker-compose.yml
```diff
@@ -118,6 +118,9 @@ services:
         volumes:
             - './frontend/src:/app/src'
             - './frontend/public:/app/public'
+            - './frontend/package.json:/app/package.json'
+            - './frontend/package-lock.json:/app/package-lock.json'
+            - './frontend/node_modules:/app/node_modules'
```
samuelmale: Are these changes related to the Biorepo work?

dawit-Tegegnwork: I removed the unrelated dev compose commit, and the PR now contains only the manifest import change

PR #3390, opened 2026-04-09
feat: Westgard QC rules foundation with automated rule evaluation and FHIR-based pipeline

samuelmale:
# Pull Requests Requirements

- [x] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary
- Implements the Westgard Rules quality control framework for laboratory instruments (spec 003-westgard-qc), covering the full 5-layer architecture from database through React UI
- Adds 8 Westgard rule evaluators (1-2s, 1-3s, 2-2s, R-4s, 3-1s, 4-1s, 7-t, 10-x) with event-driven async evaluation triggered on QC result creation
- Replaces legacy ASTM Q-segment QC identification with a FHIR-based pipeline where the analyzer bridge evaluates configurable QC rules and tags Observations before submission

### QC Control Lots

`QCControlLot` is the anchor entity — each lot is scoped to a test + instrument (analyzer) pair and tracks manufacturer lot number, expiry, expected mean/SD, and a `calculationMethod` (`INITIAL_RUNS`, `ROLLING`, or `MANUFACTURER_FIXED`). Lots start in `ESTABLISHMENT` status; `MANUFACTURER_FIXED` lots activate immediately with seeded statistics, while the other two transition to `ACTIVE` once enough results accumulate via `StatisticsCalculator` strategy implementations (`InitialRunsCalculator`, `RollingCalculator`, `ManufacturerFixedCalculator`).

### QC Results and Statistics

`QCResultService.createQCResult()` persists each QC measurement, fetches the latest `QCStatistics` for the lot, computes the z-score, and — if the lot is `ACTIVE` — publishes a `QCResultCreatedEvent`. For `ESTABLISHMENT`-phase lots, `tryBootstrapStatistics()` checks if enough data has accumulated to compute initial mean/SD and transition the lot. `QCStatistics` holds the running mean, SD, count, and calculation method, and is recalculated on each new result for `ROLLING` lots.

### Westgard Rule Evaluation (event-driven)

[...]

PR #3443, opened 2026-04-17
feat(OGC-384): Implement privilege-based RBAC with fine-grained service authorization

reagan-meant:
- Create system_privilege and system_role_privilege tables (Liquibase)
- Seed all atomic privilege definitions across 40+ categories
- Seed role-privilege mappings for all base roles (Reception, Results, Validation, Pathologist, Cytopathologist, Reports, EQA Coordinator, User Account Administrator, Audit Trail, Analyser Import)
- Add missing operational privilege mappings (provider:view, organization:view, sample_requester:view to Reception; referral:view, storage:view to Results; referral:view to Validation; patient:view to Reports)
- Migrate legacy role assignments and deactivate legacy roles
- Add Privilege valueholder, DAO, and service with full inheritance resolution (walks grouping_parent chain, Global Admin short-circuit, circular-reference guard)
- Add Privileges.java constants — single source of truth for all privilege name strings
- Update CustomUserDetailsService to load PRIV_* authorities at login
- Close auto-allow gap in ModuleAuthenticationInterceptor (REST paths now return HTTP 401 JSON when denied)
- Replace all @PreAuthorize("hasRole('ADMIN')") on ~60+ REST controllers with fine-grained hasAuthority('PRIV_*') annotations
- Rename all PRIV_ADMIN_* to action-based names (user:manage, system:configure, test:configure, report:configure)
- Annotate all service interface methods with correct @PreAuthorize — privilege name reflects what the method returns/does, not who calls it
- Mark auth-pipeline services (PrivilegeService, SystemModuleUrlService) as @CrossDomainService to exempt from coverage scan
- Fix SearchResultsService privilege (ORDER_VIEW -> PATIENT_VIEW)
- Split ResultLimitService, RenameMethodService, RenameTestSectionService, ScriptletService, UnitOfMeasureService reads vs admin writes
- Add EndpointPrivilegeCoverageTest (ASM bytecode scan — every service method must have hasAuthority @PreAuthorize)
- Add EndpointAccessControlTest (403 for wrong privilege, pass for correct)
- Add PrivilegeServiceImplTest (direct, inherited, multi-level, Global Admin, circular reference, null/blank inputs)
- Add PrivilegeResolutionTest (full Validation->Results->Reception chain)
- Add UserServiceLabUnitFilterTest (12 tests for lab section scoping)
- Add ModuleAuthenticationInterceptorTest (auto-allow gap closed)

# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


pmanko:
Pushed two commits as maintainer on-branch. Copilot comments triage + merge notes below.

---

**1. `d54e922fb` — `Merge branch 'develop' into OGC-384`**

Scope: purely resolve conflicts. No behavior changes beyond what each conflict's chosen side dictates.

*Active text conflicts (3 files, analyzer/QC):*
- `ASTMQSegmentParser.java` — accepted develop's delete (removed by #3390 Westgard refactor).
- `QCResultExtractionService.java` — same.
- `QCResultProcessingService.java` — accepted develop's rewrite. The PR's `@PreAuthorize` import no longer applies to the new architecture; RBAC on the new QC services is tracked as follow-up (see bottom of this comment).

*Files with committed `<<<<<<< HEAD` markers left by an earlier rebase attempt (7 files, cleaned up):*
- `liquibase/3.5.x.x/base.xml` — kept both sides: develop's shipment/consent/nationality changesets and the `012-*` RBAC changesets.
- `systemuser/UserServiceImpl.java` — kept the PR's `List<Integer>` type on `currentUserRoles`, retained the develop-side `selectedRoles == null` guard.
- `reports/tat/TATReportRestController.java`, `eqa/EQA{Program,Enrollment,Distribution}RestController.java` — kept PR-side (no class-level `@PreAuthorize`). Verified the corresponding service interfaces (`TATReportService`, `TATCalculationService`, `EQAProgramService`, `EQADistributionService`, `EQAProgramEnrollmentService`) all carry method-level `@PreAuthorize("hasAuthority('PRIV_*')")`, so the endpoints remain guarded at the service layer.
- `patient/merge/PatientMergeRestController.java` — kept the PR's Global Admin role check with Integer role IDs; added a null-guard on `globalAdminRoleId` that was missing; added the three missing imports (`PostConstruct`, `List`, `RoleService`) and the `@Autowired RoleService` field that the `@PostConstruct` initializer references.

Submodules pinned to latest default-branch HEAD.

---

**2. `0a3a335a1` — `security(OGC-384): address Copilot blockers B4, B6, B8 + M1, M2`**

| Copilot # | File | Action |
|---|---|---|
| **#2 (B4)** | `ImportController` | Added class-level `@PreAuthorize("hasAuthority('PRIV_SYSTEM_CONFIGURE')")`. `/import/provider` was genuinely unauthenticated — `ProviderImportService` is `@CrossDomainService` so no service-layer check fired. |
| **#5 (B6)** | `SearchResultsService` | Moved the interface-TYPE `@PreAuthorize` onto each method. Spring Method Security doesn't reliably honor class-level `@PreAuthorize` on an interface type. |
| **#4/#6 (B8)** | `LoginPageController` L240/L244 | Traced the data flow (`LabUnitRoleMap.setRoles(Set<String>)` populated by `UserServiceImpl:113-117` from strings that `updateUserRoles` later parses with `Integer.valueOf`). Confirmed `map.getRoles()` holds role IDs-as-strings. Changed `getRoleByName(r)` → `getRoleById(Integer.valueOf(r))`. |
| **#8 (M1)** | `RoleDAOImpl.readRole()` | Split `Integer.parseInt` into its own try/catch with a clear "Invalid role ID format" message. |
| **#7 (M2)** | `PrivilegeDAOImpl.getPrivilegesForRole` | Rewrote the HQL via the mapped `RolePrivilege.privilege` association — `SELECT rp.privilege FROM RolePrivilege rp WHERE rp.roleId = :roleId AND rp.privilege.active = true`. |

**Already done by you in `0dfeed076` (so I didn't re-do):**
- **#3 (B5)** LoggingController already has class-level `@PreAuthorize("hasAuthority('PRIV_SYSTEM_CONFIGURE')")`.
- **#1 (B7)** AccessionResultsRestController `userNotInRole()` is already null-safe via `RESULT_EDIT_ROLE_ID == null || !roleIds.contains(RESULT_EDIT_ROLE_ID)`.

---

**The one remaining CI failure — worth discussing**

`EndpointPrivilegeCoverageTest` (your own ASM scan) flags **76 missing `@PreAuthorize` annotations across 9 services** that develop introduced while this PR was open:

- `AnalyzerQcRuleService`, `QCResultProcessingService`, `QCAlertService`, `QCChartDataService`, `QCControlLotService`, `QCDashboardService`, `QCResultService`, `QCRuleViolationService` (from #3390 Westgard QC)
- `PatientIdDocumentService` (from #3420)

The scan is working exactly as designed — it caught that those services shipped without the RBAC annotations this PR requires. None of this is code you wrote; it's develop-side drift your test exposes.

Proper fix, roughly:
1. Add `PRIV_QC_VIEW` / `PRIV_QC_MANAGE` / `PRIV_PATIENT_DOCUMENT_*` constants to `Privileges.java`.
2. Seed them in `012-003-seed-privileges.xml` and map to the right roles in `012-004-seed-role-privilege-mappings.xml`.
3. Annotate each of the 76 methods with the appropriate privilege.

Two ways to go:
- **(a)** I (or you) take it on as a focused follow-up commit here.
- **(b)** We scope it to a separate PR that follows this one, and mark the 9 interfaces `@CrossDomainService` temporarily with a clear `TODO(OGC-XXX)` on each, so this PR's CI goes green and the QC RBAC work gets its own dedicated review.

Let me know which you prefer.

---

**Deferred items from the maintainer review (not in scope for either commit above):**

- **M3** Integer end-to-end through `PrivilegeServiceImpl` (method signatures still take `String roleId` and `Integer.parseInt` on every call).
- **M4** Expand `EndpointAccessControlTest` — currently 5 tests, ~4 endpoints. Missing multi-role, no-role, case-sensitivity, and broader endpoint coverage (patient, report, audit, analyzer).
- **M5** `<rollback>` blocks on `012-00{3-7}-*.xml` — the legacy-role deactivation step especially should be reversible.
- **M6** Audit the `@CrossDomainService` exempt list. `ProviderImportService` is a clear candidate to remove from the list (it's called from a user-facing REST endpoint).

These can be a third commit here or a follow-up PR — your call. Happy to handle any of them.

src/main/java/org/openelisglobal/common/security/SystemAwareSecurityExpressionRoot.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    public boolean hasPrivilege(String name) {
+        if (SystemInitFlag.isSet()) {
+            return true;
+        }
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        if (auth == null) {
+            return true;
+        }
```
ibacher: I thought I already commented this, but returning `true` here seems quite dangerous since we're effectively treating a non-authenticated user as having every privilege. This same comment applies to every function in this class.

src/main/java/org/openelisglobal/common/security/SystemAwareSecurityExpressionRoot.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Override
+    public boolean isAuthenticated() {
+        if (SystemInitFlag.isSet()) {
+            return true;
+        }
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        if (auth == null) {
+            return true;
+        }
```
ibacher: This is also a bizarre choice.

src/main/java/org/openelisglobal/common/security/SystemInitBeanPostProcessor.java
```diff
@@ -0,0 +1,32 @@
...
+    public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
+        SystemInitFlag.set();
+        return bean;
+    }
+
+    @Override
+    public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+        SystemInitFlag.clear();
+        return bean;
+    }
```
ibacher: The whole setup here is problematic. Basically this code assumes that `postProcessAfterInitialization()` is always called after `postProcessBeforeInitialization()`, but in point of fact [this isn't guaranteed](https://github.com/spring-projects/spring-framework/blob/2f458f909391b04eb138aba8980598dc4b0cf4a3/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractAutowireCapableBeanFactory.java#L1805). 

The mechanism here is also not quite correct in that these methods are intended to be scoped to the creation of a specific bean, but the mechanism we're trying to implement is meant to be scoped to "application startup". Basically, `postProcessBeforeInitialization()` and `postProcessAfterInitialization()` are called for _every_ bean Spring creates, which means this is flipping the context more frequently than we need. A better setup might be to set the flag in `AnnotationWebAppInitializer` and unset it in a listener for Spring's `ContextRefreshedEvent` or something like that.

src/main/java/org/openelisglobal/interceptor/ModuleAuthenticationInterceptor.java
```diff
@@ -105,12 +105,13 @@ private boolean hasPermissionForUrl(HttpServletRequest request, boolean useParam
...
-            // Full per-role module mappings are a future enhancement.
+            // REST endpoints with no SystemModuleUrl entry are denied by default.
+            // Access control for REST endpoints is handled by @PreAuthorize annotations
+            // using privilege authorities (PRIV_*) loaded at login. An unregistered
+            // REST path reaching here means it has no module mapping — deny it so
+            // that @PreAuthorize is the sole gate (closes the auto-allow gap).
             if (isRestFullPath()) {
-                return true;
+                return false;
             }
```
ibacher: I realize this isn't a new pattern introduced here, but the `path` (which `isRestFullPath()` relies on) is inherently not thread-safe making all of this code subject to kind of trivial race conditions. It would probably be best to pass the `request` to `isRestFullPath()` and extract the defined path at that point. This matters more with the change here to deny-by-default, since it's harder to guarantee that policy correctly applies in the face of the race condition.

src/test/java/org/openelisglobal/security/EndpointAccessControlTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.web.WebAppConfiguration;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+
+/**
+ * T019 — Parameterized access-control integration test.
```
ibacher: In general, do not add these kinds of things to Javadoc or comments since T019 is not referenceable outside of this jobs specs, etc.

src/test/java/org/openelisglobal/privilege/service/PrivilegeServiceImplTest.java
ibacher, line 1: This could use some additional checks, e.g., for a scenario where we have:

```
       Role 1
      /       \
   Role 2    Role 3
     \        /
       Role 4
```

Where a user is given `Role 4`, just to ensure `Role 1`'s privileges are only added once.

src/main/java/org/openelisglobal/security/login/CustomUserDetailsService.java
```diff
@@ -69,6 +75,23 @@ private List<GrantedAuthority> getGrantedAuthorities(LoginUser user) {
...
 
+        // Load resolved privileges as PRIV_ authorities
+        if (user != null && user.getSystemUserId() > 0) {
+            Set<String> resolvedPrivileges = privilegeService
+                    .getAllPrivilegesForUser(String.valueOf(user.getSystemUserId()));
+            if (resolvedPrivileges.contains(Privileges.GLOBAL_ADMIN_SENTINEL)) {
+                // Global Admin gets every privilege — load all from DB
+                List<Privilege> allPrivileges = privilegeService.getAllPrivileges();
+                for (Privilege p : allPrivileges) {
+                    authorityNames.add("PRIV_" + p.getName().toUpperCase().replaceAll("[^A-Z0-9]+", "_"));
```
ibacher: It would probably be best to refactor things to extract this pattern into a Java `Pattern` that can be shared across the places it is used. Also consider adopting the normalization we apply to roles that removes double `_`. It may make sense to  refactor that normalization into it's own method.

src/main/java/org/openelisglobal/security/login/CustomUserDetailsService.java
```diff
@@ -69,6 +75,23 @@ private List<GrantedAuthority> getGrantedAuthorities(LoginUser user) {
...
+                }
+            } else {
+                for (String priv : resolvedPrivileges) {
+                    authorityNames.add("PRIV_" + priv.toUpperCase().replaceAll("[^A-Z0-9]+", "_"));
+                }
+            }
```
ibacher, line 91: This construct could be simplified:


```suggestion
            Set<String> resolvedPrivileges = privilegeService
                    .getAllPrivilegesForUser(String.valueOf(user.getSystemUserId()));
            if (resolvedPrivileges.contains(Privileges.GLOBAL_ADMIN_SENTINEL)) {
                resolvedPrivileges = privilegeService.getAllPrivileges();
            }
            
            for (String priv : resolvedPrivileges) {
                authorityNames.add("PRIV_" + priv.toUpperCase().replaceAll("[^A-Z0-9]+", "_"));
            }
```

src/main/java/org/openelisglobal/privilege/service/PrivilegeServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public List<Privilege> getDirectPrivilegesForRole(String roleId) {
+        if (GenericValidator.isBlankOrNull(roleId)) {
+            return Collections.emptyList();
+        }
+        try {
+            return getDirectPrivilegesForRole(Integer.valueOf(roleId));
+        } catch (NumberFormatException e) {
+            return Collections.emptyList();
+        }
```
ibacher, line 51: It probably makes sense to log the failure here since we expect `roleId` to be convertible to a integer.

src/main/java/org/openelisglobal/privilege/service/PrivilegeServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Override
+    public Set<String> resolveAllPrivilegesForRole(String roleId) {
+        if (GenericValidator.isBlankOrNull(roleId)) {
+            return Collections.emptySet();
+        }
+        try {
+            return resolveAllPrivilegesForRole(Integer.parseInt(roleId), new HashSet<>());
+        } catch (NumberFormatException e) {
+            return Collections.emptySet();
```
ibacher, line 88: As above.

src/main/java/org/openelisglobal/role/service/RoleServiceImpl.java
```diff
@@ -56,7 +56,7 @@ public Role getRoleByName(String name) {
             LogEvent.logWarn(this.getClass().getSimpleName(), "getRoleByName",
                     "Role not found in database: '" + name + "'");
             Role stub = new Role();
-            stub.setId("-1");
+            stub.setId(-1);
             stub.setName(name);
```
ibacher, line 65: While not net-new, we _are_ returning what will appear to the caller to be a valid role with a valid name. It seems safer to just return `null`.

src/main/java/org/openelisglobal/privilege/daoimpl/PrivilegeDAOImpl.java
```diff
@@ -0,0 +1,34 @@
...
+import org.hibernate.Session;
+import org.hibernate.query.Query;
+import org.openelisglobal.common.daoimpl.BaseDAOImpl;
+import org.openelisglobal.privilege.dao.PrivilegeDAO;
+import org.openelisglobal.privilege.valueholder.Privilege;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
+
+@Repository
+@Transactional
```
ibacher: I'm pretty sure the intention is for transactions to be handled at the service layer. That's at least what it says in the constitution.

src/main/java/org/openelisglobal/security/login/CustomUserDetailsService.java
ibacher: However, it may be worth considering _not_ loading all privileges for an admin user since that can be fairly heavy. In OpenMRS, our privilege checks end up looking like [this](https://github.com/openmrs/openmrs-core/blob/9490dcf21f1e10d276430ee7f0cc537f2b354bd8/api/src/main/java/org/openmrs/User.java#L181) where we just explicitly check for whether the user is an administrator to avoid needing to do this.

src/main/java/org/openelisglobal/common/service/SystemPipelineService.java
```diff
@@ -0,0 +1,30 @@
...
+ * <p>
+ * <strong>Do not use this annotation to avoid adding authorization.</strong> It
+ * is only for services that are genuinely unreachable from user-initiated
+ * request paths. User-facing services must have per-method
+ * {@code @PreAuthorize("hasPrivilege('PRIV_*')")} annotations.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface SystemPipelineService {
```
ibacher: While this is probably a good idea, this doesn't appear to be used anywhere to actually by-pass privilege checks.

src/main/java/org/openelisglobal/common/security/SystemInitBeanPostProcessor.java
ibacher: A third issue I forgot about (thanks Claude!) is that `BeanPostProcessor`s are triggered for all beans, which means this can run on things with a different `@Scope` than singleton, which is a pattern apparently used in OE, which can result in privilege by-passes where those beans are used.

src/main/java/org/openelisglobal/common/security/SystemAwareSecurityExpressionRoot.java
```diff
@@ -0,0 +1,175 @@
...
+ */
+public class SystemAwareSecurityExpressionRoot implements MethodSecurityExpressionOperations {
+
+    private final MethodSecurityExpressionOperations delegate;
+
+    public SystemAwareSecurityExpressionRoot(MethodSecurityExpressionOperations delegate) {
+        this.delegate = delegate;
+    }
+
+    public boolean hasPrivilege(String name) {
```
ibacher, line 20: Do we actually use this? We seem to use `hasAuthority()` everywhere which is, in any case, pretty much an exact duplicate of this method. If we really need `hasPrivilege()` for some reason it should probably just call `this.hasAuthority(name)`.

src/main/java/org/openelisglobal/common/security/SystemInitBeanPostProcessor.java
ibacher: Also, I asked Claude to look for places where we need this and it seems like the only thing is `CreateAdminUserTask`, which could be handled similarly to how we handle things [here](https://github.com/reagan-meant/OpenELIS-Global-2/blob/2e4cbd04aa84a701f48da085c94563986de4cdfa/src/main/java/org/openelisglobal/security/login/CustomSSOAuthenticationSuccessHandler.java#L87), which would be more explicit about the fact it was by-passing privilege checks.

src/main/java/org/openelisglobal/common/security/SystemAwareSecurityExpressionRoot.java
ibacher: Realistically, if we don't need `hasPrivilege()` we can probably dispense with this class altogether, since I also don't think we need the `SystemInitFlag` hack, which would greatly simplify this PR.

src/main/java/org/openelisglobal/common/security/SystemAwareSecurityExpressionRoot.java
```diff
@@ -0,0 +1,140 @@
...
+        if (SystemInitFlag.isSet()) {
+            return true;
+        }
+        return delegate.hasAuthority(name);
+    }
+
+    @Override
+    public boolean hasAuthority(String authority) {
+        if (SystemInitFlag.isSet()) {
```
mozzy11, line 29: With #3356 merged, develop now has a sanctioned identity for system work (`DaemonAuthenticationToken` via `DaemonContextExecutor`), so this all-checks-pass bypass — which ibacher flagged as dispensable — can be removed outright in the rebase, along with `SystemInitFlag`, `SystemInitBeanPostProcessor`, and `SystemAwareMethodSecurityExpressionHandler`. Startup writers on develop are already wrapped in `executeAsDaemon()`.

src/main/java/org/openelisglobal/common/security/SystemContextTaskDecorator.java
```diff
@@ -0,0 +1,57 @@
...
+ * not lose them across the thread hop.</li>
+ * </ul>
+ */
+public final class SystemContextTaskDecorator {
+
+    private SystemContextTaskDecorator() {
+    }
+
+    public static TaskDecorator systemContext() {
+        return runnable -> () -> {
```
mozzy11, line 31: Develop's `AsyncConfig` now installs `UserContextPropagatingTaskDecorator` (#3356), which propagates the real SecurityContext and fails loud on unauthenticated @Async submissions. This decorator duplicates that with a flag instead of an Authentication — it should go away in the rebase; system-initiated submissions become `DaemonContextExecutor.executeAsDaemon(...)` around the submit.

src/main/java/org/openelisglobal/config/AsyncConfig.java
```diff
@@ -2,18 +2,25 @@
...
 @EnableAsync
 public class AsyncConfig extends AsyncConfigurerSupport {
     @Override
     public Executor getAsyncExecutor() {
-        return new SimpleAsyncTaskExecutor();
+        SimpleAsyncTaskExecutor executor = new SimpleAsyncTaskExecutor();
+        // A system-context submitter (startup config reload, scheduled jobs) must
+        // not lose its context across the async thread hop; user submitters keep
+        // their own Authentication via the security-context delegating wrapper.
+        executor.setTaskDecorator(SystemContextTaskDecorator.propagateSystemContext());
```
mozzy11: Direct conflict with develop: #3356 already set a task decorator here (`UserContextPropagatingTaskDecorator`). On rebase, keep develop's version — but note its decorator throws on unauthenticated submission, so any @Async call site this PR relied on `propagateSystemContext()` for (e.g. the FHIR order persistence fix) must instead be wrapped in `executeAsDaemon()` at the submitter.

src/main/resources/liquibase/3.5.x.x/012-001-create-system-privilege.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mozzy11, line 1: The 012-* numbering predates the current tree — develop's 3.5.x.x sequence now runs past 061 (and #3356 added 003-daemon-system-user.xml). Worth renumbering to the current tail on rebase. Related: the daemon user seeded by #3356 gets no rows in 012-004's role-privilege mappings — its access story should be decided as part of the daemon-vs-PRIV_* gate question.

src/main/java/org/openelisglobal/privilege/daoimpl/PrivilegeDAOImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+@Repository
+public class PrivilegeDAOImpl extends BaseDAOImpl<Privilege, Integer> implements PrivilegeDAO {
+
+    public PrivilegeDAOImpl() {
+        super(Privilege.class);
+    }
+
+    @Override
+    @Transactional(readOnly = true)
```
mozzy11, line 21: Still open from ibacher's review: the constitution puts @Transactional at the service layer only. `PrivilegeServiceImpl` (or its interface method) should own the transaction; the DAO method can drop the annotation.

src/main/java/org/openelisglobal/security/login/CustomUserDetailsService.java
```diff
@@ -69,6 +79,21 @@ private List<GrantedAuthority> getGrantedAuthorities(LoginUser user) {
             addAuthoritiesForRole(Constants.ROLE_GLOBAL_ADMIN, authorityNames);
         }
 
+        // Load resolved privileges as PRIV_ authorities
+        if (user != null && user.getSystemUserId() > 0) {
+            Set<String> resolvedPrivileges = privilegeService
+                    .getAllPrivilegesForUser(String.valueOf(user.getSystemUserId()));
+            if (resolvedPrivileges.contains(Privileges.GLOBAL_ADMIN_SENTINEL)) {
+                resolvedPrivileges = new HashSet<>();
+                for (Privilege p : privilegeService.getAllPrivileges()) {
```
mozzy11, line 88: Ibacher's suggestion to avoid materialising the whole privilege catalog for admins is still open — every admin login loads and normalises all ~hundreds of privileges into authorities. A lighter option: keep the sentinel authority (e.g. `PRIV_ALL`) and let the gate check `hasAuthority('PRIV_ALL') or hasAuthority('PRIV_X')` — which would also dovetail with however the daemon token ends up satisfying gates.

mozzy11 (changes requested):
Thanks for the sustained work on this — the privilege model itself (schema, seed data, inheritance resolution with diamond/cycle guards, coverage scan test, migration changesets with rollbacks) is in good shape, and most of @ibacher's May feedback has been addressed (interceptor thread-safety, Pattern extraction, parse-failure logging, diamond test, SystemPipelineService removal).

The blocking issue is that develop has since merged #3356 (Daemon/System user + centralised UserContext, 61880ada2), which supersedes this PR's entire system-context mechanism — exactly the part ibacher asked to dispense with. `SystemInitFlag`, `SystemInitBeanPostProcessor`, `SystemAwareSecurityExpressionRoot`/`SystemAwareMethodSecurityExpressionHandler`, and `SystemContextTaskDecorator` should be dropped in the rebase in favour of develop's `DaemonContextExecutor.executeAsDaemon()` + `DaemonAuthenticationToken` + `UserContextPropagatingTaskDecorator` (the 8 textual conflicts — AsyncConfig, SchedulerConfig, ConfigurationInitializationService, RolesConfigurationHandler, AppTestConfig, etc. — are all in this zone; develop already daemonizes those startup paths).

One design decision to settle for the rebase: `DaemonAuthenticationToken` carries only `ROLE_SYSTEM`, so every `hasAuthority('PRIV_*')` gate would deny the daemon and the daemonized startup writers (plugin menu registration, scheduler catch-up, plugin loader) would fail at boot. Options: grant the daemon token the full PRIV_* set, or extend the gate semantics (e.g. a custom AuthorizationManager that permits DaemonAuthenticationToken) — worth agreeing with @ibacher before reworking.

Remaining smaller items: DAO-level @Transactional (constitution: service layer only), ibacher's note on not materialising the full privilege catalog for admins, and renumbering the 012-* changesets against develop's current sequence. Requesting changes solely for the rebase/rework — the underlying RBAC design is sound and worth landing.

src/main/java/org/openelisglobal/provider/service/ProviderImportService.java
```diff
@@ -1,10 +1,19 @@
 package org.openelisglobal.provider.service;
 
 import java.io.IOException;
+import org.openelisglobal.common.service.CrossDomainService;
 import org.openelisglobal.dataexchange.fhir.exception.FhirGeneralException;
 import org.openelisglobal.dataexchange.fhir.exception.FhirLocalPersistingException;
 
+@CrossDomainService(callers = "FHIR provider import pipeline — internal infrastructure")
```
reagan-meant: **[BLOCKER] `/import/provider` is reachable by any authenticated user — the guard added earlier in this PR was lost in the rebase**

`@CrossDomainService` exempts this *entire interface* from `ServicePrivilegeCoverageTest`, and `importPractitionerList()` carries no `@PreAuthorize`. `ImportController` (`admin/controller/ImportController.java`) has no class- or method-level guard at head, and `/import/**` is not in `SecurityConfig.OPEN_PAGES`, so it falls through to `.anyRequest().authenticated()` — **any** authenticated user (Reception, Results, a phlebotomist) can `GET /import/provider` or `/import/all` and drive a full FHIR practitioner import.

The asymmetry on the very same controller makes the intent clear: `OrganizationImportService.importOrganizationList()` *is* gated `PRIV_ORGANIZATION_MANAGE`, and its `callers=` string even says "the admin Import endpoint (gated importOrganizationList)". This one says "not exposed by any controller", which is factually wrong — `ImportController:28` calls it.

Note this is a **regression**: pmanko's maintainer commit `0a3a335a1` added `@PreAuthorize("hasAuthority('PRIV_SYSTEM_CONFIGURE')")` to `ImportController` to fix exactly this (Copilot B4), and that commit is no longer in the branch history.

**Suggested fix:** Gate `importPractitionerList()` with `@PreAuthorize("hasAuthority('PRIV_PROVIDER_MANAGE')")` (mirroring the Organization side) and restore the class-level guard on `ImportController`. Correct the `callers=` text — it is the only documentation a future reader has for why the exemption is safe.

src/main/java/org/openelisglobal/security/certs/service/TruststoreService.java
```diff
@@ -5,7 +5,9 @@
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
+import org.openelisglobal.common.service.CrossDomainService;
 
+@CrossDomainService(callers = "external connection security — internal infrastructure")
```
reagan-meant: **[BLOCKER] Any authenticated user can install a trusted CA certificate into the JVM truststore**

`addTrustedCert` has no `@PreAuthorize`, and `@CrossDomainService` exempts the whole interface from the coverage scan. The justification says "internal infrastructure", but it is directly reachable from user input: `ExternalConnectionController` (`@Controller`, **no `@PreAuthorize` anywhere in the file** — its `hasRole('ADMIN')` was removed in this PR) exposes `POST /ExternalConnection`, which calls `loadCertificateIntoTruststore(...)` → `truststoreService.addTrustedCert(...)` with an uploaded `MultipartFile`.

Worse, look at the ordering in `addEditExternalConnection`: `loadCertificateIntoTruststore` is invoked at line ~121, **before** either `externalConnectionService.createNewExternalConnection(...)` or `updateExternalConnection(...)`. So even if those service methods are privilege-gated and throw `AccessDeniedException`, the attacker-supplied X.509 certificate has **already been written into the truststore**. That certificate is then trusted for outbound TLS across the application — a path to MITM of external lab/FHIR integrations.

**Suggested fix:** Gate `addTrustedCert` with `@PreAuthorize("hasAuthority('PRIV_EXTCONNECTION_MANAGE')")` (or equivalent) and remove the blanket `@CrossDomainService`. Also move the truststore write to *after* the authorization-checked persistence call so a denial cannot leave a side effect behind.

src/main/java/org/openelisglobal/analyzerresults/service/AnalyzerResultsAcceptService.java
```diff
@@ -12,6 +13,7 @@
  * Extracted from AnalyzerResultsController so that business logic lives in the
  * service layer where it belongs.
  */
+@CrossDomainService(callers = "analyzer import pipeline, result entry — cross-domain")
```
reagan-meant: **[HIGH] Accepting analyzer results into the patient record has no privilege gate at all**

`acceptAndPersist(...)` builds Sample/Analysis/Result domain objects and persists them, i.e. it writes clinical results. It carries no `@PreAuthorize`, and `@CrossDomainService` exempts the whole interface from `ServicePrivilegeCoverageTest`, so the omission is invisible to CI.

The caller is user-facing, not a pipeline: `result/controller/AnalyzerResultsController` — a `@Controller` with no `@PreAuthorize` on the class or on any handler — invokes it from `POST /rest/AnalyzerResults` (:796) and `POST /AnalyzerResults` (:825), both taking the result list straight from the request. With the interceptor now failing open on unmapped `/rest` paths (see my comment on `ModuleAuthenticationInterceptor`), any authenticated user can accept/reject/delete analyzer results.

This is inconsistent with how the sibling result path was handled: `LogbookResultsPersistService.persistDataSet` *does* carry `hasAnyAuthority('PRIV_RESULT_ENTER','PRIV_RESULT_MODIFY')`.

**Suggested fix:** Annotate `acceptAndPersist` with the same `hasAnyAuthority('PRIV_RESULT_ENTER','PRIV_RESULT_MODIFY')` and drop the interface-wide exemption. (Separately: `ServicePrivilegeCoverageTest.hasAuthorityExpression()` explicitly rejects `hasAnyAuthority(`, so `LogbookResultsPersistService` only passes CI because of *its* `@CrossDomainService` — the scan should accept `hasAnyAuthority` rather than forcing legitimate two-privilege methods to opt out of coverage entirely.)

src/main/java/org/openelisglobal/systemuser/service/UserService.java
```diff
@@ -10,34 +10,47 @@
...
 
+    @PreAuthorize("hasAuthority('PRIV_USER_MANAGE')")
     void updateLoginUser(LoginUser loginUser, boolean loginUserNew, SystemUser systemUser, boolean systemUserNew,
             List<String> selectedRoles, String loggedOnUserId);
 
+    @PreAuthorize("hasAuthority('PRIV_USER_MANAGE')")
     void saveUserLabUnitRoles(SystemUser systemUser, Map<String, Set<String>> selectedLabUnitRolesMap,
             String loggedOnUserId);
 
+    @PreAuthorize("hasAuthority('PRIV_RESULT_VIEW')")
```
reagan-meant: **[HIGH] Privilege mismatch: `PRIV_RESULT_VIEW` guards a user-administration read**

`getUserLabUnitRoles` returns another user's **role and lab-unit assignments** — identity/administration data, not result data. Gating it on `PRIV_RESULT_VIEW` means every technician who can view results can enumerate the RBAC assignments of any user id they supply.

The same underlying operation is gated correctly one layer down: `UserRoleService.getUserLabUnitRoles` requires `PRIV_USER_ROLE_VIEW`. But `UserServiceImpl.getUserLabUnitRoles` is a thin delegate to it, so routing through `UserService` **downgrades** the required privilege from `PRIV_USER_ROLE_VIEW` to `PRIV_RESULT_VIEW`.

That weaker gate is reachable with a caller-controlled id: `GET /rest/UnifiedSystemUser?ID=<any>` (no controller `@PreAuthorize` — its `hasRole('ADMIN')` was removed here) → `setPropertiesForExistingUser` → `userService.getUserLabUnitRoles(form.getSystemUserId())` (UnifiedSystemUserRestController:686), returning the target's login name, account enabled/locked state, and role assignments. (The password field is obfuscated, so no credential leak.)

**Suggested fix:** Change this to `hasAuthority('PRIV_USER_ROLE_VIEW')` so it cannot undercut the delegate, and add a controller-level guard on `UnifiedSystemUserRestController` — a user-administration screen should not be gated only by whatever the services it happens to touch require.

reagan-meant (commented):
Independent security review of the full diff against merge-base `eed1310`. Inline findings below; summary comment follows separately.

src/main/java/org/openelisglobal/interceptor/ModuleAuthenticationInterceptor.java
```diff
@@ -104,10 +104,13 @@ private boolean hasPermissionForUrl(HttpServletRequest request, boolean useParam
...
-            // auto-allowed for any authenticated user. Admin-only controllers are
-            // protected via @PreAuthorize("hasRole('ADMIN')") (added in PR #2794).
-            // Full per-role module mappings are a future enhancement.
+            // REST endpoints with no SystemModuleUrl entry are allowed PAST this
+            // interceptor for any authenticated user; their authorization is then
+            // enforced by the service-layer @PreAuthorize (PRIV_*) gates. This
+            // interceptor runs BEFORE method security, so returning false here would
+            // deny outright — breaking every non-admin on unmapped infrastructure
+            // endpoints (menu, configuration-properties, home-dashboard/metrics,
+            // user-test-sections). Defer to @PreAuthorize instead of denying.
```
reagan-meant, line 120: **[HIGH] The PR description claims this gap was closed; at head it fails open — and ~80 controller guards were removed on that assumption**

The PR body states: *"Close auto-allow gap in ModuleAuthenticationInterceptor (REST paths now return HTTP 401 JSON when denied)"*. That change was made in `acc3b5408` but **reverted in `b2381dddf`** ("two non-admin regressions caught by live RBAC testing"). At head, `return true` on line 115 means any unmapped `/rest`, `/api`, `/Provider`, `/dbImage`, or `/logging` path is allowed past for **any** authenticated user. `ModuleAuthenticationInterceptorIntegrationTest` now asserts `assertTrue(allowed)` in all three tests, codifying fail-open.

I'm not arguing the revert was wrong — deny-by-default here genuinely did break non-admins. The problem is the combination: this PR also strips `hasRole('ADMIN')` from roughly 80 controllers, and the stated justification for that is that the interceptor plus service-layer gates cover them. With the interceptor failing open, **the service `@PreAuthorize` annotations are the only remaining control** on those endpoints. That is exactly why the three ungated `@CrossDomainService` paths I flagged separately (provider import, truststore, analyzer-result accept) are directly exploitable rather than defence-in-depth issues.

Please reconcile the PR description with the code before merge, so reviewers and operators aren't relying on a control that isn't there.

**Suggested fix:** Either seed `system_module_url` rows for the infrastructure endpoints that broke and restore deny-by-default, or explicitly document fail-open as the accepted design and treat 100% service-gate coverage as a merge blocker (which requires closing the `@CrossDomainService` holes).

src/main/java/org/openelisglobal/interceptor/ModuleAuthenticationInterceptor.java
```diff
@@ -56,7 +56,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
         if (!hasPermission(errors, request, path)) {
             LogEvent.logInfo("ModuleAuthenticationInterceptor", "preHandle()",
                     "======> NOT ALLOWED ACCESS TO THIS MODULE");
-            LogEvent.logInfo(this.getClass().getSimpleName(), "preHandle", "has no permission"); //
+            LogEvent.logInfo(this.getClass().getSimpleName(), "preHandle", "has no permission");
             if (isRestFullPath(path)) {
                 response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
                 response.setContentType("application/json");
```
reagan-meant, line 69: **[MEDIUM] This returns 401 for an authorization failure; the rest of the stack returns 403 for the same condition**

Unauthenticated requests never reach this interceptor — `SecurityConfig` `.anyRequest().authenticated()` runs in the filter chain, well before Spring MVC `HandlerInterceptor`s. So every 401 emitted here is for a user who *is* authenticated but lacks module permission. That is 403 semantics; 401 additionally implies a `WWW-Authenticate` challenge, which isn't set.

This is inconsistent within the same PR: `SecurityConfig`'s `AccessDeniedHandler` returns **403** with a JSON body for `/rest`, `/api`, and `/Provider` — the identical failure class. A client hitting the same logical denial therefore gets 401 from the interceptor and 403 from method security depending on which layer rejects first. Frontends commonly treat 401 as "session expired → force re-login", so an authorization denial here will spuriously log the user out instead of showing a permission error.

**Suggested fix:** Return `SC_FORBIDDEN` here (and update the hardcoded `\"status\": 401` string in the body, which can drift from the status independently) to match the `AccessDeniedHandler`.

src/main/java/org/openelisglobal/scheduler/SchedulerConfig.java
```diff
@@ -76,16 +74,25 @@ public long getResultsResendTimeMillis() {
...
+        ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
+        scheduler.setPoolSize(10);
+        scheduler.setThreadNamePrefix("oe-scheduled-");
+        // @Scheduled jobs are system-initiated: run them in system context so
+        // service-layer @PreAuthorize gates pass without a user Authentication.
+        // NOTE: preferred over develop's DelegatingSecurityContext + daemon token
+        // here because the DaemonAuthenticationToken carries only ROLE_SYSTEM, not
+        // the PRIV_* authorities the service gates check; SystemInitFlag is what
+        // SystemAwareSecurityExpressionRoot short-circuits on.
+        scheduler.setTaskDecorator(SystemContextTaskDecorator.systemContext());
```
reagan-meant, line 87: **[HIGH] `SystemInitFlag` on a pooled scheduler makes every `@PreAuthorize` in the app return `true` for the duration of each scheduled task**

`SystemContextTaskDecorator.systemContext()` sets a ThreadLocal that `SystemAwareSecurityExpressionRoot` short-circuits on — and it short-circuits *unconditionally*, for `hasAuthority`, `hasRole`, `hasPermission`, and `isAuthenticated` alike. So this isn't "the scheduler gets the privileges it needs"; it is "all authorization is disabled on this thread", with no scoping to the job's actual needs.

The blast radius is larger than it looks because `@Scheduled` methods self-invoke `@Async` ones. `FhirApiWorkFlowServiceImpl.pollForRemoteTasks()` (`@Scheduled`) calls `processWorkflow(...)` on line 90 as a **plain `this.` call**, so the `@Async` proxy is bypassed and the entire FHIR import/transform workflow executes synchronously on this pool thread with all gates disabled. Any privilege regression in that path is silently unenforced, and the coverage test can't see it either.

This also directly conflicts with `develop`: `AsyncConfig` already uses `UserContextPropagatingTaskDecorator` (#3356), so the codebase now has two competing system-identity mechanisms — a real `Authentication` for `@Async` and a blanket bypass flag for `@Scheduled`.

**Suggested fix:** Drop `SystemContextTaskDecorator` here in favour of develop's daemon identity (`DelegatingSecurityContext*` seeded from `DaemonContextExecutor.createDaemonSecurityContext()`). That requires settling the open question @mozzy11 raised — `DaemonAuthenticationToken` carries only `ROLE_SYSTEM`, so it satisfies no `hasAuthority('PRIV_*')` gate. Granting the daemon token an explicit privilege set (or having the gates accept `ROLE_SYSTEM`) is auditable; a flag that makes every check return `true` is not.

src/main/java/org/openelisglobal/privilege/service/PrivilegeServiceImpl.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        Role role = roleService.getRoleById(roleId);
+        if (role == null) {
+            return Collections.emptySet();
+        }
+
+        // Global Administrator gets every privilege. Trim: system_role.name is
+        // CHARACTER(30) in the legacy schema, so values come back space-padded.
+        if (role.getName() != null && Constants.ROLE_GLOBAL_ADMIN.equals(role.getName().trim())) {
```
reagan-meant: **[MEDIUM] Global Admin is identified by a mutable display name, and the role name is admin-editable**

The all-privileges short-circuit keys on `Constants.ROLE_GLOBAL_ADMIN.equals(role.getName().trim())` — the literal string `"Global Administrator"`. `system_role.name` is ordinary editable data (`character(30)`, surfaced in the role admin UI), not a stable security identifier, so this bypass is reachable by *naming* rather than by identity.

Two concrete consequences:
- **Silent privilege loss.** Rename or re-localise the role (or a site that seeded it as `"Admin"` — a spelling this codebase explicitly anticipates in `UnifiedSystemUserController.initialize()`) and the match fails. Resolution then returns only the role's directly-mapped rows, which for a grouping role is typically empty, so every global admin quietly drops to near-zero privileges after upgrade.
- **Privilege escalation by naming.** Conversely, anyone able to create or rename a role to exactly `"Global Administrator"` grants its holders the entire privilege catalog, with no row ever added to `system_role_privilege`.

Note the comparison here is case-**sensitive** while `CustomUserDetailsService` (`:119`, `:131`) uses `equalsIgnoreCase` for the same concept, so `"GLOBAL ADMINISTRATOR"` yields `ROLE_ADMIN` but *not* the privilege short-circuit — the two halves of the admin model can disagree.

**Suggested fix:** Key the short-circuit on a stable identifier — the seeded role id/UUID, or better an explicit non-editable `is_global_admin` boolean column set by migration — and make the role name non-editable for that row. At minimum, align the case handling with `CustomUserDetailsService`.

reagan-meant (commented):
Second batch of inline findings.

src/main/java/org/openelisglobal/role/service/RoleService.java
```diff
@@ -2,20 +2,43 @@
...
     List<Role> getPageOfRoles(int startingRecNo);
 
+    // Name->role resolution is an operational primitive (getUserTestSections,
+    // logbook/results/reports screens all resolve a role id by name for the
+    // current user's own workflow), NOT the role-administration surface — those
+    // methods (getAllRoles/getPageOfRoles/getReferencingRoles) keep PRIV_ROLE_VIEW.
+    // Gate at PRIV_ORDER_VIEW, the privilege every operational lab-unit role holds,
+    // so ordinary Results/Reports/Validation/Reception users don't get denied
+    // (which crashed the report page: the 500 made labUnits a non-array).
+    @PreAuthorize("hasAuthority('PRIV_ORDER_VIEW')")
```
reagan-meant: **[MEDIUM] `PRIV_ORDER_VIEW` gating a role lookup couples two unrelated domains, and the comment shows it was chosen to stop a crash**

The rationale is candid — `PRIV_ORDER_VIEW` was picked as "the privilege every operational lab-unit role holds" to stop a 500 on the report page. But that makes the gate a proxy for "is a normal user", not a statement about the operation, and it wires role-name resolution to the *order* domain. Two concrete consequences:

- **Under-privileging → outage.** Any role that legitimately lacks `order:view` cannot resolve a role by name. Per the `012-004` seed, `User Account Administrator` (only `user:manage`) and `Audit Trail` (only `audit:view`) hold no `order:view`, so `getRoleByName` denies for them. `UserServiceImpl:230` calls `roleService.getRoleByName(Constants.ROLE_GLOBAL_ADMIN).getId()` unguarded — a `AccessDeniedException` there breaks the user-admin screen for the very role that administers users.
- **Over-privileging.** Conversely every Reception/Results/Validation/Reports user can now enumerate role records by name.

The deeper issue is that `RoleServiceImpl.getRoleByName` **never returns null** — on a miss it returns a stub with `id = -1` and the *requested name*. So a denial-or-miss is indistinguishable from a real role downstream, and `UnifiedSystemUserController.initialize()`'s `if (globalAdmin == null)` fallback to `'Admin'` is dead code; `GLOBAL_ADMIN_ID` silently becomes `"-1"`, after which `doRoleFiltering` (`:305`) treats *no one* as global admin and filters the admin role out of every user's assignable list.

**Suggested fix:** Introduce a dedicated low-sensitivity privilege for name→id resolution (or leave this specific lookup ungated with a comment, as `getRoleById` already is) rather than borrowing `PRIV_ORDER_VIEW`. Separately, have `getRoleByName` return `null` on a miss (as @ibacher suggested) and make callers handle it — the stub hides both misses and denials.

src/main/java/org/openelisglobal/login/controller/LoginPageController.java
```diff
@@ -180,7 +214,7 @@ private void setLabunitRolesForExistingUser(HttpServletRequest request, UserSess
             if (principal instanceof UserDetails) {
                 setLabunitRolesForExistingUserFromDB(session);
                 Set<String> roles = new HashSet<>();
-                for (String roleId : userRoleService.getRoleIdsForUser(session.getUserId())) {
+                for (Integer roleId : userRoleService.getRoleIdsForUser(session.getUserId())) {
                     roles.add(roleService.getRoleById(roleId).getName().trim());
```
reagan-meant: **[MEDIUM] Unguarded `.getName()` on `getRoleById` — a stale `system_user_role` row 500s `/session` and locks the user out of the UI**

`getRoleById` returns `null` for a missing row (`RoleDAOImpl.readRole` → `session.get(...)`, which yields null; unlike `getRoleByName` there is no stub fallback). This line dereferences it directly, as do `:275` and `:281`.

The triggering data is created by this PR's own migration. `012-007` deactivates `Results entry` / `Validator` / `Results Admin` / `Inventory mgr`, and `012-006` deletes the corresponding `system_user_role` rows — but only for users the migration matched. Any `system_user_role` row pointing at a role id that was deleted by a site-specific changeset, or orphaned by a partially-applied migration, now yields `null` here.

Because this is `GET /session` — the endpoint the React frontend polls to bootstrap — an NPE means a 500 on the *session* call, so the affected user cannot load the UI at all. There is no partial degradation. Note this sits inside the `SystemInitFlag.enter()` block, so it is not an authorization failure; it is a plain null dereference.

**Suggested fix:** Skip nulls rather than dereferencing:
```java
Role r = roleService.getRoleById(roleId);
if (r != null && r.getName() != null) {
    roles.add(r.getName().trim());
}
```
and apply the same guard at `:275` and `:281`. A `FOREIGN KEY` on `system_user_role.role_id` would prevent the orphan class entirely.

reagan-meant (commented):
Final batch of inline findings.

src/main/resources/liquibase/3.5.x.x/012-004-seed-role-privilege-mappings.xml
```diff
@@ -0,0 +1,305 @@
...
+      Global Administrator has no rows — it holds ALL privileges implicitly,
+      resolved in the privilege service via a Global Admin shortcut.
+    -->
+
+    <changeSet author="openelis" id="012-004-seed-role-privilege-mappings">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="system_role_privilege" schemaName="clinlims"/>
+            <tableExists tableName="system_privilege" schemaName="clinlims"/>
+            <tableExists tableName="system_role" schemaName="clinlims"/>
+            <sqlCheck expectedResult="0">
```
reagan-meant: **[MEDIUM] All-or-nothing precondition: one pre-existing row silently skips the entire role→privilege seed**

`sqlCheck expectedResult="0"` over the *whole* `system_role_privilege` table with `onFail="MARK_RAN"` means: if the table contains even a single row from any source, this changeset is marked as run and **none** of the ~50 mappings are inserted. Liquibase records it as applied, so a re-run will not fix it.

That is not hypothetical on the upgrade path this PR targets. A site that pilots the feature, a partially-applied migration, a restored/merged database, or any site-specific changeset that pre-seeds one mapping all land here. The failure is silent and total: `system_role_privilege` stays sparse, so `resolveAllPrivilegesForRole` returns an empty set for every non-Global-Admin role, and **every non-admin user logs in with zero `PRIV_*` authorities**. With this PR having replaced ~80 controller `hasRole('ADMIN')` guards with service-layer `hasAuthority('PRIV_*')` checks, those users are locked out of essentially the whole application after upgrade, with nothing in the logs pointing at the skipped changeset.

**Suggested fix:** Make each mapping independently idempotent instead of guarding the batch — drop the table-wide precondition and use `INSERT ... WHERE NOT EXISTS (SELECT 1 FROM system_role_privilege WHERE system_role_id = r.id AND privilege_id = p.id)` per row (the pattern `012-006` already uses correctly), plus a `UNIQUE (system_role_id, privilege_id)` constraint. That makes the changeset safe to re-run and correct on partially-seeded databases.

reagan-meant:
### Review summary

**Scope reviewed**

Full diff against the true merge-base `eed1310` (619 files, +10,702/−1,902), reviewed at head `582786b` — not the PR description, which is now out of date in one load-bearing respect (see below). I traced the chain end to end: `system_privilege`/`system_role_privilege` schema and seeds → `PrivilegeServiceImpl` resolution → `CustomUserDetailsService` authority mapping → `MethodSecurityConfig` → `ModuleAuthenticationInterceptor` → controllers → service gates. I also read callers of modified services, the `@CrossDomainService` inventory (45 annotated types), the `012-001`…`012-009` changesets and their orderng in `base.xml`, `SecurityConfig`/`AppConfig` registration, and the interaction with `#3356`'s daemon-context work already on `develop`.

Checks actually run (JDK 21, Maven 3.8.5, offline):
- `PrivilegeServiceImplTest` — **14/14 pass**
- `PrivilegeResolutionTest` — **8/8 pass**
- `ServicePrivilegeCoverageTest` — **1/1 pass**
- `ClassLevelPreAuthorizeSemanticsTest` — **2/2 pass**

Not verified: no full `mvn verify`, no integration/Failsafe suite, no Playwright E2E, and no live database — so the migration findings below are from reading the changesets, not from an executed upgrade. I did not run a DB upgrade against a production-shaped dataset, which is the one thing I'd most want before merge.

**Critical / high-risk findings**

1. **[BLOCKER] `GET /import/provider` and `/import/all` are open to any authenticated user.** `ProviderImportService.importPractitionerList()` has no `@PreAuthorize` and `@CrossDomainService` exempts the whole interface from the coverage scan; `ImportController` has no guard at head. Its sibling `importOrganizationList()` on the *same controller* is correctly gated `PRIV_ORGANIZATION_MANAGE`. This is a **regression** — @pmanko's `0a3a335a1` fixed exactly this (Copilot B4) and that commit is no longer in the branch history.
2. **[BLOCKER] Any authenticated user can install a trusted CA cert into the JVM truststore.** `TruststoreService.addTrustedCert` is ungated + exempt; `ExternalConnectionController` (`hasRole('ADMIN')` removed here) reaches it from `POST /ExternalConnection` with an uploaded file. The truststore write happens *before* the gated persistence call, so an `AccessDeniedException` later still leaves the certificate installed — a path to MITM of outbound TLS integrations.
3. **[HIGH] Analyzer result acceptance writes clinical results with no privilege gate.** `AnalyzerResultsAcceptService.acceptAndPersist` is ungated + exempt, reached from `POST /rest/AnalyzerResults` on an unguarded controller. Inconsistent with `LogbookResultsPersistService`, which does carry `hasAnyAuthority('PRIV_RESULT_ENTER','PRIV_RESULT_MODIFY')`.
4. **[HIGH] The interceptor still fails open, contrary to the PR description.** The body claims the auto-allow gap was closed; it was closed in `acc3b5408` and **reverted in `b2381dddf`**. Since this PR also removes `hasRole('ADMIN')` from ~80 controllers on the stated assumption that the interceptor plus service gates cover them, the service annotations are now the *only* control on those endpoints — which is precisely why findings 1–3 are directly exploitable rather than defence-in-depth gaps.
5. **[HIGH] `SystemInitFlag` on the pooled `@Scheduled` executor disables *all* authorization on those threads.** `SystemAwareSecurityExpressionRoot` short-circuits `hasAuthority`/`hasRole`/`hasPermission`/`isAuthenticated` unconditionally. Blast radius is wider than it appears: `FhirApiWorkFlowServiceImpl.pollForRemoteTasks()` self-invokes `@Async processWorkflow(...)` via a plain `this.` call, so the `@Async` proxy is bypassed and the whole FHIR workflow runs on the flagged thread with every gate disabled. This also duplicates — and conflicts with — `develop`'s `UserContextPropagatingTaskDecorator` from #3356.
6. **[HIGH] `UserService.getUserLabUnitRoles` gated on `PRIV_RESULT_VIEW`.** It returns another user's role/lab-unit assignments and *downgrades* the delegate's `PRIV_USER_ROLE_VIEW`. Reachable with a caller-supplied id via the unguarded `GET /rest/UnifiedSystemUser?ID=<any>`, exposing login name, account lock/disable state, and role assignments. (Password is obfuscated — no credential leak.)

**Other findings**

- **[MEDIUM]** `012-004`'s `sqlCheck expectedResult="0"` over the entire `system_role_privilege` table with `MARK_RAN` is all-or-nothing: one pre-existing row silently skips all ~50 mappings and is recorded as applied. Result on an upgraded DB is every non-admin logging in with zero `PRIV_*` authorities — effectively a full lockout, given the removed controller guards. Per-row `NOT EXISTS` (the pattern `012-006` already uses) plus a `UNIQUE` constraint would fix it.
- **[MEDIUM]** Global Admin is identified by the mutable, admin-editable display name `"Global Administrator"`. Renaming loses all privileges; creating a role with that exact name grants the full catalog with no `system_role_privilege` row. The check is case-sensitive here but `equalsIgnoreCase` in `CustomUserDetailsService`, so the two halves of the admin model can disagree.
- **[MEDIUM]** `getRoleByName` gated on `PRIV_ORDER_VIEW` — chosen (per the comment) to avoid a crash, not because it describes the operation. `User Account Administrator` and `Audit Trail` hold no `order:view` per the seed, yet `UserServiceImpl:230` calls it unguarded. Compounded by `getRoleByName` never returning null (id `-1` stub), which makes misses and denials indistinguishable and renders the `globalAdmin == null` fallbacks dead code.
- **[MEDIUM]** Interceptor returns **401** for authorization failures while `SecurityConfig`'s `AccessDeniedHandler` returns **403** for the same condition. Unauthenticated requests never reach the interceptor, so every 401 it emits is semantically a 403; frontends that treat 401 as "session expired" will spuriously log users out.
- **[MEDIUM]** `LoginPageController:218` (and `:275`, `:281`) dereference `getRoleById(...).getName()` unguarded. `getRoleById` returns null on a miss, and this PR's own `012-006`/`012-007` migration creates exactly the orphan-row conditions that trigger it — on `GET /session`, so the affected user cannot load the UI at all.

**Security areas verified (no issue found)**

- **Interface-level `@PreAuthorize` enforcement.** Copilot's claim that class-level annotations on service *interfaces* aren't enforced does not hold for this Spring config — `ClassLevelPreAuthorizeSemanticsTest` empirically disproves it for both the real-impl and JDK-proxy shapes, and I ran it. Good test to have pinned.
- **Inheritance resolution.** The cycle guard, diamond handling, and Global Admin short-circuit in `resolveAllPrivilegesForRole` are correct. I specifically checked whether the `visited` set shared across roles in `getAllPrivilegesForUser` could drop privileges — it cannot, because the per-role results are unioned. Worth noting base roles ship **flat** (`012-005`), so inheritance is largely unexercised by shipped data.
- **Migration SQL vs. `character(30)` padding.** `system_role.name` is blank-padded `bpchar`, but PostgreSQL ignores trailing blanks in `bpchar = 'literal'` comparisons, so the `012-006`/`012-007` `WHERE name = '...'` predicates match correctly. The Java-side `.trim()` calls are the necessary counterpart and are present.
- **Interceptor thread-safety.** The mutable `path` field was correctly converted to a local and threaded as a parameter — this PR *fixes* a real pre-existing race.
- Changeset ordering in `base.xml` (`012-001`→`012-009`) is correct; `012-006` de-dupe guards on the migration INSERTs are sound; the default `admin` login does receive the Global Administrator role via the pre-existing `2.7.x.x/add_admin_default_roles.xml`; the seeded per-role privilege sets are sensible and follow least privilege.

**Test gaps**

These map to concrete risks above, not coverage for its own sake:
- No test asserts that a `@CrossDomainService`-exempt service is **unreachable from a controller**. That single check would have caught findings 1–3. Since the annotation is a whole-interface kill switch on `ServicePrivilegeCoverageTest` (`visitEnd()` returns early on `isCrossDomain`), nothing currently validates the `callers=` prose — and in the `ProviderImportService` case that prose is factually wrong.
- `ServicePrivilegeCoverageTest.hasAuthorityExpression()` explicitly rejects `hasAnyAuthority(`, which pushes legitimately two-privilege methods into opting out of coverage entirely (that is the only reason `LogbookResultsPersistService` is green). It should accept `hasAnyAuthority`.
- No migration test on a **pre-seeded / partially-migrated** `system_role_privilege` — the `012-004` precondition failure mode is invisible without one.
- No test that a user holding *only* `user:manage` or *only* `audit:view` can complete their workflow; that combination is what the `PRIV_ORDER_VIEW` gate on `getRoleByName` breaks.
- No negative test for a `system_user_role` row referencing a deleted/inactive role (the `/session` NPE).

**Recommendation**

**Significant security issues should be resolved before merge.** The privilege model itself — schema, seeds, resolution with cycle/diamond guards, the coverage scan, and the semantics test — is well built, and the least-privilege seed sets are thoughtful. The exposure comes from the transition: ~80 controller guards were removed on the premise that the interceptor and service gates cover them, but the interceptor fails open at head and `@CrossDomainService` exempts 45 interfaces wholesale, three of which are user-reachable and ungated. Findings 1–3 are the merge blockers; 4 and 5 are architectural decisions worth settling alongside @mozzy11's outstanding point that `DaemonAuthenticationToken` carries only `ROLE_SYSTEM` and so satisfies no `PRIV_*` gate.

*Posted as comments only — no approve/request-changes action taken.*

src/main/java/org/openelisglobal/userrole/service/UserRoleService.java
```diff
@@ -3,26 +3,43 @@
...
-    List<String> getRoleIdsForUser(String userId);
+    /**
+     * Ungated identity read: called while BUILDING the caller's authorities —
+     * during authentication (CustomUserDetailsService) and session introspection
+     * (/session) — so no privilege can be required to discover one's own roles.
+     */
+    List<Integer> getRoleIdsForUser(String userId);
 
+    @PreAuthorize("hasAuthority('PRIV_USER_ROLE_VIEW')")
     boolean userInRole(String userId, String roleName);
```
reagan-meant, line 51: **[HIGH] `userInRole` is newly gated on `PRIV_USER_ROLE_VIEW`, which no seeded base role holds — this 403s `GET /rest/menu` for every non-admin**

`userInRole` was **ungated on `develop`**; this PR adds `hasAuthority('PRIV_USER_ROLE_VIEW')` to both overloads. But parsing `012-004`, none of the 10 seeded base roles is granted `user_role:view` — only Global Administrator holds it, via the `*` sentinel.

The reachable path is on a hot, unavoidable endpoint. `MenuController.getMenuTree` (`GET /rest/menu`, hit on **every page load**) -> `applyDomainFilter` -> `resolveAllowedDomains` -> `isGlobalScopeUser(sysUserId)`, whose **first** statement is `userRoleService.userInRole(sysUserId, Constants.ROLE_GLOBAL_ADMIN)` (`MenuController:155`). For a Reception/Results/Validation/Reports user that call now throws `AccessDeniedException`. There is no try/catch anywhere in `MenuController`, so it propagates to `SecurityConfig`'s `accessDeniedHandler` (`:438`) and returns `{ "status": 403, "message": "Access denied" }` — the navigation tree fails to load for every non-admin user.

Note the user is asking about *their own* roles here, so arguably no privilege should be required at all: this is the same self-identity case the PR already documents as deliberately ungated for `getRoleIdsForUser` ("no privilege can be required to discover one's own roles"). The identical argument applies to `userInRole` when `userId` is the caller.

Two more callers hit the same pattern with the same self-id: `OrderSearchRestController:1548` and `PatientDashBoardProvider:490`, both via their own copy of `isGlobalScopeUser`. `StorageLocationRestController:88` also calls it.

This is adjacent to, but distinct from, the `UserService.getUserLabUnitRoles` mismatch fixed in `1f56bda` — that fix correctly aligned the delegate at `PRIV_USER_ROLE_VIEW`, but on these paths `userInRole` denies *first*, so the privilege is unreachable for the roles that need it either way.

**Suggested fix:** Leave `userInRole` ungated (documented like `getRoleIdsForUser`) for the self-id case, since that is an identity primitive rather than an administrative read. If you prefer to keep the gate, grant `user_role:view` to the operational base roles in `012-004`. Either way, add a test that a user holding only `Results` can load `GET /rest/menu` — the current suite has no such assertion, which is why this is invisible to CI.

reagan-meant:
### Re-review of `1f56bda` (fix commit)

Re-verified each of my 11 earlier findings against the new head, file by file. Both **BLOCKERs are properly fixed**, and 6 of 8 lower-severity items are resolved. Tests re-run: `ServicePrivilegeCoverageTest`, `ClassLevelPreAuthorizeSemanticsTest`, `PrivilegeServiceImplTest`, `PrivilegeResolutionTest` — **25/25 pass**. Notably the coverage test now passes *with two exemptions removed*, so it's a genuinely stronger check rather than a relaxed one.

**Fixed — verified**

| # | Finding | How |
|---|---|---|
| 1 | **[BLOCKER]** `/import/provider` open | `PRIV_PROVIDER_MANAGE` on the service **and** class-level `PRIV_SYSTEM_CONFIGURE` restored on `ImportController`; misleading `callers=` prose corrected |
| 2 | **[BLOCKER]** Truststore CA install | `@CrossDomainService` removed, `PRIV_EXTCONNECTION_MANAGE` added, **and** the truststore write moved *after* the privilege-checked persistence — the ordering half of the finding was caught too |
| 3 | **[HIGH]** Analyzer result accept | `hasAnyAuthority('PRIV_RESULT_ENTER','PRIV_RESULT_MODIFY')`, exemption dropped |
| 6 | **[HIGH]** `getUserLabUnitRoles` mismatch | Realigned to `PRIV_USER_ROLE_VIEW`; `UnifiedSystemUserRestController` GET+POST now gated `PRIV_USER_MANAGE`, with a note on why `/users` stays open |
| — | **[MEDIUM]** `012-004` all-or-nothing precondition | Table-wide `sqlCheck` removed; all **47** INSERTs converted to per-row `NOT EXISTS`, backed by `system_role_privilege_uq`. This was the widest-blast-radius item |
| — | **[MEDIUM]** 401 vs 403 | Now `SC_FORBIDDEN` + matching body, consistent with the `AccessDeniedHandler` |
| — | **[MEDIUM]** `getRoleByName` on `PRIV_ORDER_VIEW` | Now ungated as an identity primitive, with the reasoning recorded |
| — | **[MEDIUM]** `/session` NPE | Null-guarded at all three call sites via `resolveRoleName` |
| — | Coverage-test gap | `hasAuthorityExpression()` now accepts `hasAnyAuthority`, so two-privilege methods no longer have to opt out of the scan |

I also confirmed `provider:manage` and `extconnection:manage` are seeded in `012-003` and correctly held by Global Admin only.

**Still open**

- **[HIGH] #5 — `SystemInitFlag` on the pooled `@Scheduled` executor.** `SchedulerConfig:87` still wires `SystemContextTaskDecorator.systemContext()`, and `SystemInitFlag`/`SystemInitBeanPostProcessor`/`SystemAwareSecurityExpressionRoot` are all still present. This still disables *all* authorization on those threads, and the `pollForRemoteTasks()` -> plain-`this.` -> `@Async processWorkflow()` self-invocation still drags the whole FHIR workflow into that window. This is the piece @ibacher called dispensable and @mozzy11 asked to drop in favour of #3356's `DaemonContextExecutor` — it needs the daemon-vs-`PRIV_*` design decision settled, so I understand it not landing in a fix commit.
- **[HIGH] #4 — interceptor still fails open** on unmapped `/rest` paths. Reasonable as a deliberate choice, but the **PR description still claims the gap was closed**; please correct the body so nobody relies on a control that isn't there.
- **[MEDIUM] Global Admin keyed on the mutable name** `"Global Administrator"` (`PrivilegeServiceImpl:113`), still case-sensitive here vs `equalsIgnoreCase` in `CustomUserDetailsService`.

**One new finding (posted inline)**

**[HIGH] `userInRole` is now gated on `PRIV_USER_ROLE_VIEW`, which no seeded base role holds.** It was ungated on `develop`. `MenuController.isGlobalScopeUser` calls it as its first statement on `GET /rest/menu` — hit on every page load — with no try/catch, so every non-admin gets a 403 and the navigation tree fails to load. Same pattern in `OrderSearchRestController:1548` and `PatientDashBoardProvider:490`. Since these all query the caller's *own* roles, this is the self-identity case already documented as deliberately ungated for `getRoleIdsForUser`. Details and suggested fix in the inline comment.

**Recommendation**

**Address findings before merge** — down from "significant security issues" in my first pass. The exploitable ungated paths are genuinely closed and the migration is now safe to re-run, which were my real blockers. What remains is one new access-denial regression (`userInRole`, likely a small fix), plus two architectural items that are decisions rather than oversights. Caveat unchanged: I have not run a DB upgrade against a production-shaped dataset, and no non-admin E2E run covers `/rest/menu`, which is exactly the gap that let the `userInRole` regression through.

*Comments only — no approve/request-changes action taken.*

PR #3471, opened 2026-04-21
issue-3469: fix(inventory): resolve Hibernate 6 enum/bytea type mismatch in InventoryItemDAOImpl.getByItemType()

### LINKED ISSUE #3469: [Bug] InventoryItemDAOImpl.getByItemType() throws PostgreSQL "operator does not exist: character varying = bytea" — Hibernate 6 enum type mismatch
## Description

Calling `getByItemType()` in `InventoryItemDAOImpl` causes the `GET /api/inventory/items/type/{itemType}` endpoint to fail completely in production with the following error:
```
ERROR: operator does not exist: character varying = bytea
Hint: No operator matches the given name and argument types.
You might need to add explicit type casts.
```

**Why this was not caught:** `getByItemType()` had zero integration tests covering it. The existing `InventoryItemServiceTest` covers `get()`, `getAllActive()`, `insert()`, and `getTotalCurrentStock()` , but this method was never tested. Because the CI pipeline never executed this code path, the bug survived undetected through the Hibernate 6 upgrade.

---

## Affected Component

- **File:** `src/main/java/org/openelisglobal/inventory/daoimpl/InventoryItemDAOImpl.java`
- **Method:** `getByItemType(ItemType itemType)` — line ~58
- **Endpoint affected:** `GET /api/inventory/items/type/{itemType}`

---

## Steps to Reproduce

1. Clone the repository and set up the local development environment
2. Add the following integration test to `InventoryItemServiceTest`:

```java
@Test
public void getByItemType_shouldReturnItemsOfSpecifiedType() {
    List<InventoryItem> reagents = inventoryItemService.getByItemType(ItemType.REAGENT);
    assertNotNull("Result should not be null", reagents);
    assertFalse("Should return at least one reagent", reagents.isEmpty());
}
```

3. Run the test:

```bash
mvn test "-Dtest=org.openelisglobal.inventory.service.InventoryItemServiceTest#getByItemType_shouldReturnItemsOfSpecifiedType"
```

---

## Actual Behavior

The test fails immediately with:

Image

---

## Expected Behavior

`getByItemType(ItemType.REAGENT)` should return a list of active `InventoryItem` objects where `item_type = 'REAGENT'`, with no exception thrown.

---

## Root Cause

The method currently uses the JPA Criteria API and passes a raw Java enum object directly as a query parameter. In **Hibernate 6** the Criteria API no longer automatically resolves `@Enumerated(EnumType.STRING)` annotated fields when the parameter is a raw enum. Instead, it serializes the enum as binary data (`bytea`). PostgreSQL then tries to compare the `VARCHAR` column `item_type` with `bytea`, two completely incompatible types, and throws the error. 
Here is the code producing the producing the bug, you can find it in `InventoryItemDAOImpl`:

Image

This same class of issue was already identified and fixed elsewhere in the codebase by @pmanko in `AnalyzerErrorDAOImpl`. Here is the code and how it was solved, found in `AnalyzerErrorDAOImpl` file on line 159-162

Image
The fix was applied there but was not propagated to the inventory module.

---

## Proposed Fix

The fix involves two parts:

**1. Fix the query:** Replace the Criteria API query in `getByItemType()` with a JPQL query. JPQL correctly respects the `@Enumerated(EnumType.STRING)` annotation on the `itemType` field and resolves the enum to its string name before sending the value to PostgreSQL. This is consistent with how other DAO classes in the codebase handle the same Hibernate 6 type resolution issue.

**2. Add the missing integration test:** Since `getByItemType()` had no integration test, one will be added as part of this fix. The test will call the method with each `ItemType` value and assert that results are returned without exception ensuring this regression cannot silently reappear in future Hibernate or Spring upgrades.

---

## Impact

- `GET /api/inventory/items/type/{itemType}` is broken in production for all item types
- Any code path calling `InventoryItemService.getByItemType()` returns a `LIMSRuntimeException` that hides the real PostgreSQL error from the caller
- The entire item-type filtering feature of the Inventory module is non-functional

**Severity: Medium** — production endpoint broken, no workaround available for end users.

noel-mugisha:
## Issue Number

Closes #3469

## Objective

Fix a production bug where `InventoryItemDAOImpl.getByItemType()` throws a PostgreSQL type mismatch error due to a Hibernate 6 behavioral change introduced with the Java 21 / Spring Boot upgrade. Also adds integration tests for this method, which had zero test coverage.

### 1. `InventoryItemDAOImpl.java` — bug fix

Replaced the JPA Criteria API query in `getByItemType()` with a JPQL query and added the `.name()` method for the `ItemType` enum.

**Why:** In Hibernate 6, the Criteria API no longer automatically resolves `@Enumerated(EnumType.STRING)` annotated fields when a raw Java enum is passed as a parameter to `cb.equal()`. Instead, it serializes the enum as binary data (`bytea`). PostgreSQL cannot compare a `VARCHAR` column with `bytea`

### 2. `InventoryItemServiceTest.java` — new integration tests

Added an integration tests for the updated `getByItemType()` method:
- `getByItemType_shouldReturnItemsOfSpecifiedType()` — verifies correct type filtering and verifies items are returned successfully

## How to Test

```bash
mvn test "-Dtest=org.openelisglobal.inventory.service.InventoryItemServiceTest"
```

**Before this fix:** `getByItemType_shouldReturnItemsOfSpecifiedType` fails with `LIMSRuntimeException: Error getting inventory items by type` caused by `PSQLException: operator does not exist: character varying = bytea`.

**After this fix:** All 8 tests in `InventoryItemServiceTest` pass with no errors whatsoever

## Test Results

image

Only one test is skipped, this was done by the original contributor and it is intentional, so everything works as planned.
<br>
**This is part of my ongoing GSoc contributions to the project of increasing overall integraion tests coverage**

### DIFF
--- a/src/main/java/org/openelisglobal/inventory/daoimpl/InventoryItemDAOImpl.java
+++ b/src/main/java/org/openelisglobal/inventory/daoimpl/InventoryItemDAOImpl.java
@@ -46,14 +46,12 @@ public class InventoryItemDAOImpl extends BaseDAOImpl<InventoryItem, Long> imple
     @Transactional(readOnly = true)
     public List<InventoryItem> getByItemType(ItemType itemType) throws LIMSRuntimeException {
         try {
-            CriteriaBuilder cb = entityManager.getCriteriaBuilder();
-            CriteriaQuery<InventoryItem> cq = cb.createQuery(InventoryItem.class);
-            Root<InventoryItem> root = cq.from(InventoryItem.class);
+            String jpql = "SELECT i FROM InventoryItem i " + "WHERE i.itemType = :itemType AND i.isActive = 'Y' "
+                    + "ORDER BY i.name";
 
-            cq.select(root).where(cb.and(cb.equal(root.get("itemType"), itemType), cb.equal(root.get("isActive"), "Y")))
-                    .orderBy(cb.asc(root.get("name")));
-
-            return entityManager.createQuery(cq).getResultList();
+            return entityManager.createQuery(jpql, InventoryItem.class)
+                    // use .name() to avoid PostgreSQL varchar/bytea type mismatch
+                    .setParameter("itemType", itemType.name()).getResultList();
         } catch (Exception e) {
             throw new LIMSRuntimeException("Error getting inventory items by type", e);
         }
--- a/src/main/java/org/openelisglobal/inventory/daoimpl/InventoryItemDAOImpl.java
+++ b/src/main/java/org/openelisglobal/inventory/daoimpl/InventoryItemDAOImpl.java
@@ -46,11 +46,11 @@ public class InventoryItemDAOImpl extends BaseDAOImpl<InventoryItem, Long> imple
     @Transactional(readOnly = true)
     public List<InventoryItem> getByItemType(ItemType itemType) throws LIMSRuntimeException {
         try {
-            String jpql = "SELECT i FROM InventoryItem i " + "WHERE i.itemType = :itemType AND i.isActive = 'Y' "
+            String jpql = "SELECT i FROM InventoryItem i "
+                    + "WHERE i.itemType = :itemType AND i.isActive = 'Y' "
                     + "ORDER BY i.name";
 
             return entityManager.createQuery(jpql, InventoryItem.class)
-                    // use .name() to avoid PostgreSQL varchar/bytea type mismatch
                     .setParameter("itemType", itemType.name()).getResultList();
         } catch (Exception e) {
             throw new LIMSRuntimeException("Error getting inventory items by type", e);
--- a/src/test/java/org/openelisglobal/inventory/service/InventoryItemServiceTest.java
+++ b/src/test/java/org/openelisglobal/inventory/service/InventoryItemServiceTest.java
@@ -87,4 +87,15 @@ public class InventoryItemServiceTest extends BaseWebContextSensitiveTest {
         assertNotNull("Saved item should be retrievable", savedItem);
         assertEquals("Test Item Created", savedItem.getName());
     }
+
+    @Test
+    public void getByItemType_shouldReturnItemsOfSpecifiedType() {
+        List<InventoryItem> reagents = inventoryItemService.getByItemType(ItemType.REAGENT);
+
+        assertNotNull("Result should not be null", reagents);
+        assertFalse("Should return at least one REAGENT item from dataset", reagents.isEmpty());
+        for (InventoryItem item : reagents) {
+            assertEquals("All returned items should be of type REAGENT", ItemType.REAGENT, item.getItemType());
+        }
+    }
 }


noel-mugisha:
Hey @HerbertYiga @mozzy11, you can review the changes I made when you get time. 
And your input and review is mostly welcomed.

mozzy11 (commented):
Thanks @noel-mugisha — confirmed develop still carries the CriteriaBuilder query from issue #3469 and your JPQL + .name() binding sidesteps the Hibernate 6 enum/bytea mismatch, with a regression test to hold it in place. Build+Test is green; the E2E checkpoint failure looks unrelated to a backend DAO change, so we'll re-run it. Two small asks: move the '// use .name() to avoid...' inline comment into the method Javadoc (we keep method bodies comment-free), and paste the original stack trace from #3469 into the PR description so the fix is self-documenting. Then this is good to merge.

src/main/java/org/openelisglobal/inventory/daoimpl/InventoryItemDAOImpl.java
```diff
@@ -46,14 +46,12 @@ public List<InventoryItem> getAllActive() throws LIMSRuntimeException {
...
-            Root<InventoryItem> root = cq.from(InventoryItem.class);
+            String jpql = "SELECT i FROM InventoryItem i " + "WHERE i.itemType = :itemType AND i.isActive = 'Y' "
+                    + "ORDER BY i.name";
 
-            cq.select(root).where(cb.and(cb.equal(root.get("itemType"), itemType), cb.equal(root.get("isActive"), "Y")))
-                    .orderBy(cb.asc(root.get("name")));
-
-            return entityManager.createQuery(cq).getResultList();
+            return entityManager.createQuery(jpql, InventoryItem.class)
+                    // use .name() to avoid PostgreSQL varchar/bytea type mismatch
```
mozzy11: Per the earlier review ask, please move this explanation into the method's Javadoc (e.g. "Uses JPQL with enum .name() binding because Hibernate 6 Criteria serializes raw enum parameters as bytea, which PostgreSQL cannot compare to varchar") — project convention keeps method bodies comment-free.

mozzy11 (commented):
Re-verified against current develop: `getByItemType()` still uses the CriteriaBuilder query and `InventoryItem.itemType` is `@Enumerated(EnumType.STRING)`, so the Hibernate 6 varchar/bytea mismatch from #3469 is still live and the JPQL + `.name()` binding is the correct fix, held in place by the new integration test. Build + Test is green; the red E2E checkpoint is a stale run predating recent E2E infra fixes and is unrelated to a backend DAO change.

Remaining before approval: the earlier ask to move the `// use .name() ...` comment out of the method body into the method's Javadoc is still outstanding (inline below). One heads-up for maintainers: #3840 replaces the hardcoded `ItemType` enum with an admin-managed item-type table and rewrites `InventoryItemDAOImpl`, which will supersede this query entirely — this small fix is still worth landing first as it is merge-ready and unblocks the module today, but sequencing should be deliberate.

src/main/java/org/openelisglobal/inventory/daoimpl/InventoryItemDAOImpl.java
noel-mugisha: @mozzy11 Done, I think it is ready to be merged

PR #3509, opened 2026-04-27
fix(reports): scope Routine CSV columns to selected lab unit

### LINKED ISSUE #3508: Routine CSV Report includes test columns from other lab units when filtering by Lab Unit
## Summary

When generating a Routine CSV Report and selecting a specific lab unit (e.g., **Hematology**), the exported CSV still contains column headers (and empty cells) for tests that belong to other units (e.g., Albumin, ALT from Chemistry). The unit-type dropdown effectively only filters cell values, not the column set.

## Steps to reproduce

1. Log in as Lab Technician or Admin.
2. Reports → Routine → **Routine CSV Report**.
3. Set start/end dates and select **Hematology** in the unit type dropdown.
4. Click **Generate Printable version** and open the CSV.

## Expected

CSV contains the base demographic columns (`Lab No`, `Lab Unit`, `Identifier`, ..., `Status`) followed only by columns for tests belonging to the selected unit.

mherman22:
Closes #3508

## Summary

The Routine CSV Report's lab-unit dropdown only filtered cell values, not the column set, so a Hematology export still listed every active test in the catalog (Albumin, ALT, Anti-CD*, etc.) as columns. This PR scopes the column set to the selected unit.

## Why it was broken

`CIRoutineColumnBuilder`'s constructor calls `defineAllTestsAndResults()` (which loads `allTests`) and `defineAllReportColumns()` (which iterates `allTests` to add CSV column headers). But `ExportRoutineByDate` was setting `selectedLabUnit` *after* construction via a setter — so the field was always `null` at the moment column structure was being built, and `allTests` ended up holding the entire test catalog.

## What changed

- **`CIRoutineColumnBuilder`**: new 2-arg constructor `(DateRange, String selectedLabUnit)` that assigns the field *before* `defineAll*` runs. The 1-arg constructor is kept for back-compat (used by `TBColumnBuilder`).
- **`RoutineColumnBuilder`**: 2-arg constructor passthrough.
- **`ExportRoutineByDate.getColumnBuilder()`**: passes `selectedLabUnit` via the new constructor; the now-redundant post-construction setter call is removed.
- **`CSVRoutineColumnBuilder`**:
  - `defineAllTestsAndResults()` filters `allTests` by `testSection.id == selectedLabUnit` after loading.
  - `appendResultCrosstab()` adds a matching `AND t.test_section_id = N` to the inner crosstab category SQL. The Java filter and SQL filter must move in lockstep — `allTests` drives the AS-clause column count, and Postgres `crosstab(text, text)` errors out if the inner category SQL doesn't return the same number of rows as columns the AS clause declares.
  - Drops the dead `setSelectedLabUnit(...)` setter (no longer called now that the field is set via constructor).

## Test plan

- [ ] Reports → Routine → **Routine CSV Report** → pick **Hematology** → Generate. Verify only Hematology assays appear after `STATUS`; no Albumin/ALT/Anti-CD* spillover.
- [ ] Repeat with **Chemistry** — only chemistry-section tests appear.
- [ ] Run with **no unit selected** — every active test appears, matching prior behavior.
- [ ] Cells in the data rows align with the right test columns (no Postgres `tuple descriptions are incompatible` error).
- [ ] TB CSV Report (`ExportTBOrdersByDate` → `TBColumnBuilder`) still builds via the 1-arg constructor and isn't regressed.


mozzy11:
Thanks @mherman22

PR #3528, opened 2026-05-01
Make sure label quantities work fine on label printing depending on the configured numbers

### LINKED ISSUE #3531: Label quantities entered on AddSample form are not honored on print
## Bug

Label quantities entered on the AddSample form's `LabelsSection` (Order labels: N, Specimen labels: M) do not propagate through to the actual printed labels. Every print job produces just **one label** regardless of the number entered, on both the per-row Print buttons in the 3-step Order workflow and the post-save print dialog.

## Steps to reproduce

1. Open the AddSample form to create a new order.
2. In the LabelsSection, set **Order labels: 7** and **Specimen labels: 4**.
3. Save the order.
4. On the success screen, click **Print** on the Order label row.
5. Click **Print** on the Specimen label row.

## Expected

- The order print job produces 7 copies of the order label.
- The specimen print job produces 4 copies of the specimen label per sample (independent counts per sample row).

## Actual

- Both print jobs produce **1 label** regardless of the quantities entered.
- For multi-sample orders, the bug is worse: the summed specimen count is interpreted as "every sample item", so when N samples × M specimen labels is requested, the servlet prints `(sum_of_specimens)` per sample item (over-printing by a factor of the sample count).

## Additional gaps surfaced while investigating

- The post-save dialog's **Done** button is wired to an `onDone` callback that no consumer of `PostSavePrintDialog` actually passes — clicking it is a true no-op.
- The success checkmark icon used a hardcoded English `aria-label` and rendered smaller than the original 120 px PNG.
- The `freezer` label type passes servlet validation but produces an empty PDF — `BarcodeLabelMaker.generateLabels` only has a `freezerOrder` branch.
- Several places in the dialog hardcoded English copy (`"Order label"`, `"Specimen label"`) instead of using React Intl.

## Affected screens

- AddOrder success page (`OrderSuccessMessage` → `PostSavePrintDialog`)
- 3-step Order workflow Step 3 (`OrderLabel.jsx` per-row Print)
- Backend `BarcodeWorkflowPrintServiceImpl.buildPostSavePrintDialog`
- `LabelMakerServlet` validation

mherman22:
## Summary

Make the order/specimen label counts entered in the AddSample form's `LabelsSection` (`numOrderLabels` / `numSpecimenLabels`) actually flow through to printed output. Previously every print job came back with **1 label** regardless of input, on both the per-row Print buttons in the Order workflow and the post-save print dialog.

The chain broke in three places:
1. **`OrderSuccessMessage.jsx`** rebuilt each printable label with `quantity: 1` hardcoded, throwing away the backend-computed count.
2. **`BarcodeWorkflowPrintServiceImpl.buildPrintUrl`** never embedded `quantity` in the URL, so even when the dialog displayed the right number, the servlet defaulted to 1.
3. **`addPositiveTypes`** summed every sample row's specimen count into a single `type=specimen` URL with no `sortOrder`. Since `LabelMakerServlet` interprets that as "every sample item", the summed count was multiplied by N samples — over-printing by a factor of the sample count.

This PR makes the AddSample form the single source of truth and tidies the post-save dialog while we're here.

### Backend
- `BarcodeWorkflowPrintServiceImpl.buildPostSavePrintDialog` emits **one printable entry per sample row** instead of summing. Each per-sample entry carries its `sortOrder` via the new `sampleNumber` field on `PrintableLabelOptionForm` and a `labNo.<sortOrder>` print URL.
- `buildPrintUrl` carries the user-selected quantity. **No `override=true` by default** — the persistent `barcode_label_info.num_printed` cap is preserved; the existing servlet override prompt still fires when needed.
- `mapLabelTypeForUrl` normalizes `block`→`blockOrder`, `slide`→`slideOrder`, **and `freezer`→`freezerOrder`** (the bare `freezer` type passed servlet validation but produced an empty PDF since `BarcodeLabelMaker.generateLabels` only has a `freezerOrder` branch).
- `LabelMakerServlet.validate` accepts `freezerOrder`.

### Frontend
- `OrderSuccessMessage` reads each backend label option's `quantity`, `sampleNumber`, and `printUrl`, lets `PostSavePrintDialog` open `printUrl` directly (no duplicate URL building), and URL-encodes every component of the local fallback URL. The success checkmark renders as `CheckmarkFilled` at 120 px, centered above the heading; the icon's `aria-label` is internationalized; Done resets the form and returns to page 0.
- `OrderLabel.jsx` per-row Print URLs (3-step Order workflow) drop the always-on `&override=true` so reprints respect the safety cap.
- `PostSavePrintDialog` uses message-driven label names (`barcode.label.{order|specimen|block|slide|freezer}`), suffixes the `sampleNumber` for per-sample entries (e.g., "Specimen 2"), and only renders Done when the consumer wires `onDone` (no surprise default navigation in pathology / cytology / batch screens). Footer flex layout moved from inline style to SCSS; styles use Carbon semantic tokens (`--cds-text-secondary`, `--cds-support-success`) for theme support.

### Tests
- `BarcodeWorkflowPrintServiceTest` updated for the new URL contract: `&quantity=N`, per-sample `labNo.<sortOrder>` for specimens, `freezerOrder` mapping. Added `buildPostSavePrintDialog_emitsOneSpecimenEntryPerSampleRow` to lock in the no-summing contract so future regressions fail loudly.

## Test plan
- [x] Create an order with 3 samples, set `Order: 7`, `Specimen: 4` on the AddSample form
- [x] Verify the post-save dialog shows `Order label · Qty: 7` and `Specimen 1 / 2 / 3 · Qty: 4` (three separate rows, not summed)
- [x] Click each Print button → verify the resulting PDF has the right number of copies for that specific specimen
- [x] Click Done → verify the form resets and returns to page 0
- [x] Repeat with a freezer-typed sample → verify the URL uses `type=freezerOrder` and a non-empty PDF is produced
- [x] Repeat with `block` / `slide` (pathology) → verify URLs map to `blockOrder` / `slideOrder`
- [x] Reprint the same label many times in a row → verify the existing "max reached / Override?" prompt still fires (safety cap preserved)
- [x] `mvn test -Dtest=BarcodeWorkflowPrintServiceTest` passes
- [x] `cd frontend && npm test -- PostSavePrintDialog` passes


pmanko:
**Pre-merge polish landed as `1bd6e76a3`** — derived from a tier-1+2 audit + security review + cross-component pass on `55dd6c23b`.

**R1 — DoS hardening (HIGH).** Without an upper bound on `quantity`, an authenticated caller can submit `?override=true&quantity=999999999` and hang the JVM in `BarcodeLabelMaker`'s render loop — the per-label cap (`Label.getMaxNumLabels`, default 10) is bypassed by `override=true`. Added `MAX_REQUEST_LABEL_QUANTITY` config property (`numMaxRequestLabelQuantity`, default 100) read via `BarcodeConfigUtil.parseIntSafe`, matching the existing `MAX_*_LABEL_PRINTED` pattern.

**R2 — Carbon prop fix.** `<StructuredListWrapper condensed flush>` → `isCondensed isFlush`. The bare prop names are silently dropped by `@carbon/react ^1.15.0`; every other call site in the repo uses the `is*` prefix.

**R3 — Fallback URL specimen-aware.** `buildFallbackPrintUrl` in `OrderSuccessMessage.jsx` now threads `sampleNumber` so a specimen entry whose backend `printUrl` is missing produces `labNo.<sortOrder>` instead of bare `labNo`. The latter would re-create the original "multiply by N" overprinting bug on the fallback path.

**R4 — Done button moved out of dialog.** Only 1 of 7 consumers (`OrderSuccessMessage`) ever wired `onDone`; the conditional render in the dialog was hiding design rot rather than fixing it. Removed the `onDone` prop entirely; the Done button now lives in `OrderSuccessMessage`'s `orderEntrySuccessActions` row alongside "Place Another Order", where the reset-and-go-to-page-0 semantic belongs. The other 6 consumers (case views, batch entry, reprint screens) didn't pass `onDone` and have no "done" semantic.

**R4b — Popup-block toast.** `PostSavePrintDialog` gained an `onPopupBlocked(label)` callback. `OrderSuccessMessage` wires it to `addNotification` via `NotificationContext`, mirroring `OrderLabel.openPrintWindow`. Without this, a blocked popup logged to console only.

**R5 — Playwright coverage.** New foundational spec at `frontend/playwright/tests/foundational/core/post-save-print-quantity.spec.ts` locks the cap at the wire level using a real in-progress accession via the established `ORDERS_IN_PROGRESS` seed-lookup pattern (matches `storage-assign-order-label.spec.ts`). The OGC-284 demo spec gains a "FR-011a — Per-sample dialog rows" scene asserting `Order label` is visible after the save success panel renders.

**Out of scope, will file as separate issues:**
- Cytology + IHC fallback in `buildFallbackPostSavePrintDialog` uses bare `type=specimen` without `.sortOrder` — same multiplication pattern, pre-existing.
- `LabelMakerServlet.programCode` accepted without validation, latent.
- Frontend-wide sweep for `window.open` followed by unconditional success-state mutation — ibacher's `openPrintWindow` is the model.

**Local verification:** `mvn test -Dtest=LabelMakerServletTest,BarcodeLabelMakerTest,BarcodeWorkflowPrintServiceTest` ✅ · 21 vitest tests across `PostSavePrintDialog`/`OrderSuccessMessage`/`OrderLabel` ✅ · `npx playwright test --list --project=core-app` confirms new spec is registered · `mvn spotless:apply` + `npm run format` clean.

PR #3555, opened 2026-05-05
Add Diagnostic report facade to openelis

Isabirye1515:
This change introduces the transformation logic that converts a FHIR DiagnosticReport into an internal ResultsUpdateDataSet, enabling OpenELIS to process and persist incoming FHIR lab results using the existing result workflow.

What This Method Does
Initializes a ResultsUpdateDataSet using the current system user
Resolves an existing Analysis using basedOn (FHIR reference → fhirUuid)
Falls back to creating a new Analysis when no match is found
FHIR Mapping Logic
DiagnosticReport.result → TestResultItem
Resolves each referenced Result
Converts Result → Observation → TestResultItem
Adds items to analysisOnlyChangeResults
DiagnosticReport.specimen → SampleItem
Extracts specimen reference
Links Analysis to corresponding SampleItem
DiagnosticReport.code → Test
Resolves tests from CodeableConcept
Assigns first matched Test to Analysis
DiagnosticReport.status → Analysis.statusId
FINAL → Finalized
PRELIMINARY → TechnicalAcceptance
PARTIAL → TechnicalRejected
REGISTERED → NotStarted
Behavior Notes
Safely handles missing or partial FHIR fields
Supports both:
Updating existing Analysis (via basedOn)
Creating new Analysis when none exists


mozzy11 (commented):
Thanks for pushing the DiagnosticReport facade forward! Two parts of this are clear wins: generalizing resolveTestsFromServiceRequest into resolveTestsFromCodeableConcept (nice reuse for both ServiceRequest and DiagnosticReport codes, and you updated the call-site correctly), and the soft-delete approach of cancelling the Analysis rather than hard-deleting.

The main issue is that the PR's headline feature isn't wired up:

- **createResultUpdateDataSetFromReport is never called.** It's on the interface and impl, but no @Create/@Update endpoint in DiagnosticReportProvider uses it, so it ships as dead code. Either add the consuming endpoint (with tests) in this PR, or pull the method out and land it together with the endpoint later. If you keep it: the new Analysis() fallback goes into getModifiedAnalysis() with no test/sampleItem/status set, which the update path can't persist safely — new analyses need their own path; the Result→Observation→TestResultItem roundtrip can be short-circuited; and guard UUID.fromString so a bad reference is a 400/422, not a 500.
- **deleteDiagnosticReport:** please inject IStatusService instead of SpringContext.getBean(...), and wrap UUID.fromString so an invalid id returns 400. Also add a short comment documenting that deleting a report cancels the underlying analysis — that's a significant workflow side effect.
- **Tests:** the only test change is a blank line — the delete operation needs at least happy-path + missing-id coverage (your DeviceFacadeTest on #3560 is a great template).
- The expanded @Search declares eight params but still forwards everything to the FHIR store — fine as an interim step, but add a comment saying so, or trim the unused params.

Scope-wise this will be a much stronger PR as 'soft delete + CodeableConcept refactor, fully tested', with the inbound-results transformation as a follow-up. Happy to review the revision quickly.

mozzy11 (commented):
Hi @Isabirye1515 — Thanks — the refactor of resolveTestsFromServiceRequest to resolveTestsFromCodeableConcept is a genuinely good generalization, and createResultUpdateDataSetFromReport puts the DiagnosticReport->Analysis mapping in the transform service where it belongs. Two duplication flags on the provider side:

- **deleteDiagnosticReport is a near-verbatim copy of ServiceRequestProvider's delete**: same fhirUuid lookup, same Canceled soft-delete, same 'FHIR sync failed during delete (non-blocking)' log string, same three-branch catch chain, plus a third private copy of `safeMessage()` (ServiceRequestProvider and the LocationProvider in #3563 have the others). It also hand-builds the 204 MethodOutcome instead of using FhirProviderUtils.buildDeleteOutcome like Practitioner/Specimen do. At minimum, `safeMessage` should move into FhirProviderUtils; ideally the whole delete template becomes shared.
- **The expanded search still forwards**: searchDiagnosticReports now declares patient/subject/issued/code/result/_id/_lastUpdated/_sort/_include, but the body is unchanged — `util.forwardSearchToFhirStore(request)` — so the new params are decorative. Project direction (#3649) is for facade searches to query the local DB, and that shared search framework is being consolidated between #3720/#3874. Growing the param list on a forwarding stub means we'll rewrite this method twice.

Concrete next step: keep the transform-service changes (they're the durable part of this PR), move safeMessage into FhirProviderUtils, and rebase the delete + search onto the shared base provider / search framework rather than merging another copy of the stub pattern — the param declarations you've written here are exactly what the local-search version will want, so they won't be wasted.

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceImpl.java
```diff
@@ -1863,6 +1863,71 @@ private void includeDeviceIfNeeded(Analysis analysis, FhirOperations fhirOperati
         }
     }
 
+    @Override
+    public ResultsUpdateDataSet createResultUpdateDataSetFromReport(DiagnosticReport report, String sysuserId) {
```
mozzy11, line 1867: This method has no caller in the PR (the DiagnosticReport provider never invokes it) and no test — none of the mapping behavior in the PR body is exercised. Please wire it into the provider (e.g. a `@Create`/`@Update` handler) with facade tests, or defer it to the PR that uses it. Also: the sibling read-only transforms carry `@Transactional(readOnly = true)` — this method does several repository lookups and should too, so all data is compiled within one transaction.

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceImpl.java
```diff
@@ -1863,6 +1863,71 @@ private void includeDeviceIfNeeded(Analysis analysis, FhirOperations fhirOperati
...
+
+        ResultsUpdateDataSet dataSet = new ResultsUpdateDataSet(sysuserId);
+        Analysis analysis = null;
+
+        if (report.hasBasedOn() && report.getBasedOnFirstRep().hasReference()) {
+            String analysisUUID = report.getBasedOnFirstRep().getReferenceElement().getIdPart();
+            List<Analysis> matches = analysisService.getAllMatching("fhirUuid", UUID.fromString(analysisUUID));
```
mozzy11, line 1874: `UUID.fromString(analysisUUID)` throws `IllegalArgumentException` on a malformed reference id → surfaces as a 500. The class already has `getItemByFhirId(...)` which handles both invalid-UUID and not-found — please reuse it here.

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceImpl.java
```diff
@@ -1863,6 +1863,71 @@ private void includeDeviceIfNeeded(Analysis analysis, FhirOperations fhirOperati
...
+        if (analysis == null) {
+            analysis = new Analysis();
+        }
+
+        if (report.hasResult()) {
+            List<TestResultItem> resultItems = new ArrayList<>();
+            for (Reference resultRef : report.getResult()) {
+                String resultId = resultRef.getReferenceElement().getIdPart();
+                Result result = getItemByFhirId(resultId, resultService);
+                Observation observation = transformResultToObservation(result);
```
mozzy11, line 1889: `getItemByFhirId` returns `null` for an unresolvable reference, so `transformResultToObservation(result)` NPEs on an unknown `DiagnosticReport.result` reference. The Result→Observation→TestResultItem round-trip also re-resolves LOINC and re-fetches the analysis — wasteful, and can throw `UnprocessableEntityException` for data already valid in the DB. Consider building the `TestResultItem` directly from the `Result`.

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceImpl.java
```diff
@@ -1863,6 +1863,71 @@ private void includeDeviceIfNeeded(Analysis analysis, FhirOperations fhirOperati
...
+        if (report.hasBasedOn() && report.getBasedOnFirstRep().hasReference()) {
+            String analysisUUID = report.getBasedOnFirstRep().getReferenceElement().getIdPart();
+            List<Analysis> matches = analysisService.getAllMatching("fhirUuid", UUID.fromString(analysisUUID));
+            if (matches != null && !matches.isEmpty()) {
+                analysis = matches.get(0);
+            }
+        }
+
+        if (analysis == null) {
+            analysis = new Analysis();
```
mozzy11, line 1881: The fallback creates an empty transient `Analysis` and unconditionally adds it to `modifiedAnalysis` — with no sample item, test, sample-type, or revision data, persisting that dataset fails or writes an incomplete row. Reject with `UnprocessableEntityException` when nothing matches `basedOn`, or fully populate the new Analysis. Relatedly, `PARTIAL → TechnicalRejected` looks wrong — a partial report isn't a rejection.

src/main/java/org/openelisglobal/fhir/providers/DiagnosticReportProvider.java
```diff
@@ -89,13 +99,73 @@ public DiagnosticReport readDiagnosticReport(@IdParam IdType theId) {
...
+
+        try {
+            if (theId == null || theId.getIdPart() == null) {
+                throw new InvalidRequestException("Missing DiagnosticReport ID in URL");
+            }
+
+            String sysUserId = FhirProviderUtils.getSysUserId(request);
+            String analysisUuid = theId.getIdPart();
+
+            List<Analysis> existingAnalyses = analysisService.getAllMatching("fhirUuid", UUID.fromString(analysisUuid));
```
mozzy11, line 114: `UUID.fromString(analysisUuid)` on a non-UUID id throws `IllegalArgumentException`, caught by the generic `catch (Exception)` and returned as 500 — a malformed client id should be 400 (`InvalidRequestException`). Also prefer injecting `IStatusService` via `@Autowired` over `SpringContext.getBean(...)`, consistent with this class's other dependencies.

src/test/java/org/openelisglobal/fhir/DiagnosticReportFacadeTest.java
```diff
@@ -114,4 +114,5 @@ public void searchDiagnosticReport_endpointExists_shouldNotReturn404() throws Ex
         assertNotNull(response);
         org.junit.Assert.assertTrue(response.getStatus() != 404);
     }
+
```
mozzy11, line 117: The only change here is a blank line, yet the PR adds delete, new search params, and a whole report→dataset transform. Please add: delete cancels the analysis (assert status), delete of unknown/malformed id → 404/400, and coverage for `createResultUpdateDataSetFromReport` once wired in.

mozzy11 (changes requested):
Thanks @Isabirye1515 — the `resolveTestsFromServiceRequest` → `resolveTestsFromCodeableConcept` refactor is clean and correctly propagated, and the added `@Delete` + expanded search params on `DiagnosticReportProvider` are useful.

**Blocking:** `createResultUpdateDataSetFromReport` — the headline feature in the PR description — has **no caller anywhere** (it appears only as interface declaration + implementation; nothing invokes it) and no test (the only test-file change is a blank line). All of its claimed behavior is unverified dead code, and it contains several latent bugs. Either wire it into the provider with facade tests, or split it out of this PR. Details inline.

mozzy11 (commented):
**Quick re-review (summary).** DiagnosticReport facade: soft-delete (cancel Analysis), expanded search params, `resolveTestsFromCodeableConcept` refactor, and a report→`ResultsUpdateDataSet` transform.

- The headline `createResultUpdateDataSetFromReport` transform is **dead code** — no `@Create`/`@Update` caller and no test; its empty `new Analysis()` fallback won't persist; `PARTIAL→TechnicalRejected` mapping is wrong; missing `@Transactional(readOnly=true)`.
- Raw `UUID.fromString(...)` in the transform and in `deleteDiagnosticReport` → 500 on a bad id (reuse `getItemByFhirId`); a null lookup then NPEs.
- Delete duplicates `ServiceRequestProvider`'s delete (3rd copy of `safeMessage`, hand-built 204 vs `FhirProviderUtils.buildDeleteOutcome`); 8 new search params but the body still just forwards to the FHIR store.
- Tests: the only change is a blank line — no delete/happy-path coverage. Prior CHANGES_REQUESTED not addressed.
- **Status:** DIRTY. Next: split to "soft-delete + CodeableConcept refactor, fully tested"; defer the untested transform.

mozzy11 (approved):

- Dropped the uncalled `createResultUpdateDataSetFromReport` transform. A DiagnosticReport is a projection of Analysis + Results, so writes flow through ServiceRequest and Observation; the provider Javadoc now says so.
- Delete validates the id (400 on a non-UUID), injects `IStatusService`, uses the shared delete outcome, and `safeMessage` moved to `FhirProviderUtils` (ServiceRequestProvider's copy removed).
- Fixture panels now carry `lastupdated`; Panel is version-mapped on it, so the null version made every analysis merge fail as transient.
- Facade tests cover cancel-on-delete, 404 and 400. DiagnosticReport, Observation and AnalyzerResultsAcceptHold suites pass locally.

PR #3560, opened 2026-05-06
Adding device provider

Isabirye1515:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


mozzy11 (commented):
Thanks for this — exposing Analyzers as FHIR Devices is a genuinely useful addition, and the transform work is thorough: status mapping in both directions, well-namespaced extensions for communication mode/protocol/transport, and a DeviceFacadeTest that checks real DB state after create/update/delete. The read path's 400-vs-404 handling is exactly right.

A few things to fix before merge:

- **Restore methodService's @Autowired:** in FhirTransformServiceImpl, the new analyzerTypeService field was inserted between an existing @Autowired and private MethodService methodService;, so methodService lost its annotation. It's unused today so nothing crashes, but please put the annotation back.
- **createDevice NPE:** analyzer.getAnalyzerType().setSysUserId(...) — analyzerType is only set when Device.type matches an existing AnalyzerType by name, so any unknown type text throws an NPE (which becomes a 500). Null-check it, and decide whether an unknown type should be a 422 or create the type.
- **updateDevice ignores the URL id:** it transforms only the request body, so a body without an id quietly creates a new Analyzer instead of updating theId. Please resolve the analyzer by theId and reject mismatched body ids.
- **deleteDevice returns 500 for missing devices:** ResourceNotFoundException isn't rethrown in the catch chain, so a 404 becomes a 500. Also, the missing-ID checks throw IllegalArgumentException (HAPI maps that to 500) — use InvalidRequestException for a 400.
- **Copy-paste sweep:** the search method is named searchPractitionerBundle, several error messages say 'Practitioner', update's method variable says "createDevice", and delete mentions 'ServiceRequest UUID'. A quick find-and-replace pass will fix all of these.
- The search declares params but forwards to the remote FHIR store while reads are local — same note as on #3563: filter locally or make the delegation explicit.

Also heads-up: this branch is from mid-May, and the analyzer domain has moved on develop since (plus your own #3595 relocates this transform), so please rebase and decide the merge order between the two. The bones are good — after this cleanup round it'll be in nice shape.

mozzy11 (commented):
Hi @Isabirye1515 — Nice work getting analyzers exposed as FHIR Devices — the transformAnalyzerToDevice/transformDeviceToAnalyzer pair in FhirTransformServiceImpl is genuinely new capability and the facade tests (DeviceFacadeTest) are a good template. My concern is that DeviceProvider.java itself is ~320 lines of which maybe 60 are Device-specific; the rest is a re-type of PractitionerProvider's skeleton, and the copy shows:

- **Copied revIncludes**: searchDeviceBundle allows `Encounter:participant` and `ServiceRequest:requester` — those come straight from PractitionerProvider's search and don't apply to Device. (Same failure mode that left develop's SpecimenProvider throwing 'Error searching Practitioners' — copy-paste always leaks.)
- **Search stub copy #9**: the body ignores every declared param and forwards to the remote store, with yet another null-Bundle policy (here: throw 500). Per #3649 these facades should search the local DB; that shared search is being consolidated between #3720/#3874, so let's not merge another forwarding copy.
- **Exception-mapping drift**: readDevice/deleteDevice correctly map a bad UUID to 400, but createDevice/updateDevice let the IllegalArgumentException from transformDeviceToAnalyzer fall into `catch (Exception)` -> 500 for a client error; updateDevice skips UUID validation entirely. Also the inline fhirUuid lookup duplicates the generic getItemByFhirId that already exists in FhirTransformServiceImpl (and a third copy is being added in #3563).

Concrete next step: keep the transforms and tests, but let's not merge a fourth copy of the CRUD skeleton — either extract the read/create/update/delete templates into a shared BaseResourceProvider (validate-UUID, fhirUuid lookup with 404/duplicate-500, sysUserId stamping, syncToFhirStore, one exception chain) as part of this PR, or park the provider until that base + the search framework land and this file becomes just param declarations + transform delegation + AnalyzerService wiring. Also drop the Practitioner revIncludes either way.

src/main/java/org/openelisglobal/fhir/providers/DeviceProvider.java
```diff
@@ -0,0 +1,322 @@
...
+
+            LogEvent.logError(getClass().getSimpleName(), method, e.getMessage());
+
+            throw new InternalErrorException("Unexpected error creating Device", e);
+        }
+
+    }
+
+    @Update
+    public MethodOutcome updateDevice(@IdParam IdType theId, @ResourceParam Device device, HttpServletRequest request) {
```
mozzy11, line 169: **Blocking:** `theId` is checked for presence but never used — the target analyzer is resolved purely from `device.getId()` inside `transformDeviceToAnalyzer`. So (1) `PUT /Device/{uuid}` with a body missing `id` builds a brand-new transient Analyzer and passes it to `analyzerService.update(...)`; (2) a body id differing from the URL id updates a *different* resource than addressed. Please resolve the analyzer by `theId`, 404 when absent, and reject body/URL id mismatches with `InvalidRequestException`.

src/main/java/org/openelisglobal/fhir/providers/DeviceProvider.java
```diff
@@ -0,0 +1,322 @@
...
+            }
+
+            if (analyzers.size() > 1) {
+
+                throw new InternalErrorException("Multiple Analyzer records found");
+            }
+
+            Analyzer analyzer = analyzers.get(0);
+
+            analyzer.setActive(false);
```
mozzy11, line 259: The soft delete sets `active=false` but leaves `analyzer.status` unchanged, so the Device re-synced to the FHIR store just below still reports `status: active` — external consumers see a "deleted" device as active. Set the analyzer status to inactive/deleted here so the FHIR representation reflects the deletion.

src/main/java/org/openelisglobal/fhir/providers/DeviceProvider.java
```diff
@@ -0,0 +1,322 @@
...
+
+        String method = "createDevice";
+
+        try {
+
+            if (device == null) {
+                throw new InvalidRequestException("Device resource is required");
+            }
+
+            Analyzer analyzer = fhirTransformService.transformDeviceToAnalyzer(device);
```
mozzy11, line 125: If a client POSTs a Device whose `id` matches an existing analyzer's fhirUuid, `transformDeviceToAnalyzer` returns that existing entity and `save(...)` mutates it — `@Create` silently becomes an update. Per FHIR, create should ignore/reject a client-supplied id: strip it in the create path or reject when a matching analyzer exists.

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceImpl.java
```diff
@@ -2062,38 +2070,438 @@ public DiagnosticReport transformResultToDiagnosticReport(Analysis analysis) {
...
+        if (communicationExtension != null && communicationExtension.getValue() instanceof CodeableConcept) {
+
+            CodeableConcept concept = (CodeableConcept) communicationExtension.getValue();
+
+            if (concept.hasCoding()) {
+
+                String code = concept.getCodingFirstRep().getCode();
+
+                analyzer.setCommunicationMode(CommunicationMode.fromValue(code));
+            }
```
mozzy11, line 2185: `CommunicationMode.fromValue(code)` returns `null` for unrecognized codes, so an update with a bogus code silently nulls the analyzer's existing communication mode (same for `ProtocolVersion.fromValue`). These come from external clients — prefer `UnprocessableEntityException` on unknown codes, or at least skip assignment when the lookup returns null.

mozzy11 (changes requested):
Thanks @Isabirye1515 — the Device↔Analyzer mapping is thorough (status mapping, extensions for comm mode/protocol/transport/test units) and the facade tests are genuine integration tests over read/create/update/delete.

**Blocking:** `updateDevice` validates `@IdParam theId` for presence but never *uses* it — resolution relies entirely on the id inside the request body, violating FHIR update semantics (a body that omits `id` creates a new transient Analyzer; a mismatched body id updates the wrong resource). Details inline, plus a soft-delete status issue and a create-as-update hole.

mozzy11 (commented):
**Quick re-review (summary).** FHIR `DeviceProvider` (CRUD + search) mapping Analyzer ↔ Device, with transforms and a `DeviceFacadeTest`.

- **Blocking (CHANGES_REQUESTED, unaddressed):** `updateDevice` validates `@IdParam theId` but never uses it — it resolves only from the body `id`, so a body missing `id` creates a new Analyzer and a mismatched id updates the wrong resource.
- `@Create` acts as update when the client sends an existing `fhirUuid`; unknown `CommunicationMode`/`ProtocolVersion` codes null out existing values (should be 422); `createDevice` NPEs on an unknown type; delete leaves `analyzer.status=active` so a "deleted" device re-syncs as active.
- ~320 lines largely copy-pasted from `PractitionerProvider` (carries over `Encounter`/`ServiceRequest` revIncludes, `searchPractitionerBundle` naming, "Practitioner" error strings) — extract a shared base rather than a 4th copy.
- **Status:** DIRTY (conflicts). Next: rebase, fix the `theId` update semantics, dedupe onto a shared base, re-request review.

mozzy11 (approved):

- PUT resolves the analyzer from the URL id (404 when absent) and rejects a differing body id with 400; a PUT without a body id is already refused by HAPI before the provider runs.
- POST strips a client-supplied id so it can never mutate an existing analyzer.
- DELETE sets the analyzer status to INACTIVE as well as active=false, so the mirrored Device reads back as inactive.
- Unknown communication-mode or protocol-version codes are a 422 instead of silently nulling the analyzer's settings.
- DeviceFacadeTest covers each of these (12 tests pass locally, plus AnalyzerServiceTest).

PR #3563, opened 2026-05-06
Adding Storage location provider 

Isabirye1515:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


Isabirye1515:
thanks @pmanko

pmanko (dismissed):
REQUEST_CHANGES — two correctness defects and a service-layer bypass the green tests don't actually cover.

1. **Bypasses `StorageLocationService`.** `LocationProvider` injects the five raw `Storage*DAO`s and calls `insert`/`update` directly, with no transaction. `StorageLocationServiceImpl` already does this exact 5-type dispatch behind `@Transactional` (`insert(Object)`/`update(Object)`), and every sibling provider (Specimen, Organization, Practitioner) injects a `*Service`, never a DAO. Inject `StorageLocationService` and route writes through it — that's the transaction boundary, and it removes most of the duplication below.
2. **NPE in the room create/update path.** In `createStorageRoomFromLocation`, when `location.hasId()` but `getItemByFhirId` returns null, `room` is null and `room.setName(...)` NPEs. The other four builders guard this (`if (x == null) { x = new …; setFhirUuid(uuid); }`); room doesn't. Add the same guard.
3. **`searchLocationBundle` is a stub.** It declares seven `@OptionalParam`s and uses none — every search forwards to the remote FHIR store, while read/create/update/delete hit local DAOs, so results can diverge. Either filter locally, or drop the params and make the delegation explicit.
4. The 5-way type switch is copy-pasted across persist/get/delete (~half the file) — collapse to one type-keyed handler; most of it falls away once writes go through the service.
5. Minor: both the room and device builders pass `"shelf"` as the code-gen context — should be `"room"`/`"device"`.

mozzy11 (commented):
Thanks for coming back to this after the review — the follow-up commits genuinely improved it. Writes now go through StorageLocationService and the new @Transactional per-type services instead of raw DAOs, the room-builder NPE is guarded, and the duplicate setLabel calls are gone. That's most of the requested changes done, and the LocationFacadeTest coverage across all five hierarchy levels remains a highlight.

Two items from the original review are still open:

- **Device code-gen context:** in StorageLocationFhirTransform.createOrUpdateStorageDeviceFromLocation, the call is still generateCodeFromName(location.getName(), "shelf") — it should be "device" (the room path is fixed, this one was missed).
- **The search stub:** searchLocationBundle still declares eight @OptionalParams and ignores all of them, forwarding to the remote FHIR store. Since reads/creates/updates now hit the local database, a search can return different data than a read of the same Location. Either implement local filtering against the storage services (even just identifier + name to start), or drop the unused parameters and add a short comment stating that search intentionally delegates to the FHIR store for now.

Once those two are in, please re-request review from @pmanko since his CHANGES_REQUESTED is the blocking state. This is close — nice persistence on a big PR.

Isabirye1515:
@pmanko 
The existing search methods (searchRooms, searchDevices, searchShelves, searchRacks, etc.) return Map<String, Object> objects that are specifically designed for the web API/UI layer, not for FHIR resources. These maps contain flattened, presentation-oriented data such as hierarchical paths, parent names, display labels, and occupancy counts rather than the complete domain entities. FHIR Location resources, however, are created from the underlying domain objects (StorageRoom, StorageDevice, StorageShelf, StorageRack, StorageBox) using the transformToFhirLocation(...) methods. Since the entity instances are not available in the mapped results, using the existing search methods would require reconstructing entities from the maps or performing additional database lookups, which introduces unnecessary complexity, duplicated logic, and extra database access. Therefore, the FHIR search implementation should query and return the domain entities directly, allowing them to be transformed into FHIR Location resources efficiently and consistently.

@mozzy11 , lets use this as we build the complehensive search with hibernate search please

mozzy11 (commented):
Hi @Isabirye1515 — Thanks for pushing the storage hierarchy over FHIR forward — the mCSD Location mapping and the DAO->Service migration in StorageLocationRestController are solid groundwork. Before this merges I'd like us to deal with the duplication, because this PR is currently the largest single source of it:

- **5x repeated type blocks**: persistRoom/persistDevice/persistShelf/persistRack/persistBox in LocationProvider.java are ~25 identical lines each, modulo the type; handleGetLocation and handleDeleteLocation then repeat the lookup/soft-delete/transform block five more times each. That's roughly 260 lines that could be one generic handler (~25 lines) resolved from the physicalType category — especially since you already wrote the generic `getItemByFhirId(UUID, BaseObjectService)` in StorageLocationFhirTransform; let's use it everywhere instead of alongside the inline copies.
- **Search stub copy #8**: searchLocationBundle declares 8 params, then ignores them and calls `util.forwardSearchToFhirStore(request)` — the same stub now exists in 8 providers, each with a slightly different null-Bundle policy. Per #3649 the facade direction is local-DB search, and that framework is being consolidated between #3720/#3874, so please don't add another forwarding copy; keep the param declarations and let the shared search own the body.
- **Small repeats**: `safeMessage()` is a copy of ServiceRequestProvider's private helper (third copy incoming in #3555) — it belongs in FhirProviderUtils. Also note readLocation maps a missing ID to 404 where every other provider maps it to 400, and `UUID.fromString` in the transform's create* methods can escape as IllegalArgumentException -> 500; a shared base would give us one correct mapping.

Concrete next step: extract the 5-type fan-out behind a small StorageLocationHandler interface now (that part is Location-specific and worth doing regardless), and hold the CRUD/search skeletons until the shared BaseResourceProvider + search consolidation lands so this provider shrinks to params + transforms + wiring. Happy to pair on the handler extraction if useful!

src/main/java/org/openelisglobal/fhir/providers/LocationProvider.java
```diff
@@ -0,0 +1,588 @@
...
+
+            @OptionalParam(name = "_tag") TokenAndListParam tag,
+
+            HttpServletRequest request) {
+
+        final String methodName = "searchLocationBundle";
+
+        try {
+
+            Bundle bundle = util.forwardSearchToFhirStore(request);
```
mozzy11, line 206: Search still forwards to the remote FHIR store while read/create/update/delete hit the local DB — a search can return different data than a read of the same Location, and all eight declared `@OptionalParam`s are ignored. Either implement local filtering (even just identifier + name to start), or drop the unused params and comment that search intentionally delegates for now. Given #3649 is consolidating local search (#3720/#3874), let's not add another forwarding stub.

src/main/java/org/openelisglobal/fhir/providers/LocationProvider.java
```diff
@@ -0,0 +1,588 @@
...
+        String text = location.getPhysicalType().getText();
+
+        if (text == null || text.isBlank()) {
+            throw new InvalidRequestException("physicalType.text is required and cannot be blank");
+        }
+
+        return text.trim();
+    }
+
+    private Location handleGetLocation(String uuidStr) {
```
mozzy11: `handleGetLocation` repeats an identical lookup-then-transform block five times (and `handleDeleteLocation` five more); `transform.getItemByFhirId(uuid, service)` is already generic — iterating a list of the five services collapses ~120 lines to one loop.

src/main/java/org/openelisglobal/fhir/providers/LocationProvider.java
```diff
@@ -0,0 +1,588 @@
...
+            if (result == null) {
+                throw new InternalErrorException("Failed to transform deleted StorageBox to FHIR Location");
+            }
+            return result;
+        }
+
+        throw new ResourceNotFoundException("No storage location found for UUID: " + uuidStr);
+    }
+
+    private String safeMessage(Exception e) {
```
mozzy11, line 615: `safeMessage()` is a copy of ServiceRequestProvider's private helper (a third copy is queued in #3555) — move it to `FhirProviderUtils` and delete the copies.

src/test/java/org/openelisglobal/fhir/LocationFacadeTest.java
```diff
@@ -0,0 +1,637 @@
...
+    private static final String RACK_FHIRID = "f2cdeff8-8d5b-4023-bd7c-932b4b98b6f3";
+    private static final String SHELF_FHIRID = "f2cdeff8-8d5b-4023-bd7c-932b4b98b6a3";
+    private static final String BOX_FHIRID = "f2cdeff8-8d5b-4023-bd7c-932b4b98b1a3";
+    private static final String DEVICE_FHIRID = "f2cdeff8-8d5b-4023-bd7c-932b4b98b6d6";
+
+    private RestfulServer fhirServlet;
+    private ObjectMapper objectMapper;
+
+    @Autowired
+    private LocationProvider locatiobProvider;
```
mozzy11, line 51: Typo: `locatiobProvider` → `locationProvider` (also used a few lines below).

mozzy11 (commented):
Follow-up review (not superseding @pmanko's open CHANGES_REQUESTED — that stays the blocking state until he re-reviews). Credit where due: the follow-up commits substantially addressed his review — writes now go through `StorageLocationService`/per-type `@Transactional` services rather than raw DAOs, the room-builder NPE is guarded, and the device code-gen context is fixed. CI is fully green.

Two things still open from the review threads (inline): the search stub that forwards to the remote FHIR store while reads/writes hit the local DB (data-divergence risk, flagged twice), and the 5-way copy-paste in get/delete handling. Once those are resolved, please re-request pmanko's review.

mozzy11 (commented):
Status check (2026-07-28): no changes since the 2026-07-17 fixes; CI remains fully green. @pmanko's CHANGES_REQUESTED is still the blocking state, though his substantive points are now largely addressed in code (writes via `StorageLocationService`/per-type `@Transactional` services, room-builder NPE guard, device code-gen context fix).

The two items from the open review threads are still pending:
1. `searchLocationBundle` still declares eight params, ignores them, and forwards to the remote FHIR store while reads/writes hit the local DB — search can return different data than a read of the same Location. Implement local filtering (even identifier + name to start) or drop the dead params and document the delegation.
2. The 5-way copy-paste in the get/delete handlers (~260 lines that could be one generic handler resolved from `physicalType`).

@Isabirye1515 once those are in, please re-request review from @pmanko so he can lift the block.

mozzy11 (commented):
**Quick re-review (summary).** FHIR `LocationProvider` (CRUD + search) mapping the storage hierarchy (room/device/shelf/rack/box) → `Location`, with transforms and a `LocationFacadeTest`.

- **Good:** pmanko's CHANGES_REQUESTED largely addressed in code — writes now route through `@Transactional` storage services (not raw DAOs), room-builder NPE guarded, device code-gen context fixed, duplicate `setLabel` removed.
- **Still open — search stub:** `searchLocationBundle` declares 8 `@OptionalParam`s, ignores all of them, and forwards to the *remote* FHIR store while read/write use the local DB → a search can return different data than a read. Flagged 3×.
- **Still open — 5-way copy-paste:** get/delete lookup-then-transform repeated across all five services (~120 lines) despite the generic `getItemByFhirId`; `safeMessage` duplicates `ServiceRequestProvider`'s private helper (belongs in `FhirProviderUtils`).
- **Status:** BEHIND + pmanko's CHANGES_REQUESTED still blocking. Substance mostly fixed; next: resolve the search stub + get/delete dedup, rebase, re-request pmanko.

mozzy11 (approved):

- The storage level is read from the `storage-hierarchy` tag (physicalType.text as fallback), rooms carry `physicalType.text`, and a box's container format travels in `Location.type`, so a Location read from the facade can be PUT back unchanged (round-trip tests for room and box).
- Missing or unresolvable `partOf`, missing equipment type, and duplicate codes are 400/422 instead of database errors; POST ignores a client id; PUT of an unknown id is 404; the hierarchical identifier no longer overwrites a device code.
- The response Location is built once and that instance is mirrored to the FHIR store, ending the concurrent double transform of one Hibernate entity; the box occupancy query is flush-safe since StorageBox's lifecycle hook reaches it mid-flush.
- Per-type services use Spring's `@Transactional`, `safeMessage` lives in `FhirProviderUtils`, five-way switches collapsed, and `StorageLocationServiceSearchTest` now mocks the services (the CI failure).
- LocationFacadeTest (33), storage service/controller/DAO suites pass locally. Search still forwards to the FHIR store; local search for every provider follows in the next PR.

PR #3607, opened 2026-05-16
feat(vector v03): vector testing & identification workflow

mherman22:
## What this PR delivers

Fixes https://uwdigi.atlassian.net/browse/OGC-582

### Backend

- `VectorSpecimenIdentification` entity — species, method (MORPHOLOGICAL/MOLECULAR/BOTH), confidence, notes, physiological state (Detinova), per-specimen lifecycle stage override, molecular record FK
- `VectorMolecularRecord` entity — target gene, assay name, GenBank accession, linked result
- `VectorPool.deconvolutionStatus` + `deconvolutionOutcomePct` moved to pool level (one positive pool does not pollute sibling pools of the same sample)
- Liquibase migrations: `sample` (identification_status), `sample_item` (collection_location_id, collection_notes), `vector_specimen_identification`, `vector_molecular_record`, `vector_pool` (external_id, deconvolution_status, deconvolution_outcome_pct)
- `VectorIdentificationRestController` — identify, bulk-identify, worklist (status filter), specimens-for-lot, bloodmeal-panel add/dismiss, result-candidates
- `VectorDeconvolutionRestController` — worklist, initiate, GET/PUT `/pool/{id}`, preview
- `VectorDeconvolutionServiceImpl` — sub-pool creation (redistributes existing `sample_item`s via `vector_pool_member`, never duplicates), analysis copy with reflex eager-evaluation, pool-level positive evaluation, child-result completion check, force-complete, reflex preview

### Frontend

- `VectorIdentificationWorklist` — status dropdown filter, inline lot row expansion
- `VectorLotDetail` — species distribution bar chart, unified pool + specimen table with `└╴/├╴` tree connectors, deconvolution status badge, inline decon trigger
- `VectorSpecimenForm` — species ComboBox, method/confidence selects, lifecycle stage select, physiological state select (hidden when lifecycle ≠ ADULT), blood-meal panel suggestion notification for BLOOD_FED specimens, molecular details accordion (target gene, assay name, GenBank accession, linked result)
- `VectorBulkApplyTile` — batch-action inline tile; copies species/method/confidence/notes, skips molecular fields per FRS BR-V03-003
- `VectorInlineDeconPanel` — random / auto-by-species / manual (drag-and-drop) pool assignment, preview grouping, sticky save bar
- `VectorDeconvolutionWorklist` — deconvolution status view
- `VectorIdentificationService.js` — promise-wrapped API client for all identification + deconvolution endpoints

### Tests (54 green, 0 failures)

- `VectorIdentificationRestControllerIntegrationTest` — 17 tests (worklist filters, identify, bulk-identify, bloodmeal panel, unknown lot returns `[]`)
- `VectorDeconvolutionServiceImplTest` — 14 tests (sub-pool creation, member redistribution, analysis copy, pool-level status transitions, rejection rules)
- `VectorDeconvolutionServiceExtendedTest` — 14 tests (preview, forceComplete, child-result completion + outcome pct)
- `VectorDeconvolutionRestControllerTest` — 9 tests (worklist, initiate 400/409, pool GET/PUT, preview)
- DBUnit fixtures: `vector-decon-initiate.xml`, `vector-specimen-identification.xml`


reagan-meant:
1. The barcodes are not as we want them but I will fix this in a follow up otherwise see how env is doing it. Env might be broken too.
2. Vector sample summary not rendering the panels and tests well on the QA review page ( see how env displays them)
3. When you click on an order from the vector dashboard and navigate to the "Order" page (page 1),  it doesn't mantain the concept of pools instead you see a bunch of vectors that were created in the pool. We should mantain the idea of those vectors in their respective pools
4. In decon page, the identification page is not showing the tests whose results it's showing. It shouldn't just show results like "Detected" without showing which test. 
5. Also apply results to entire pool should be per result coz it might be true for one of the tests and not the other. 
6. If result is confirmed for a pool then we should show that the test result is confirmed for entire pool.

src/main/java/org/openelisglobal/dictionary/controller/rest/DictionaryRestController.java
```diff
@@ -66,6 +66,38 @@ public void initBinder(WebDataBinder binder) {
         binder.setAllowedFields(ALLOWED_FIELDS);
     }
 
+    @GetMapping(value = "/dictionary/categories/{categoryName}/entries", produces = "application/json")
+    @PreAuthorize("hasAnyRole('RESULTS', 'ADMIN', 'RECEPTION', 'VALIDATION', 'PATHOLOGIST', 'CYTOPATHOLOGIST')")
+    public ResponseEntity<List<DictionaryEntryDTO>> getDictionaryEntriesByCategoryName(
```
reagan-meant, line 71: @mherman22 don't we have a method that does this?
Can we move this logic to the service class as well if there is no method that does this?

src/main/java/org/openelisglobal/result/service/LogbookPersistServiceImpl.java
```diff
@@ -136,12 +139,37 @@ public List<Analysis> persistDataSet(ResultsUpdateDataSet actionDataSet, List<IR
...
 
+        evaluateVectorPositiveResults(actionDataSet, sysUserId);
+
         for (IResultUpdate updater : updaters) {
             updater.transactionalUpdate(actionDataSet);
         }
         return reflexAnalysises;
     }
 
+    private void evaluateVectorPositiveResults(ResultsUpdateDataSet actionDataSet, String sysUserId) {
```
reagan-meant: @mherman22 do we want to rename this method to something like 

_evaluateVectorResults_

src/main/java/org/openelisglobal/vector/deconvolution/controller/VectorDeconvolutionRestController.java
```diff
@@ -0,0 +1,447 @@
...
+        try {
+            return ResponseEntity.ok(deconvolutionService.getAvailablePanelTests(poolId));
+        } catch (RuntimeException e) {
+            LogEvent.logError(e);
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+        }
+    }
+
+    @GetMapping(value = "/pool/{poolId}", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<DeconvolutionResult> getDeconvolution(@PathVariable Long poolId) {
```
reagan-meant, line 150: @mherman22 can we ensure the business logic is moved to the service layer

src/main/java/org/openelisglobal/vector/identification/controller/VectorIdentificationRestController.java
```diff
@@ -0,0 +1,554 @@
...
+            }
+            return ResponseEntity.ok(rows);
+        } catch (RuntimeException e) {
+            LogEvent.logError(e);
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+        }
+    }
+
+    @GetMapping(value = "/lots/{lotId}/specimens", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<List<SpecimenDetailDTO>> getSpecimensForLot(@PathVariable Long lotId) {
```
reagan-meant, line 131: @mherman22 can we move the business logic to the service layer

reagan-meant (changes requested):

PR #3615, opened 2026-05-17
issue-3536: Remove unused programCode parameter from LabelMakerServlet

### LINKED ISSUE #3536: Cleanup: remove unused programCode parameter from LabelMakerServlet
[`LabelMakerServlet.java:274`](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/src/main/java/org/openelisglobal/common/servlet/barcode/LabelMakerServlet.java#L274) reads `request.getParameter("programCode")` and passes it through `validate(labNo, programCode, type, quantity, override)` at [line 309](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/src/main/java/org/openelisglobal/common/servlet/barcode/LabelMakerServlet.java#L309), but `validate()` itself never uses the parameter. The `accessionNumberValidator` it eventually calls (`getGeneralAccessionNumberValidator()`) doesn't take a programCode either.

No security issue — the value flows nowhere — just dead code. Cleanup is to drop the field from the validate signature and remove the `request.getParameter` call. ~5-line diff.

**Note:** `programCode` IS used elsewhere in the codebase (`ScanGeneratorProvider`, `ProgramAccessionValidator`) for accession-number generation against project programs — this issue is specifically about the unused copy in `LabelMakerServlet`.

---

Surfaced from PR #3528 audit. Tracked under the Frontend Modernization epic #3449.

Junaidarmaan:
## Related Issue

Fixes #3536

## Summary

Removed the unused `programCode` parameter from `LabelMakerServlet` and updated related tests.

## Changes

* Removed unused `programCode` request parameter
* Removed `programCode` from `validate()` method signature and calls
* Updated `LabelMakerServletTest`
* Cleaned up outdated JavaDoc


mozzy11 (commented):
Nice focused cleanup, @Junaidarmaan — I verified on develop that programCode is parsed and threaded into validate() but never actually used in the body, so this is legitimate dead-code removal with the test updated to match. One heads-up: #3817 (from the core team) also touches LabelMakerServlet test URLs, so expect a small rebase depending on merge order — I'd suggest rebasing after #3817 lands. I'll trigger the CI run; assuming the backend suite is green this should be a quick merge.

mozzy11 (commented):
Re-checked against current develop: `programCode` is still parsed and threaded through `validate()` without ever being used (LabelMakerServlet.java:274/309/378), and the current test file still passes the empty-string placeholder — so this cleanup remains valid and the branch is still mergeable. #3817 (the related test-URL fix) has not merged yet, so no rebase is forced at this point.

Two things before this can be approved:

1. Only the trivial workflow has run on this branch — the backend build/test suite never executed. A maintainer needs to approve the workflow run; approval should wait for a green backend suite since a test file changed.
2. Copilot's Javadoc point above is still open and squarely in scope: the retained `@param patientId` documents a parameter `validate()` does not have. Please drop that line (and optionally note the additional accepted `type` values) in the same pass.

With CI green and the Javadoc line removed, this is a quick merge.

Junaidarmaan:
Thanks for the review, @mozzy11.

I've addressed the remaining Javadoc comment by removing the stale `@param patientId` entry from `validate()` since it no longer exists in the method signature. I also ran Spotless and verified the related `LabelMakerServletTest` passes locally.

The branch is updated and ready for CI once the workflow is approved. Thanks!

PR #3622, opened 2026-05-19
Adding questionnaire to be db accessible

Isabirye1515:
## Summary
This PR introduces the foundational Questionnaire module support for OpenELIS, including entity modeling, persistence, Liquibase schema creation, dataset support, and service-layer testing.

### Added
- `Questionnaire`
- `QuestionnaireItem`
- `QuestionnaireAnswerOption`
- `QuestionnaireItemInitial`

### Features
- Parent/child questionnaire item hierarchy
- Support for multiple questionnaire item types using enums
- Initial/default item values
- Choice, boolean, integer, decimal, and string answer support
- UUID support for FHIR interoperability
- Hibernate sequence-based string ID generation aligned with existing OpenELIS patterns

### Database
Added Liquibase changelog support for:
- `questionnaire`
- `questionnaire_item`
- `questionnaire_answer_option`
- `questionnaire_item_initial`

Schema updates include:
- `last_updated`
- `sys_user_id`
- entity relationships
- indexes for lookup fields

### Tests
Added service-layer tests covering:
- retrieval of questionnaires
- insertion of questionnaires
- persistence of questionnaire items


mozzy11 (commented):
This is an impressive amount of careful groundwork — seven entities with a proper parent/child item hierarchy, the full DAO/service layering, indexes on the lookup columns, and real service-layer tests with DBUnit datasets. You clearly studied the OpenELIS patterns before writing this.

Before we build on it, one design question and a few fixes:

**Design question (main one):** develop currently treats FHIR Questionnaires as JSON parsed by QuestionnaireConfigurationHandler. This PR moves to full relational storage, which is a real architectural fork — it enables in-app editing and querying, but it means every FHIR field we care about must be mapped by hand forever. Let's get explicit maintainer agreement on this direction (maybe a short note in the PR description on why relational vs storing the FHIR resource) before merging.

**Fixes:**
- AppTestConfig.java: the new scan entry is the literal string "package org.openelisglobal.questionnaire" — the word *package* is inside the string. Please fix to "org.openelisglobal.questionnaire" and re-run the tests to confirm the beans were really being scanned.
- Class typo: QuestionionnaireDaoImpl → QuestionnaireDaoImpl.
- Package names: use lowercase daoimpl, and put service impls in service next to their interfaces, matching e.g. org.openelisglobal.provider.
- Liquibase: you created 3.6.x.x/001-questionnaire.xml, but current changesets are landing in 3.5.x.x — please check with the maintainers which directory this should live in.
- @Transactional belongs on services only per the project constitution — please move it off the DAO impls.
- Consider renaming the QuestionnaireResponse entity (it shadows the HAPI model class you'll be transforming to/from later).

None of these are hard changes — the bones here are genuinely good. Looking forward to the next revision.

src/main/resources/liquibase/qc/001-questionnaire.xml
```diff
@@ -0,0 +1,401 @@
...
+
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="questionnaire"/>
+            </not>
+        </preConditions>
+
+        <createTable tableName="questionnaire">
+
+            <column name="id" type="INTEGER">
```
mozzy11: **Blocking:** the entities use `GenerationType.SEQUENCE` (`questionnaire_seq`, `questionnaire_item_seq`, …) but this changelog has no `createSequence` for any of them — on a real deployment every insert fails with "sequence does not exist". The service tests don't catch it because `test-persistence.xml` uses `hibernate.hbm2ddl.auto=update`, which auto-creates them. Please add `createSequence` changesets for all seven sequences and validate against a Liquibase-built DB.

src/main/resources/liquibase/qc/001-questionnaire.xml
```diff
@@ -0,0 +1,401 @@
...
+
+            <column name="code" type="VARCHAR(10)">
+                <constraints nullable="false"/>
+            </column>
+
+            <column name="has_item" type="BOOLEAN" defaultValueBoolean="false">
+                <constraints nullable="false"/>
+            </column>
+
+            <column name="parent_questionnaire_id" type="VARCHAR(20)"/>
```
mozzy11: Type mismatch with the entity mapping: `parent_questionnaire_id` is `VARCHAR(20)` but joins `questionnaire.id` which is `INTEGER` (same for `questionnaire_item.questionnaire_id`, `parent_item_id`, `questionnaire_answer_option.questionnaire_item_id`, `questionnaire_item_initial.questionnaire_item_id`). Also no `addForeignKeyConstraint`s are declared — only indexes — so referential integrity is unenforced. Please align column types with the referenced PKs and add FK constraints.

src/main/resources/liquibase/base-changelog.xml
```diff
@@ -30,4 +30,5 @@
   <include file="liquibase/qc/007-add-qc-menu-items.xml" />
   <include file="liquibase/qc/008-add-instrument-fk-constraints.xml" />
   <include file="liquibase/qc/009-nullable-unit-of-measure.xml" />
+  <include file="liquibase/qc/001-questionnaire.xml" />
```
mozzy11: This changelog lives in `liquibase/qc/` (the Quality Control module's directory) and its `001-` prefix collides with the existing `qc/001-create-qc-tables.xml`. New schema changesets belong under `src/main/resources/liquibase/3.5.x.x/` per project convention. Please relocate and renumber.

src/main/java/org/openelisglobal/questionnaire/valueholder/Questionnaire.java
```diff
@@ -0,0 +1,204 @@
...
+import org.openelisglobal.common.valueholder.BaseObject;
+
+@Entity
+@Table(name = "questionnaire")
+public class Questionnaire extends BaseObject<Integer> {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "questionnaire_generator")
+    @SequenceGenerator(name = "questionnaire_generator", sequenceName = "questionnaire_seq", allocationSize = 1)
+    private int id;
```
mozzy11, line 30: `private int id` (primitive) is inconsistent with the other entities in this PR (`Integer`), and `setId(Integer)` will NPE on auto-unboxing if passed null. Also the nested enum `QuestionaireStatus` has a typo (should be `QuestionnaireStatus`) — worth fixing before it spreads.

mozzy11 (changes requested):
Thanks @Isabirye1515 — the Questionnaire/QuestionnaireResponse persistence module follows the Valueholder→DAO→Service layering correctly and the DBUnit-backed service tests are real integration tests. Nice structure overall.

**Blocking:** the Liquibase changelog never creates any of the sequences the entities' `@SequenceGenerator`s reference (14 generator references, 0 `createSequence`) — inserts will fail on any Liquibase-provisioned database. Tests pass only because `test-persistence.xml` runs `hibernate.hbm2ddl.auto=update`, which silently creates a divergent schema. There are also FK type mismatches (VARCHAR(20) columns joining INTEGER PKs) and the changelog is misfiled under the QC module's directory. Details inline.

mozzy11 (commented):
**Quick re-review (summary).** New relational persistence for FHIR Questionnaires/Responses — 7 entities + DAO/Service layers, a Liquibase changeset, and DBUnit service tests.

- **Good:** prior schema CHANGES_REQUESTED largely addressed — all 7 `createSequence` present (match the generator refs), FK columns retyped to INTEGER, 9 FK constraints added, changelog collision fixed, and the component scan corrected.
- **Still open:** `@Transactional` remains on all 4 DAO impls (constitution: services only). Package `daoImpl` is mixed-case (convention is lowercase `daoimpl`). Changelog landed in a new `3.6.x.x/` dir though the maintainer pointed to `3.5.x.x` — confirm the target.
- **Main unresolved:** the design fork (relational storage vs storing the FHIR JSON) still needs explicit maintainer sign-off + a note in the PR description.
- **Status:** BEHIND + CHANGES_REQUESTED. Schema mechanics now look sound; next: get maintainer agreement on direction, strip `@Transactional` off the DAOs, rebase.

src/test/resources/persistence/test-persistence.xml
```diff
@@ -190,6 +190,13 @@
         <class>org.openelisglobal.esig.valueholder.EsigFirstUseCertification</class>
         <class>org.openelisglobal.barcode.valueholder.SampleBarcodeInfo</class>
         <class>org.openelisglobal.barcode.valueholder.SampleItemBarcodeInfo</class>
+        <class>org.openelisglobal.questionnaire.valueholder.Questionnaire</class>
```
mozzy11, line 193: The main app builds its EntityManagerFactory from `src/main/resources/persistence/persistence.xml` (explicit class list, package scan is commented out in `HibernateConfig`), and these seven classes are not listed there. At runtime the DAOs would fail with an unknown-entity error on first use. I will add them there in the follow-up PR.

src/main/resources/liquibase/3.6.x.x/001-questionnaire.xml
```diff
@@ -0,0 +1,415 @@
...
+            <column name="id" type="INTEGER">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="fhir_uuid" type="UUID">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="questionnaire_id" type="INTEGER">
+                <constraints nullable="false"/>
+            </column>
+            <column name="subject_id" type="VARCHAR(20)">
```
mozzy11, line 263: `QuestionnaireResponse.subject` is a `@ManyToOne Patient`, but `patient.id` is numeric, so a join on this `VARCHAR(20)` column fails in Postgres. Programme responses can also belong to patientless environmental samples, so the subject needs to be optional. Both handled in the follow-up.

src/main/java/org/openelisglobal/questionnaire/valueholder/Questionnaire.java
mozzy11, line 30: Primitive `int` while `BaseObject<Integer>` treats the id as nullable, so `setId(null)` would NPE and a new instance reports id 0 instead of null. The other six entities use `Integer`.

mozzy11 (approved):
**Decision:** Approve. The module is self-contained, nothing calls it yet, and it gives us the local storage layer the FHIR-independence work needs, so I am merging it as the foundation and taking the items below in the follow-up PR.

**Summary**
- Entities, DAOs and services follow the `BaseObject` / `BaseDAO` / `AuditableBaseObjectServiceImpl` pattern; sequences, FK types, indexes and `MARK_RAN` preconditions in the changeset are right.
- Both new test classes pass here (9 + 8) and the backend CI step passed; the red check is the SBOM download hitting a GitHub 504, rerun queued.
- Runtime gap: the seven entities are registered in the test persistence unit only, so the main app cannot use them yet (inline).
- `questionnaire_response.subject_id` is `VARCHAR(20)` while the entity joins it to `patient.id` (numeric), and programme responses can belong to patientless environmental samples (inline).

**Required (follow-up PR, mine)**
- Register the entities in `src/main/resources/persistence/persistence.xml`; fix the `subject_id` type and make the subject optional.

**Optional**
- `Questionnaire.id` is a primitive `int` while `BaseObject<Integer>` is nullable; DAO impls carry `@Transactional` (services own transactions here); package `daoImpl` vs the usual lowercase `daoimpl`.

PR #3625, opened 2026-05-19
feat(sample-types): Sample Type Management module with domains, terminology, display order, disposal & bidirectional test links (OGC-296)

HerbertYiga:
Implements **[OGC-296 — Sample Type Management Module](https://uwdigi.atlassian.net/browse/OGC-296)** per **FRS v2.1** (2026-07-14, the canonical spec superseding the v1.0 five-tab design).

Design references:
- [FRS v2.1](https://github.com/DIGI-UW/openelis-work/blob/main/designs/admin-config/sample-type-management.md) · [JSX mockup](https://github.com/DIGI-UW/openelis-work/blob/main/designs/admin-config/sample-type-management.jsx) · [gallery entry](https://digi-uw.github.io/openelis-work/#/admin-config/sample-type-management) · [interactive walkthrough](https://digi-uw.github.io/openelis-work/designs/admin-config/test-catalog-panels-sampletypes.html)

## What this delivers

**Sample Types management surface** (`/MasterListsPage/SampleTypeManagement`, admin-gated), reached under **Admin → Test Catalog Management → All Sample Types**:
- **List** with search + domain filter, Domain tags, test counts, and status — one row per sample type.
- **Editor** with SideNav sections (shell-context pattern, no tabs), all five FRS v2.1 sections:
  - **Basic Info** — name (localized), description, abbreviation, single required **Domain** (Clinical / Environmental / Vector), active toggle with an in-use deactivation warning.
  - **Associated Tests** — **read-only** view of the tests carrying this sample type (per v2.1: specimen membership is managed from the test side), with the view-only note.
  - **Display Order** — positions the sample type in the order-entry Sample Type menu (the real `sortOrder`); lists all types in order with this one highlighted, renumbers the whole sequence densely on save.
  - **Disposal** — free-text disposal instructions (reference guidance; the authoritative structured handling stays per-test `TestSampleHandling` and per-specimen Sample Storage).
  - **Terminology** — the full terminology mapper (LOINC / SNOMED / CIEL / OCL / **WHONET**) with relationship qualifiers, soft-delete reconcile semantics mirroring the test terminology store.
- **Backend**: `sample_type_terminology_mapping` store (service/DAO/valueholder), `whonet_code` + `disposal_instructions` on `type_of_sample`, `GET /rest/sample-types`, `GET /rest/sample-types/{id}`, `PUT /rest/sample-types/{id}`, `PUT /rest/sample-types/{id}/display-order`, `GET/PUT /rest/sample-types/{id}/terminology` — all `hasRole('ADMIN')`.
- **Schema**: Liquibase `064` (whonet_code), `065` (terminology mapping), `066` (domain enum migration), `067` (disposal instructions) — all precondition-guarded.

## Domain migration (v2.1 Dependency 4 — now included)

`type_of_sample.domain` was a legacy `varchar(1)` (`'H'`-style codes). Changeset `066` widens it to `varchar(20)`, backfills the legacy chars to the enum per D-030 (H/N → CLINICAL, E → ENVIRONMENTAL, A → VECTOR, else CLINICAL — matching the OGC-936 `test.domain` backfill), and makes it `NOT NULL DEFAULT 'CLINICAL'`. All writers now store the enum value; readers stay bilingual (`SampleTypeDomainMapper.normalize`) so fixture/plugin-inserted legacy-char rows keep resolving. The order-entry Sample Type menu and the Test Catalog D-030 compatibility guard accept both forms.

## Review fixes applied

Both open reviews (@samuelmale, @mozzy11) are addressed: orphaned components deleted (~3,400 lines), all debug scaffolding/simulation/mock code stripped, `hasRole('ROLE_ADMIN')` → `hasRole('ADMIN')` pinned by security tests, `UserServiceImpl` order-entry change reverted, create restored to inactive-until-configured, localization rename fixed, en.json regressions restored + orphaned keys pruned, Liquibase renumbered behind develop's 060–063. See the [review-fixes comment](https://github.com/DIGI-UW/OpenELIS-Global-2/pull/3625#issuecomment-5094386059). Sidenav follow-up: entity links pinned atop the Test Catalog menu and the menu no longer collapses when navigating between the lists/editors.

## Deliberately out of this PR
- Bulk export (deferred in the FRS, parity with panels).

## Verification
- **Backend**: full `mvn test` suite green; new `SampleTypeDomainMapperTest` (D-030 mapping), reorder service tests, and 14 controller security tests (incl. display-order + get-by-id 401/403/404 + 422).
- **Frontend**: 125 files / 814 tests green (incl. new DisplayOrderSection + DisposalSection tests); vite build clean; spotless + prettier applied.
- **Live** (dev stack, Liquibase 064–067 confirmed applied): list/create/edit/terminology/associated-tests + the new Display Order (dense renumber verified in DB) and Disposal (persists across reload) round-trips all pass; domain edit writes the enum value; order-entry Sample Type list still resolves post-migration.


.gitignore
```diff
@@ -149,3 +152,10 @@ frontend/npm-test-*.log
 # Per-PR scratch session notes (never commit)
 PR-*-STATUS-REPORT.md
 volume/nce-attachments/
+
+# Debug files for domain persistence work (local dev artifacts)
+debug-domain-component.jsx
+debug_domain.sql
+domain-debug.patch
+domain-persistence-test.md
+test-domain-persistence.sql
```
samuelmale: These are too specific to your dev environment; we probably won't need them.

src/main/java/org/openelisglobal/testconfiguration/controller/rest/SampleTypeCreateRestController.java
```diff
@@ -93,10 +93,19 @@ public SampleTypeCreateForm postSampleTypeCreate(HttpServletRequest request,
...
         String userId = getSysUserId(request);
+        String frontendDomain = form.getDomain();
+        String backendDomainCode = mapFrontendDomainToBackendCode(frontendDomain);
+
+        // Log domain mapping for verification
+        System.out.println("=== SAMPLE TYPE CREATION DEBUG ===");
+        System.out.println("Sample Name: " + identifyingName);
+        System.out.println("Frontend Domain: " + frontendDomain);
+        System.out.println("Backend Domain Code: " + backendDomainCode);
+        System.out.println("=====================================");
```
samuelmale: Use standard logging

src/main/java/org/openelisglobal/testconfiguration/controller/rest/SampleTypeCreateRestController.java
```diff
@@ -112,8 +121,40 @@ public SampleTypeCreateForm postSampleTypeCreate(HttpServletRequest request,
         try {
             sampleTypeCreateService.createAndInsertSampleType(localization, typeOfSample, workplanModule, resultModule,
                     validationModule, workplanResultModule, resultResultModule, validationValidationModule);
+
+            // Verification: confirm the sample type was saved with correct domain
+            System.out.println("✓ Sample Type '" + identifyingName + "' successfully saved to database with domain: " + backendDomainCode);
```
samuelmale: Ditto

src/main/java/org/openelisglobal/testconfiguration/controller/rest/SampleTypeCreateRestController.java
```diff
@@ -112,8 +121,40 @@ public SampleTypeCreateForm postSampleTypeCreate(HttpServletRequest request,
...
+                    }
+                } else {
+                    System.err.println("✗ DATABASE VERIFICATION: Could not find saved sample type - may not be committed yet");
+                }
+            } catch (Exception verificationException) {
+                System.err.println("⚠ Verification check failed: " + verificationException.getMessage());
+            }
+
         } catch (LIMSRuntimeException e) {
+            System.err.println("✗ Failed to save Sample Type '" + identifyingName + "' to database: " + e.getMessage());
```
samuelmale: Ditto

src/main/java/org/openelisglobal/testconfiguration/controller/rest/SampleTypeCreateRestController.java
```diff
@@ -144,20 +185,46 @@ private RoleModule createRoleModule(String userId, SystemModule workplanModule,
...
                 identifyingName.length() > 10 ? identifyingName.substring(0, 10) : identifyingName);
-        typeOfSample.setIsActive(false);
+        typeOfSample.setIsActive(true);
         typeOfSample.setSortOrder(Integer.MAX_VALUE);
         typeOfSample.setSysUserId(userId);
         String identifyingNameKey = identifyingName.replaceAll(" ", "_");
         typeOfSample.setNameKey("Sample.type." + identifyingNameKey);
+
+        // Verification log
+        System.out.println("TypeOfSample created with backend domain code: " + backendDomainCode);
```
samuelmale: Ditto

samuelmale (dismissed):

src/main/java/org/openelisglobal/common/management/controller/rest/SampleTypeManagementRestController.java
```diff
@@ -1,41 +1,661 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
 @RequestMapping("/rest")
-@PreAuthorize("hasRole('ADMIN')")
-public class SampleTypeManagementRestController extends BaseController {
+@PreAuthorize("hasRole('ROLE_ADMIN')")
```
mozzy11: **Auth gate looks wrong (blocking).** Every other admin REST controller in this codebase uses `@PreAuthorize("hasRole('ADMIN')")` (86 usages), and `CustomUserDetailsService` grants the authority `ROLE_ADMIN`. Spring's `hasRole('ROLE_ADMIN')` prepends `ROLE_` and checks for `ROLE_ROLE_ADMIN`, which no user has — so this would return **403 for real admins**. Please revert to `hasRole('ADMIN')`. (A non-admin/admin auth test on a mutating route would catch this — see my note on the security test.)

src/main/java/org/openelisglobal/common/management/controller/rest/SampleTypeManagementRestController.java
```diff
@@ -1,41 +1,661 @@
...
+        }
+
+        public SampleTypeManagementDTO(TypeOfSample typeOfSample) {
+            this.id = typeOfSample.getId();
+
+            // Debug the localization to see what's causing "en" to appear
+            String nameValue = typeOfSample.getDescription(); // Default to description
+            if (typeOfSample.getLocalization() != null) {
+                String localizedValue = typeOfSample.getLocalization().getLocalizedValue("en");
+                LogEvent.logInfo("SampleTypeManagementDTO", "constructor",
```
mozzy11: **Debug logging must go.** This `logInfo`/`logWarn` block (`"DEBUG - ID: ..."`) runs on *every* DTO construction — once per sample type on every list load — and logs localized values. samuelmale already flagged logging on the sibling controller. Please drop the debug logs; the domain fallback doesn't need to log.

src/main/java/org/openelisglobal/common/management/controller/rest/SampleTypeManagementRestController.java
```diff
@@ -1,41 +1,661 @@
...
+            return message;
+        }
+
+        public T getData() {
+            return data;
+        }
+    }
+
+    @GetMapping(value = "/sample-types/test")
+    public ResponseEntity<String> testEndpoint() {
```
mozzy11: Leftover debug endpoint (`return ResponseEntity.ok("...is working!")`). Please remove before merge.

src/main/java/org/openelisglobal/common/management/controller/rest/SampleTypeManagementRestController.java
```diff
@@ -1,41 +1,661 @@
...
+
+        } catch (Exception e) {
+            LogEvent.logError("SampleTypeManagementRestController", "createSampleType", e.getMessage());
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR)
+                    .body(new ApiResponse<>(false, "Error creating sample type: " + e.getMessage(), null));
+        }
+    }
+
+    @PostMapping(value = "/sample-types/update")
+    // @PreAuthorize("hasRole('ROLE_ADMIN')") // Temporarily removed to debug
```
mozzy11: Remove the commented-out `@PreAuthorize` and the `// Temporarily removed to debug` note — dead scaffolding, and misleading since the class-level gate still applies.

src/main/java/org/openelisglobal/common/management/controller/rest/SampleTypeManagementRestController.java
```diff
@@ -1,41 +1,661 @@
...
+        switch (backendDomain.toUpperCase()) {
+        case "H":
+            return "CLINICAL"; // Human/Clinical
+        case "E":
+            return "ENVIRONMENTAL"; // Environmental
+        case "V":
+            return "VECTOR"; // Vector surveillance
+        case "A":
+            return "CLINICAL"; // Animal samples show as Clinical for now
+        case "B":
```
mozzy11: **Domain mapping round-trips lossily and conflates the legacy column.** Here `B` (Both) and `A` (Animal) both collapse to `CLINICAL`, but the forward map writes `BOTH`→`B`; a 'Both' type therefore shows as Clinical and is silently rewritten to `H` on the next save — data loss. More fundamentally, per Casey's **OGC-296 v2.1** note the legacy `TYPE_OF_SAMPLE.DOMAIN` varchar(1) is **not** the Clinical/Environmental/Vector enum (that enum isn't built yet), and the design calls for a *declared migration* backfilling existing rows to `CLINICAL`. Overloading the legacy column reinterprets existing data. This needs to align with v2.1 before merge.

src/main/java/org/openelisglobal/testconfiguration/controller/rest/SampleTypeCreateRestController.java
```diff
@@ -112,8 +118,46 @@ public SampleTypeCreateForm postSampleTypeCreate(HttpServletRequest request,
...
                     validationModule, workplanResultModule, resultResultModule, validationValidationModule);
+
+            LogEvent.logInfo(this.getClass().getSimpleName(), "postSampleTypeCreate", "Sample Type '" + identifyingName
+                    + "' successfully saved to database with domain: " + backendDomainCode);
+
+            // Additional verification: Query the database to confirm the domain was saved
+            // correctly
+            try {
+                // Wait a moment for transaction to commit
+                Thread.sleep(500);
```
mozzy11: **Blocking `Thread.sleep(500)` on every create.** This whole 'DATABASE VERIFICATION' block sleeps 500 ms and then scans `getAllTypeOfSamples()` purely to log a DOMAIN MATCH/MISMATCH check. It adds latency, loads the full table, and is debug-only. Please remove the entire verification block.

src/main/java/org/openelisglobal/testconfiguration/controller/rest/SampleTypeCreateRestController.java
```diff
@@ -144,20 +188,53 @@ private RoleModule createRoleModule(String userId, SystemModule workplanModule,
...
         typeOfSample.setLocalAbbreviation(
                 identifyingName.length() > 10 ? identifyingName.substring(0, 10) : identifyingName);
-        typeOfSample.setIsActive(false);
+        if (whonetCode != null) {
+            String trimmed = whonetCode.trim();
+            if (!trimmed.isEmpty() && trimmed.length() <= 5) {
+                typeOfSample.setWhonetCode(trimmed);
+            }
+        }
+        typeOfSample.setIsActive(true);
```
mozzy11: This flips the legacy create flow: sample types were previously created **inactive** (`setIsActive(false)`) until tests were assigned, and the Activation screen still says types need tests before they can be activated. Active-by-default is a behavior change — intended? If so, the Activation copy/gating needs to match.

src/main/java/org/openelisglobal/systemuser/service/UserServiceImpl.java
```diff
@@ -308,13 +308,27 @@ public List<IdValuePair> getUserSampleTypes(String systemUserId, String roleName
             });
         }
 
+        // Also include all active clinical sample types even if they don't have tests
+        // associated yet
+        // This ensures newly created clinical sample types appear in order entry
+        // dropdowns
+        List<TypeOfSample> allActiveClinicalSampleTypes = typeOfSampleService
```
mozzy11: **Order-entry behavior change — please double-check.** This injects *every* active HUMAN-domain sample type into the user's order-entry dropdown, regardless of test association or the role/lab-unit scoping computed above, and silently excludes ENVIRONMENTAL/VECTOR. That can flood the dropdown with types that have no orderable tests. Is this intended for OGC-296, or leaking in from the domain work? It needs its own test either way.

frontend/src/components/admin/sampleTypeManagement/SampleTypeManagement.jsx
```diff
@@ -0,0 +1,1880 @@
...
+      return "VECTOR";
+    case "A":
+      return "CLINICAL"; // Animal samples show as Clinical for now
+    default:
+      return "CLINICAL";
+  }
+};
+
+// ─── Mock Data ────────────────────────────────────────────────────
+const MOCK_SAMPLE_TYPES = [
```
mozzy11: ~135 lines of `MOCK_SAMPLE_TYPES` shipped in the bundle (and 'Drinking Water' is miscategorised as VECTOR). Please remove the mock data and the simulation path.

frontend/src/components/admin/sampleTypeManagement/SampleTypeManagement.jsx
```diff
@@ -0,0 +1,1880 @@
...
+    active: true,
+    domain: "CLINICAL",
+    testCount: 8,
+  },
+];
+
+// ─── Main Component ───────────────────────────────────────────────
+
+// Set to true for testing without authentication, false for real database integration
+const USE_SIMULATION_MODE = false; // DATABASE MODE - Real PostgreSQL integration
```
mozzy11: Dead feature flag — always `false`. Remove it and the `if (USE_SIMULATION_MODE)` branch.

frontend/src/components/admin/sampleTypeManagement/SampleTypeManagement.jsx
```diff
@@ -0,0 +1,1880 @@
...
+  // Fetch sample types from backend or use mock data
+  useEffect(() => {
+    const fetchSampleTypes = async () => {
+      try {
+        setIsLoading(true);
+        setLoadError(null);
+
+        if (USE_SIMULATION_MODE) {
+          // Use mock data for testing without authentication
+          console.log("Using simulation mode - loading mock data");
```
mozzy11: Several `console.log`/`console.error` debug lines here (incl. `=== DOMAIN LOADING DEBUG ===`). Please strip all console logging from the component.

frontend/src/components/admin/sampleTypeManagement/SampleTypeActivation.jsx
```diff
@@ -0,0 +1,605 @@
+import React, { useContext, useState, useRef, useEffect } from "react";
```
mozzy11: **Orphaned file.** This — along with `SampleTypeAdd`, `SampleTypeBulkImport`, `SampleTypeRenameEntry`, `SampleTypeTestAssignment`, `AddSampleTypeModal`, `EditSampleTypeModal` — is **not imported by the routed `SampleTypeManagement.jsx`** (which only pulls in `TerminologySection`), and they call REST endpoints that don't exist in the new controller (`/rest/SampleTypeManagement/activate`, `/rename`, `/testAssignments`, `/reassignTests`, `/bulkImport`). That's ~3.5k lines of dead code, also full of hardcoded hex colors/emoji and hardcoded English strings (violating the Carbon + React-Intl constitution principles). Please delete what isn't wired in, or wire it in properly.

frontend/src/components/admin/sampleTypeManagement/SampleTypeBulkImport.jsx
```diff
@@ -0,0 +1,906 @@
+import React, { useContext, useState, useRef, useEffect } from "react";
```
mozzy11: Orphaned (see the SampleTypeActivation note), the upload 'progress' is faked with a `setInterval` timer rather than real progress, and the `/bulkImport` endpoint doesn't exist. Bulk import isn't in OGC-296 v2.1 scope — recommend dropping this file.

mozzy11 (commented):
## Review against OGC-296 (Sample Type Management Module)

Thanks @HerbertYiga — there's real, well-built work here (the **terminology backend + `TerminologySection.jsx`** are clean, follow the DAO→Service→Controller pattern, mirror the Test Catalog editor nicely, and are covered by soft-delete reconcile logic). But the PR as a whole isn't mergeable yet, mainly because it carries a large amount of debug/scaffolding code and has drifted from the current spec. Note I'm **not** approving — @samuelmale's earlier `CHANGES_REQUESTED` (logging, `.gitignore`) is still open and some of it recurs here.

### Spec alignment (please read first)
OGC-296 has moved to **FRS v2.1** (Casey's comments, 2026-07-14/15), which supersedes the original 5-tab / multi-domain design this PR started from. Key deltas that affect this PR:
- Sample Types is a **context in the Test Catalog Management shell** (the routed `SampleTypeManagement.jsx` + `TerminologySection` already move this way — good).
- **Associated Tests is read-only** — adding a test to a sample type means creating a *specimen variant* in the Test Catalog, not reassigning here. `SampleTypeTestAssignment.jsx` conflicts with this.
- **Single-value domain**, and the enum domain **isn't built yet** — `TYPE_OF_SAMPLE.DOMAIN` is a legacy `varchar(1)`, so a *declared migration* (backfill → CLINICAL) is required rather than overloading the column.

### Blocking
1. **~3,500 lines of orphaned code.** Seven components (`SampleTypeAdd/Activation/BulkImport/RenameEntry/TestAssignment`, `Add/EditSampleTypeModal`) are never imported by the routed page and call **backend endpoints that don't exist** (`/activate`, `/rename`, `/testAssignments`, `/reassignTests`, `/bulkImport`). Delete what isn't wired in.
2. **Debug scaffolding in backend hot paths** — `Thread.sleep(500)` + full-table scan on every create; `"DEBUG - ..."` logging per DTO; a `/sample-types/test` "is working!" endpoint; a commented-out `@PreAuthorize`.
3. **Likely broken auth gate** — `hasRole('ROLE_ADMIN')` (double-prefixes to `ROLE_ROLE_ADMIN`) vs the universal `hasRole('ADMIN')`.
4. **Liquibase 060/061 collide** with existing develop changesets — renumber to 064+.

### High
- **Domain handling** overloads the legacy varchar(1) column and round-trips lossily (`Both`/`Animal` → Clinical → `H`). Align with v2.1's migration approach.
- **`UserServiceImpl` order-entry change** floods dropdowns with all active HUMAN sample types regardless of scoping; excludes ENV/VECTOR. Needs justification + a test.
- **Carbon / i18n violations** in the orphaned files: inline hex colors, emoji, and hardcoded English strings (constitution requires Carbon tokens + React-Intl keys).
- **Frontend mock data + `USE_SIMULATION_MODE` + `console.log`** shipped in `SampleTypeManagement.jsx`.

### Medium / Minor
- Accidental i18n regressions in `en.json` (a blanked key, a changed sidenav label) — look like rebase artifacts.
- Security test only covers `GET`; the new mutating + terminology endpoints are untested.
- REST shape inconsistency: update is `POST /sample-types/update` rather than `PUT /sample-types/{id}`.
- Trailing-newline / spotless nits on `SampleTypeManagementForm.java`, `SampleTypeDisplay.java`, the controller (`\ No newline at end of file`). `SampleTypeDisplay` also looks unused.

### Suggested path forward
1. Rebase onto develop and renumber the Liquibase changesets; resolve the flagged conflicts.
2. Delete the orphaned components (and the mock/simulation/console/debug code) — keep the routed page + terminology feature, which are the parts aligned with v2.1.
3. Fix the auth annotation and add a non-admin auth test on a mutating endpoint.
4. Re-scope domain handling to the v2.1 single-domain + declared-migration model, or split it into its own issue so this PR can land the master-list + terminology surface first.
5. Run `mvn spotless:apply` and `cd frontend && npm run format`.

Happy to re-review once the dead code is stripped and the rebase/auth items are sorted — the terminology work is genuinely close.

frontend/src/components/admin/sampleTypeManagement/SampleTypeTestAssignment.jsx
```diff
@@ -0,0 +1,412 @@
+import React, { useContext, useState, useRef, useEffect } from "react";
```
mozzy11: Orphaned, and it contradicts the current spec: **OGC-296 v2.1 makes Associated Tests read-only** (a test's specimen is part of its identity — you add a test to a sample type by creating a specimen variant in the Test Catalog, not by reassigning here). This reassignment UI should be removed.

frontend/src/components/admin/sampleTypeManagement/EditSampleTypeModal.jsx
```diff
@@ -0,0 +1,585 @@
...
+                            placeholder="Nom en français"
+                          />
+                        </Column>
+                      </Grid>
+                    </div>
+                  </Stack>
+                </div>
+              </TabPanel>
+
+              {/* Associated Tests Tab */}
```
mozzy11: The 'Associated Tests' tab renders an empty `<div>`, and this modal is orphaned. It also presents a single-select Domain enum that isn't built in the backend yet per v2.1. Recommend removing this modal with the other unrouted files.

frontend/src/languages/en.json
```diff
@@ -1074,7 +1074,7 @@
   "notificationtrigger.column.trigger": "Trigger",
   "notificationtrigger.config.subtitle": "Enable or disable notification triggers, select delivery channels, and configure recipient types per event.",
   "notificationtrigger.config.title": "Notification Trigger Configuration",
-  "notificationtrigger.disabled.hint": "Enable this trigger to configure its channels and recipients.",
+  "notificationtrigger.disabled.hint": "",
```
mozzy11: Unrelated existing key blanked: `"notificationtrigger.disabled.hint": ""`. Looks like an accidental rebase artifact — please restore its English value. (Also `sidenav.label.addorder` changed 'Add Generic Order'→'Add Order' and a plural-format change on `referral.banner.stuckReferralsSubtitle` — revert if unintended. New keys only, per the i18n principle.)

src/test/java/org/openelisglobal/common/management/controller/rest/SampleTypeManagementRestControllerSecurityTest.java
```diff
@@ -24,20 +28,21 @@ public class SampleTypeManagementRestControllerSecurityTest extends SecuritySlic
...
 
     @Test
     public void testSampleTypeManagement_NonAdminRole_Returns403() throws Exception {
-        mockMvc.perform(get("/rest/SampleTypeManagement").with(user("results").roles("RESULTS"))
+        mockMvc.perform(get("/rest/sample-types").with(user("results").roles("RESULTS"))
                 .contentType(MediaType.APPLICATION_JSON)).andExpect(status().isForbidden());
     }
 
     @Test
     public void testSampleTypeManagement_AdminRole_Returns200() throws Exception {
```
mozzy11, line 43: Coverage gap: only `GET /rest/sample-types` auth is tested. The new mutating endpoints (`POST /sample-types`, `POST /sample-types/update`, `GET/PUT /sample-types/{id}/terminology`) have no auth tests. Adding a non-admin 403 case on a mutating route would also surface the `hasRole('ROLE_ADMIN')` bug above.

src/main/resources/liquibase/3.5.x.x/065-sample-type-terminology-mapping.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mozzy11, line 1: **Changeset filename collision.** develop already has `060-terminology-sample-type-scope.xml` and `061-awaiting-specimen-order-status.xml` (plus 062, 063). This PR adds `060-add-whonet-code...` and this `061-...` — after rebase you'll have two 060s and two 061s. Please renumber to 064/065+ and fix the `base.xml` include order. Likely part of the conflict the bot flagged.

mozzy11:
Pushed `392205ada` + `548b938f2` applying both open reviews (@samuelmale's and mine) so this can move — @HerbertYiga please pull before continuing.

**Merge/conflicts:** merged current develop; the changeset collision is resolved by renumbering this PR's Liquibase files to **064/065** (develop's 060–063 shipped with OGC-1145). Both changesets now carry MARK_RAN preconditions.

**Applied from the reviews:**
- Deleted the 7 orphaned components (~3,400 lines) incl. `SampleTypeTestAssignment` (v2.1 makes Associated Tests read-only) — the routed page + `TerminologySection` + the terminology backend are kept intact.
- All debug scaffolding removed: `Thread.sleep(500)` + DB-verification block, per-DTO DEBUG logs (also @samuelmale's logging comments), the `/sample-types/test` endpoint, the commented `@PreAuthorize`, all `console.log`/`alert`s and the `USE_SIMULATION_MODE` mock path.
- **Auth fixed:** `hasRole('ROLE_ADMIN')` → `hasRole('ADMIN')` (the former checks `ROLE_ROLE_ADMIN` and 403s real admins), pinned by new non-admin 403 security tests on the mutating + terminology endpoints.
- **Domain:** mapped via the D-030 legacy chars OGC-1145 just shipped (H/N→Clinical, E→Environmental, **A**→Vector — the invented `'V'` would be incompatible with the new domain guard); chars only rewritten when the domain actually changes. The enum domain column stays **v2.1 Dependency 4** (own migration, not this PR).
- Reverted the `UserServiceImpl` order-entry change (flooded dropdowns past role/lab-unit scoping; its alphabetical sort defeated the FRS Display Order feature).
- Create restores legacy **inactive-until-configured**; rename updates the existing localization in place instead of orphaning it.
- REST shape: `POST /sample-types/update` → **`PUT /sample-types/{id}`**; removed the unused `POST /sample-types` (it skipped the role-module wiring the legacy create flow performs — creation stays on `/rest/SampleTypeCreate`).
- en.json: restored the blanked/changed unrelated keys and pruned 63 keys orphaned by the deleted components. `.gitignore` was already clean.

**Verified:** backend compiles; security tests 8/8; systemuser rename tests + sample-type suites green (124 tests); full frontend suite 806/0; vite build clean; spotless + prettier applied.

Remaining for a future story (out of this PR's scope per v2.1): the enum-valued domain migration (Dependency 4), and the Display Order / Disposal editor sections.

mozzy11 (approved):
All review feedback from both rounds (samuelmale + mozzy11) has been applied and verified:

- Auth gate fixed (`hasRole('ADMIN')`), debug endpoint/logging/`Thread.sleep` removed, standard logging restored
- D-030 domain mapping (H/N↔CLINICAL, E↔ENVIRONMENTAL, A↔VECTOR) round-trips correctly — verified against the live DB
- Inactive-by-default create flow restored; `UserServiceImpl` order-entry scoping reverted
- ~3,400 lines of orphaned/mock frontend code removed; en.json regressions restored
- Liquibase changesets renumbered to 064/065 with MARK_RAN preconditions — verified applying cleanly
- Security tests extended to the mutating + terminology endpoints
- Full UI verification against FRS v2.1 passed (list/create/edit/terminology/read-only associated tests); remaining FRS items (Display Order, Disposal, enum domain migration) are declared follow-ups in the description

PR #3656, opened 2026-05-26
Test:Add-shipment-service-Test

Agaba-derrick:
# Pull Requests Requirements

- [ ] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary

Adds integration test coverage for `ShipmentService` covering 8 of 10 methods. Two tests — `createShipment` and `updateShipment` — are skipped due to production bugs that will be resolved in separate issues.

## Tests Added

- 8 passing integration tests for `ShipmentService`
- Test class follows project standards (`BaseWebContextSensitiveTest`, DBUnit XML fixture, Testcontainers PostgreSQL)

## Known Issues (tracked separately)

- `createShipment_shouldGenerateIdAndDefaultToPending` skipped pending fix in #3654
- `updateShipment_shouldModifyExistingFields`  skipped pending fix in #3653
## Screenshots
Screenshot from 2026-05-26 23-35-23

## Related Issue
Closes https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3655

## Other
@HerbertYiga  @mherman22

PR #3657, opened 2026-05-26
Add a JPA-mapped field to Shipment.java

### LINKED ISSUE #3653: sys_user_id` NOT NULL constraint always violated on shipment creation
The `shipment` database table defines `sys_user_id` as a `NOT NULL` column (see Liquibase changeset `shipment-002-create-shipment-table`). However, the `Shipment` JPA entity has no corresponding mapped field for this column.

`BaseObject` (the superclass) declares `sysUserId` but marks it `@Transient`:

```java
// BaseObject.java
@Transient
private String sysUserId;
```

`@Transient` means Hibernate never includes this field in any SQL statement. As a result, every `INSERT INTO shipment (...)` omits `sys_user_id`, violating the `NOT NULL` constraint.

### Steps to Reproduce

Call `ShipmentService.createShipment(shipment)` with any valid `Shipment` object.

### Error

```
PSQLException: ERROR: null value in column "sys_user_id" of relation "shipment"
violates not-null constraint
```

### Root Cause

`Shipment.java` has no `sys_user_id` JPA column mapping. Because `BaseObject.sysUserId` is `@Transient`, Hibernate silently omits the value from all INSERT statements.

### Affected Files

- `src/main/java/org/openelisglobal/shipment/valueholder/Shipment.java`
- `src/main/resources/liquibase/3.5.x.x/shipment-002-create-shipment-table.xml`

### Recommended Fix

Add a JPA-mapped field to `Shipment.java`, following the pattern already used in `ShippingBox.java`:

```java
// Add to Shipment.java
@Column(name = "sys_user_id", nullable = false)
private Integer systemUserId;

public Integer getSystemUserId() { return systemUserId; }
public void setSystemUserId(Integer systemUserId) { this.systemUserId = systemUserId; }
```

Then populate `systemUserId` in `ShipmentServiceImpl.createShipment()` from the caller-supplied `sysUserId`.

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
Fixes a production bug where every call to `ShipmentService.createShipment()` fails with a NOT NULL constraint violation on `sys_user_id`.

The `Shipment` entity had no JPA mapping for this column. Since `BaseObject.sysUserId` is `@Transient`, Hibernate excluded it from every INSERT, causing the constraint to fire on every create.

## Changes

- `Shipment.java`  added `systemUserId` field with `@Column(name = "sys_user_id")` mapping, following the same pattern used in `ShippingBox.java`
- `ShipmentServiceImpl.java`  populated `systemUserId` from the audit context value before INSERT

## Testing

The integration test `createShipment_shouldGenerateIdAndDefaultToPending` is currently `@Ignored` on the `Add-shipment-service-Test` branch. Once this PR merges, that branch will be rebased and the test re-enabled.
## Related Issue
Closes https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3653


src/main/java/org/openelisglobal/shipment/service/ShipmentServiceImpl.java
```diff
@@ -87,6 +87,12 @@ public Shipment createShipment(Shipment shipment) {
             if (shipment.getStatus() == null) {
                 shipment.setStatus(ShipmentStatus.PENDING);
             }
+            // Persist the audit user to the mapped DB column.
+            // sysUserId (from BaseObject) is @Transient and never written by Hibernate.
+            if (shipment.getSystemUserId() == null) {
```
HerbertYiga, line 91: do you also want to  write an integration test to verify field persistence?

pmanko (changes requested):
The mapping fix is right and matches the `ShippingBox`/`BoxSampleItem` pattern — nice root-cause analysis. Two changes before merge:

1. **Drop the `: 1` fallback in `createShipment`.** Everywhere else in this module the user id comes from the controller (`getSysUserId(request)`) with no admin default, and we rejected this same hardcoded-fallback pattern on #3597. For a NOT NULL *audit* column, throw a `LIMSRuntimeException` when no user id is present rather than silently attributing the record to admin. Also wrap the `Integer.parseInt` so a malformed id fails with a clear message.

2. **Update-path regression risk:** now that `sys_user_id` is a mapped column, `updateShipment()` merging a detached `Shipment` with null `systemUserId` will write NULL and violate the constraint — a failure mode that didn't exist before this PR. Either copy the existing value forward when null in `updateShipment`, or note it explicitly in the #3653 follow-up.

Also: re @HerbertYiga's question about test coverage — the `@Ignore`'d `createShipment_shouldGenerateIdAndDefaultToPending` test on #3656 covers this and will be un-ignored once this merges. A quick reply in that thread would close it out.

Agaba-derrick:
> The mapping fix is right and matches the `ShippingBox`/`BoxSampleItem` pattern — nice root-cause analysis. Two changes before merge:
> 
>     1. **Drop the `: 1` fallback in `createShipment`.** Everywhere else in this module the user id comes from the controller (`getSysUserId(request)`) with no admin default, and we rejected this same hardcoded-fallback pattern on [fix alert notificationservice test #3597](https://github.com/DIGI-UW/OpenELIS-Global-2/pull/3597). For a NOT NULL _audit_ column, throw a `LIMSRuntimeException` when no user id is present rather than silently attributing the record to admin. Also wrap the `Integer.parseInt` so a malformed id fails with a clear message.
> 
>     2. **Update-path regression risk:** now that `sys_user_id` is a mapped column, `updateShipment()` merging a detached `Shipment` with null `systemUserId` will write NULL and violate the constraint — a failure mode that didn't exist before this PR. Either copy the existing value forward when null in `updateShipment`, or note it explicitly in the [sys_user_id` NOT NULL constraint always violated on shipment creation #3653](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3653) follow-up.
> 
> 
> Also: re @HerbertYiga's question about test coverage — the `@Ignore`'d `createShipment_shouldGenerateIdAndDefaultToPending` test on #3656 covers this and will be un-ignored once this merges. A quick reply in that thread would close it out.

makes alot of sense !

Agaba-derrick:
@HerbertYiga @pmanko  The @Ignore'd createShipment_shouldGenerateIdAndDefaultToPending test on PR #3656 actually covers this field persistence! It will be un-ignored once this PR merges to keep everything clean.

mozzy11 (commented):
Great debugging, Derrick — you traced this all the way to BaseObject.sysUserId being @Transient, which is exactly right, and following the ShippingBox pattern is the correct call. I checked develop today and the bug is still live, so this fix is very much wanted.

Two things before we merge: (1) the E2E checkpoint failed on the June run — please rebase on latest develop and re-run so we get a green signal; (2) in updateShipment, the extra fetch to backfill systemUserId loads a second managed copy of the entity — can you double-check that the subsequent DAO update doesn't trip on the duplicate persistence-context entry, or consider a targeted query for just the sys_user_id value? Once those are sorted this is good to go.

mozzy11:
Sweep note: holding further review while the existing CHANGES_REQUESTED is open. Also flagging the sequencing question raised on #3656 — these two are coupled (the test PR @Ignores the exact bugs this PR fixes), so please state on both whether they merge together or in which order, and lift the @Ignores in whichever change lands second.

Agaba-derrick:
> Sweep note: holding further review while the existing CHANGES_REQUESTED is open. Also flagging the sequencing question raised on #3656 — these two are coupled (the test PR @ignores the exact bugs this PR fixes), so please state on both whether they merge together or in which order, and lift the @ignores in whichever change lands second.

@mozzy11 Since this is a bug, lets merge this first I can always re-edit the **@ignores** annotations in the Pr after we merge this one !

src/main/java/org/openelisglobal/shipment/service/ShipmentServiceImpl.java
```diff
@@ -101,6 +113,15 @@ public Shipment createShipment(Shipment shipment) {
     public Shipment updateShipment(Shipment shipment) {
         try {
             shipment.setLastupdated(new Timestamp(System.currentTimeMillis()));
+
+            if (shipment.getSystemUserId() == null && shipment.getId() != null) {
+                Integer storedUserId = shipmentDAO.findSystemUserIdById(shipment.getId());
+                if (storedUserId == null) {
+                    throw new IllegalArgumentException("Shipment not found with ID: " + shipment.getId());
```
mozzy11, line 120: Minor: `createShipment` throws `LIMSRuntimeException` for the equivalent guard; consider the same here instead of `IllegalArgumentException` so callers handle one exception family for shipment persistence failures.

mozzy11 (commented):
Re-reviewed the current diff against both open review threads — the requested changes look fully addressed now:

- pmanko's item 1: the `: 1` admin fallback is gone; `createShipment` now throws `LIMSRuntimeException` when no user id is present and wraps the `parseInt` with a clear message.
- pmanko's item 2 + my duplicate-persistence-context concern: `updateShipment` backfills a null `systemUserId` via the new scalar projection `findSystemUserIdById`, so no second managed entity is materialised before the merge.

@pmanko this is ready for your re-review — your CHANGES_REQUESTED is the remaining formal blocker, and Derrick has confirmed the merge order (this PR first, then #3656 lifts its `@Ignore`s).

One CI note: the E2E checkpoint is red, but the failure is in the *Shared Build (Fork Image Publish)* stage — the E2E suites themselves never executed. That stage needs a re-run (or a maintainer-triggered run) to get a clean signal before merge; the backend Build+Test and Shared Build jobs are green. One minor inline nit.

PR #3667, opened 2026-05-30
fix: change getReferralsBySampleItemId parameter type from Integer to String

### LINKED ISSUE #3660: Type Mismatch in ReferralDAOImpl Causes Crashes in BoxSampleItemService
There is a parameter type mismatch between `ReferralDAOImpl` and the `SampleItem` entity mapping, which causes `IllegalArgumentException` crashes under Hibernate 6 strict type validation. This prevents `BoxSampleItemService.removeSampleItemFromBox` from completing successfully.

## Steps to Reproduce
1. Attempt to remove a sample item from a shipping box via `BoxSampleItemService.removeSampleItemFromBox(boxSampleItemId, systemUserId)`.
2. The method internally calls `referralDAO.getReferralsBySampleItemId(Integer.parseInt(sampleItem.getId()))`.
3. Hibernate attempts to execute the `ReferralDAO` query.
4. The system crashes with an `IllegalArgumentException` from `QueryParameterBindingValidator`.

## Root Cause
- In `org.openelisglobal.referral.daoimpl.ReferralDAOImpl` (Line ~249), the query parameter is set via:
  ```java
  query.setParameter("sampleItemId", sampleItemId); // sampleItemId is an Integer
  ```
- The HQL query is `WHERE si.id = :sampleItemId`.
- The `si` reference is to the `SampleItem` entity. In `SampleItem.java`, the ID field is defined as a `String` (`extends BaseObject<String>`).
- Hibernate 6 strictly validates parameter types against the entity property types. Because an `Integer` is passed to match a `String` property, Hibernate immediately throws `IllegalArgumentException: Parameter value [X] did not match expected type [java.lang.String (n/a)]`.

## Code References
**BoxSampleItemServiceImpl.java:**
```java
// Passes parsed Integer to DAO
List<Referral> referrals = referralDAO.getReferralsBySampleItemId(Integer.parseInt(sampleItem.getId()));
```

**ReferralDAOImpl.java:**
```java
// Method signature expects Integer
public List<Referral> getReferralsBySampleItemId(Integer sampleItemId) { ... }
```

**SampleItem.java:**
```java
public class SampleItem extends BaseObject<String> implements NoteObject { ... }
```

## Proposed Fix
Update `ReferralDAO.getReferralsBySampleItemId` and its implementation to either accept a `String`, or convert the `Integer` to a `String` before setting the query parameter:
```java
query.setParameter("sampleItemId", String.valueOf(sampleItemId));
```
Additionally, `BoxSampleItemServiceImpl.java` should just pass `sampleItem.getId()` directly instead of parsing it to an integer first.

kavya685:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Problem
`BoxSampleItemService.removeSampleItemFromBox` was crashing with `IllegalArgumentException` under Hibernate 6 due to a type mismatch.

`SampleItem` extends `BaseObject<String>` so its ID is a `String`, but `ReferralDAOImpl.getReferralsBySampleItemId` accepted an `Integer` and bound it to `si.id` which is a `String` field.

Hibernate 6 strictly validates parameter types and throws:
`IllegalArgumentException: Parameter value [X] did not match expected type [java.lang.String]`

## Changes
- `ReferralDAO.java` — updated interface signature from `Integer` to `String`
- `ReferralDAOImpl.java` — updated implementation signature from `Integer` to `String`
- `BoxSampleItemServiceImpl.java` — removed unnecessary `Integer.parseInt()` and pass `sampleItemId`/`sampleItem.getId()` directly

## Related Issue
Fixes #3660

### DIFF
--- a/src/test/resources/testdata/box_sample_item_service.xml
+++ b/src/test/resources/testdata/box_sample_item_service.xml
@@ -113,6 +113,10 @@
         description="Confirmation requested"
         display_key="referral.reason.confirmation"
         lastupdated="2024-01-01 10:00:00" />
+    <referral_reason id="2" name="New Case"
+        description="Initial consultation"
+        display_key="referral.reason.newcase"
+        lastupdated="2024-07-02 09:15:00" />
 
     <!-- Referral for sample_item 1, already assigned to shipping_box 1 (mirrors
         the state that removeSampleItemFromBox must correctly unassign) -->
--- a/src/test/resources/testdata/box_sample_item_service.xml
+++ b/src/test/resources/testdata/box_sample_item_service.xml
@@ -92,4 +92,35 @@
         reception_notes="Arrived in good condition" sys_user_id="1"
         lastupdated="2024-01-01 11:00:00" />
 
+    <!-- Referral fixtures: sample_item 1 (already in BOX-001 via box_sample_item
+        100) has a referral assigned to that same box, for remove/unassign testing -->
+    <test_section id="1" name="Chemistry"
+        description="Chemistry Section" name_localization_id="1"
+        lastupdated="2024-01-01 10:00:00" />
+    <test id="1" description="Blood Chemistry Test" name="Blood Chem"
+        guid="TEST-BLOOD-CHEM-001" test_section_id="1"
+        lastupdated="2024-01-01 10:00:00" />
+    <analysis id="1" sampitem_id="1" test_sect_id="1" test_id="1"
+        revision="1" status="1" status_id="1" is_reportable="Y"
+        entry_date="2024-01-01 10:00:00" analysis_type="ROUTINE"
+        lastupdated="2024-01-01 10:00:00" />
+
+    <referral_type id="1" name="Confirmation"
+        description="Sent out to confirm result"
+        display_key="referral.type.confirmation"
+        lastupdated="2024-01-01 10:00:00" />
+    <referral_reason id="1" name="Confirmation requested"
+        description="Confirmation requested"
+        display_key="referral.reason.confirmation"
+        lastupdated="2024-01-01 10:00:00" />
+
+    <!-- Referral for sample_item 1, already assigned to shipping_box 1 (mirrors
+        the state that removeSampleItemFromBox must correctly unassign) -->
+    <referral id="1" analysis_id="1" organization_id="1"
+        organization_name="Test Facility" referral_reason_id="1"
+        referral_type_id="1" requester_name="Doctor" status="SENT"
+        assigned_to_box_id="1" referral_request_date="2024-01-01 10:00:00"
+        lastUpdated="2024-01-01 10:00:00" canceled="false"
+        fhir_uuid="c3d4e5f6-a7b8-9012-cdef-123456789012" />
+
 </dataset>
--- a/src/main/java/org/openelisglobal/referral/dao/ReferralDAO.java
+++ b/src/main/java/org/openelisglobal/referral/dao/ReferralDAO.java
@@ -71,7 +71,7 @@ public interface ReferralDAO extends BaseDAO<Referral, String> {
      * @param sampleItemId - the PK of a SampleItem
      * @return list of referrals for this sample item
      */
-    public List<Referral> getReferralsBySampleItemId(Integer sampleItemId);
+    public List<Referral> getReferralsBySampleItemId(String sampleItemId);
 
     /**
      * Get all unassigned referrals grouped by sample item. Returns distinct sample
--- a/src/main/java/org/openelisglobal/referral/daoimpl/ReferralDAOImpl.java
+++ b/src/main/java/org/openelisglobal/referral/daoimpl/ReferralDAOImpl.java
@@ -238,7 +238,7 @@ public class ReferralDAOImpl extends BaseDAOImpl<Referral, String> implements Re
 
     @Transactional(readOnly = true)
     @Override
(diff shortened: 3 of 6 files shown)


kavya685:
@Agaba-derrick, Found the same bug in UnassignedSampleItemServiceImpl.java — also calling referralDAO.getReferralsBySampleItemId(sampleItemIdInt) with an Integer. Should I fix this in this PR?

mozzy11 (commented):
Good catch, @kavya685 — I verified on current develop that getReferralsBySampleItemId still binds an Integer against the String si.id, so the Hibernate 6 IllegalArgumentException in BoxSampleItemService.removeSampleItemFromBox is live and this fix is exactly right. Two next steps: rebase onto develop to freshen the branch, and if possible add a small integration test exercising removeSampleItemFromBox on a sample item with a referral. I'll approve the CI run; this is close to merge-ready.

kavya685:
> Good catch, @kavya685 — I verified on current develop that getReferralsBySampleItemId still binds an Integer against the String si.id, so the Hibernate 6 IllegalArgumentException in BoxSampleItemService.removeSampleItemFromBox is live and this fix is exactly right. Two next steps: rebase onto develop to freshen the branch, and if possible add a small integration test exercising removeSampleItemFromBox on a sample item with a referral. I'll approve the CI run; this is close to merge-ready.

Rebased onto latest develop and added the integration test for removeSampleItemFromBox with a referral — covers both the unassign-on-removal case and a check that other boxes' referrals aren't touched.

Also had to fix UnassignedSampleItemServiceImpl.java — same Integer/String mismatch, it stopped compiling once the DAO signature changed so it's part of this PR now too.

src/test/resources/testdata/box_sample_item_service.xml
```diff
@@ -92,4 +92,35 @@
...
+    <analysis id="1" sampitem_id="1" test_sect_id="1" test_id="1"
+        revision="1" status="1" status_id="1" is_reportable="Y"
+        entry_date="2024-01-01 10:00:00" analysis_type="ROUTINE"
+        lastupdated="2024-01-01 10:00:00" />
+
+    <referral_type id="1" name="Confirmation"
+        description="Sent out to confirm result"
+        display_key="referral.type.confirmation"
+        lastupdated="2024-01-01 10:00:00" />
+    <referral_reason id="1" name="Confirmation requested"
```
mozzy11, line 112: Declaring `referral_reason` here TRUNCATEs the table and leaves only id=1 — `referral-item.xml` (used by `ReferralItemServiceTest`) inserts referrals with `referral_reason_id="2"` without declaring the table, so it fails with `referral_reason_fk` after this test runs. Please re-seed the rows later tests depend on (at minimum id=2), and audit the newly-declared `test_section`/`test`/`analysis` tables for the same interaction.

mozzy11 (changes requested):
The production fix is exactly right — I re-verified the Integer-vs-String binding is still live on develop, and covering all call sites (including `UnassignedSampleItemServiceImpl`) plus the requested integration tests is what we asked for. However, the CI run I approved failed Build + Test with two errors introduced by the new fixtures, so this needs one more pass:

1. **`ReferralItemServiceTest` — deterministic breakage (details inline).** `executeDataSetWithStateManagement` TRUNCATEs every table a dataset declares. Your fixture declares `referral_reason` with only id=1, wiping the seeded id=2 row that `testdata/referral-item.xml` later FKs to (`referral_reason_fk` violation in its setUp).
2. **`ObservationFacadeTest.createObservation_shouldCreateNewResult` — 201 vs 500** in the same run. Your fixture also newly declares `test_section`, `test`, and `analysis`, so those tables are truncated down to just your id=1 rows for every test that runs afterwards without re-declaring them — please verify this one is the same leakage by running the full backend suite locally (`mvn test`, full run — surefire ordering matters here) before pushing the fix.

Once those fixture interactions are resolved and Build + Test is green, this is merge-ready. Also rebase onto develop while you're at it — the branch is a bit behind.

kavya685:
> The production fix is exactly right — I re-verified the Integer-vs-String binding is still live on develop, and covering all call sites (including `UnassignedSampleItemServiceImpl`) plus the requested integration tests is what we asked for. However, the CI run I approved failed Build + Test with two errors introduced by the new fixtures, so this needs one more pass:
> 
> 1. **`ReferralItemServiceTest` — deterministic breakage (details inline).** `executeDataSetWithStateManagement` TRUNCATEs every table a dataset declares. Your fixture declares `referral_reason` with only id=1, wiping the seeded id=2 row that `testdata/referral-item.xml` later FKs to (`referral_reason_fk` violation in its setUp).
> 2. **`ObservationFacadeTest.createObservation_shouldCreateNewResult` — 201 vs 500** in the same run. Your fixture also newly declares `test_section`, `test`, and `analysis`, so those tables are truncated down to just your id=1 rows for every test that runs afterwards without re-declaring them — please verify this one is the same leakage by running the full backend suite locally (`mvn test`, full run — surefire ordering matters here) before pushing the fix.
> 
> Once those fixture interactions are resolved and Build + Test is green, this is merge-ready. Also rebase onto develop while you're at it — the branch is a bit behind.

Fixed — added the missing referral_reason id=2 row back to box_sample_item_service.xml, since the fixture's truncate was wiping it and breaking the referral tests that depend on it.

Also ran the full suite locally a couple of times post-rebase to check the test_section/test/analysis concern — ObservationFacadeTest passed both times, didn't reproduce the failure. Let me know if it still shows up on your end and I'll dig further.

PR #3675, opened 2026-06-01
Update observation to include performer

Isabirye1515:
Performer wasnt included in the Observation, and added request validation and some tests to verify


mozzy11 (commented):
Nice focused PR — the 500-to-422 error-code fixes are exactly right (client data problems should never look like server crashes), and the volume of servlet-level validation tests you added is great to see.

Two things to address before merge:

- **Null provider:** in transformResultToObservation (FhirTransformServiceImpl), getProviderForSample(...) can return null for samples that were entered without a provider, and the next line calls provider.getFhirUuidAsString(). This method runs in the general result-entry FHIR sync, not just your facade path, so a provider-less sample would now throw an NPE in existing workflows. Please guard it (only add performer when the provider and its fhirUuid exist) and add a test with a sample_human row that has no provider_id.
- **performer vs requester:** the provider on the sample is the *ordering* clinician, while Observation.performer means who *produced* the result. Can you double-check with the team whether this reference belongs in performer, or whether the ordering provider should stay on ServiceRequest.requester only? A one-line note in the PR description on the intended semantics would settle it.

Smaller cleanups: the Localization/Panel setup block is copy-pasted into six tests — pull it into a @Before or helper method; and the status == 400 || status == 422 assertion would be stronger pinned to one expected code.

With the null-guard in, this is a quick approve from me. Good work!

Agaba-derrick:
@Isabirye1515 Are we able to address the reviews here and rebase the branch ?
Can you link the issue you are Solving  on this PR ?

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceImpl.java
```diff
@@ -2125,6 +2125,7 @@ public Observation transformResultToObservation(Result result) {
         Test test = analysis.getTest();
         SampleItem sampleItem = analysis.getSampleItem();
         Patient patient = sampleHumanService.getPatientForSample(sampleItem.getSample());
+        Provider provider = sampleHumanService.getProviderForSample(sampleItem.getSample());
```
mozzy11, line 2128: `sampleHumanService.getProviderForSample(...)` can return `null` — `SampleHumanDAOImpl.getProviderForSample` returns `null` when the `sample_human` row has no `provider_id` (the column is nullable). The unconditional `provider.getFhirUuidAsString()` in the new performer block will then NPE and break Observation transformation for every result on a provider-less sample. Please guard the performer block with a null check (and also skip it when `provider.getFhirUuid()` is null, otherwise you emit a `Practitioner/null` reference).

src/test/resources/testdata/result-facade.xml
```diff
@@ -250,8 +250,6 @@
     <sample_human id="2" provider_id="1" samp_id="2"
```
mozzy11, line 250: This change gives every `sample_human` row a `provider_id`, which removes the no-provider case from the fixture — the exact scenario in which the new performer code NPEs. Please keep (or add) a sample without a provider and a test asserting the Observation still transforms successfully.

src/test/resources/testdata/result-facade.xml
```diff
@@ -152,7 +152,7 @@
         scriptlet_id="1" sort_order="1" is_quantifiable="true"
         is_active="true" is_normal="true" />
     <test_result id="2" test_id="2" result_group="2"
-        flags="FLAG2" tst_rslt_type="N" value="14.5" significant_digits="0"
+        flags="FLAG2" tst_rslt_type="D" value="4" significant_digits="0"
```
mozzy11, line 155: Changing `test_result` id=2 from `tst_rslt_type="N" value="14.5"` to `tst_rslt_type="D" value="4"` alters a fixture shared with `DiagnosticReportFacadeTest`. Why is this needed here? If only one new test needs it, prefer a dedicated dataset over mutating shared test data.

src/test/java/org/openelisglobal/fhir/ObservationFacadeTest.java
```diff
@@ -92,6 +92,260 @@ public void readObservation_shouldReturnSuccess() throws Exception {
...
+                }
+                """.formatted(patientFhirUuid, specimenFhirUuid, analysisFhirUuid);
+
+        request.setContent(createJson.getBytes());
+        MockHttpServletResponse response = new MockHttpServletResponse();
+
+        fhirServlet.service(request, response);
+
+        int statusCode = response.getStatus();
+        assertTrue(statusCode == 400 || statusCode == 422);
```
mozzy11, line 346: `assertTrue(statusCode == 400 || statusCode == 422)` weakens the contract — `validateObservation` throws `UnprocessableEntityException`, so missing `status` should deterministically be 422; please assert the exact code. Also the ~20-line Localization/Panel setup block is duplicated across ~10 tests — worth extracting into a private helper.

mozzy11 (changes requested):
Thanks @Isabirye1515 — adding `performer` to Observations and the new provider-side validation (422 for unresolvable LOINC, required subject/specimen/basedOn/status/code) are the right FHIR semantics, and the facade test coverage of the validation paths is genuinely good.

One **blocking** correctness issue: the performer reference is built unconditionally, but `sample_human.provider_id` is nullable — `getProviderForSample` returns `null` for provider-less samples, so transformation NPEs. The fixture edit that gives every sample a provider hides exactly that case. Details inline.

Not approving yet; happy to re-review once the null-guard + a no-provider test are in.

mozzy11 (commented):
**Quick re-review (summary).** Adds `Observation.performer` (ordering provider) and upgrades unresolvable-LOINC/validation failures from 500 → 422, with servlet validation tests.

- **Blocking:** prior CHANGES_REQUESTED not addressed — `getProviderForSample` can return null, but `provider.getFhirUuidAsString()` is still called unconditionally → NPE / `Practitioner/null` for provider-less samples. Needs a null-guard (emit nothing when `fhirUuid` is null). The new fixture gives *every* `sample_human` a `provider_id`, which hides this case — keep a provider-less sample + a test.
- **Semantics:** `performer` = who produced the result, but a sample's provider is the *ordering* clinician (`ServiceRequest.requester`) — confirm `performer` is the right target vs `requester`.
- **Minor:** pin the status assertion to `422` (not `400 || 422`).
- **Status:** DIRTY (merge conflicts). Next: rebase on develop, add null-guard + no-provider test, re-request review.

mozzy11 (approved):
Merged develop and addressed the open threads on top of Elijah's work:

- performer is now null-guarded (provider and its fhirUuid) so provider-less samples no longer NPE or emit `Practitioner/null`; sample 2 in the fixture stays provider-less and two read tests cover both branches.
- missing-status test asserts 422 exactly; the repeated panel setup is one helper; the unrelated test_result/dictionary fixture edits are reverted.
- ObservationFacadeTest, DiagnosticReportFacadeTest and AnalyzerResultsAcceptHoldIntegrationTest pass locally (26 tests).

PR #3680, opened 2026-06-02
feat: implement Bethesda System for cytology reporting

HerbertYiga:
- Add comprehensive Bethesda System dictionary categories and entries via Liquibase migration
- Include specimen adequacy classifications, epithelial abnormalities, HPV testing results
- Add ASCCP recommendations and clinical management guidelines
- Provide complete English translations for all cytology terminology
- Include Bethesda data dictionary CSV for reference

This implementation follows the standard Bethesda System for reporting cervical cytology results, enabling proper cytopathology workflow and clinical decision support.


src/main/resources/liquibase/2.8.x.x/cytology.xml
```diff
@@ -332,4 +332,1023 @@
...
+        </preConditions>
+        <comment>Create Bethesda System dictionary categories</comment>
+
+        <!-- Specimen Assessment Categories -->
+        <insert tableName="dictionary_category" schemaName="clinlims">
+            <column name="id" valueSequenceNext="dictionary_category_seq"/>
+            <column name="description" value="Cytology Specimen Adequacy"/>
+            <column name="lastupdated" valueComputed="${now}"/>
+            <column name="local_abbrev" value="CYTO_SPEC"/>
+            <column name="name" value="CYTOLOGY_SPECIMEN_ADEQUACY"/>
```
mozzy11, line 352: This category name (and the eleven others in this changeset) isn't registered in `DisplayListService.ListType`, so `/rest/displayList/CYTOLOGY_SPECIMEN_ADEQUACY` can't serve it. The adequacy list the UI actually reads is `cytology_adequacy_satisfactory` / `cytology_adequacy_unsatisfactory`, already created further up this same file at lines 195 and 202 — with no entries. Seeding into those instead makes this work without any Java change.

src/main/resources/liquibase/2.8.x.x/cytology.xml
```diff
@@ -332,4 +332,1023 @@
...
+        <comment>Insert Bethesda System dictionary entries - Part 1 (Specimen Assessment)</comment>
+
+        <!-- Specimen Adequacy -->
+        <insert tableName="dictionary" schemaName="clinlims">
+            <column name="id" valueSequenceNext="dictionary_seq"/>
+            <column name="is_active" value="Y"/>
+            <column name="dict_entry" value="Satisfactory for evaluation"/>
+            <column name="local_abbrev" value="SATISFACTORY"/>
+            <column name="sort_order" value="1"/>
+            <column name="display_key" value="cytology.adequacy.satisfactory"/>
```
mozzy11, line 477: `display_key` is resolved **server-side**: `createDictionaryListForCategory` returns `dictionary.getLocalizedName()`, which falls back to `MessageUtil.getContextualMessage(nameKey)` — the backend bundle. These keys were added only to `frontend/src/languages/en.json`, which the frontend never consults for these values. They need to be in `src/main/resources/languages/message_en.properties` or the entries silently fall through to `dict_entry` and can never be translated.

src/main/resources/liquibase/2.8.x.x/cytology.xml
```diff
@@ -332,4 +332,1023 @@
                                  baseTableName="cytology_report" baseColumnNames="cytology_sample_id"
                                  referencedTableName="cytology_sample" referencedColumnNames="id"/>
     </changeSet>
+
+    <!-- Bethesda System Dictionary Implementation -->
+    <changeSet author="openelis" id="13">
```
mozzy11, line 337: Please move these changesets to a new file under `src/main/resources/liquibase/3.5.x.x/` (next free number is `068`) and include it from `base.xml`, rather than appending to a `2.8.x.x` changelog. Bare sequential ids like `13` in a shared legacy file are also collision-prone if anyone else appends to it.

Cytology_Bethesda_DataDictionary.csv
```diff
@@ -0,0 +1,83 @@
+dictionary_category,code,display_name,display_order,description,risk_level,bethesda_section
+CYTOLOGY_SPECIMEN_ADEQUACY,SATISFACTORY,Satisfactory for evaluation,1,Specimen is adequate for cytologic interpretation,,"ADEQUACY"
```
mozzy11, line 2: This duplicates the changeset contents as a second source of truth at the repo root, where it will drift silently. If it's meant as the authoring input, keep it under `src/main/resources/` next to the changeset it generates; otherwise it's better dropped from the PR.

mozzy11 (changes requested):
Thanks for this, Herbert — and I want to be clear up front that the gap you're filling is real. develop creates the cytology `dictionary_category` rows in `2.8.x.x/cytology.xml` but seeds **zero** `dictionary` entries under them, so every Bethesda dropdown in `CytologyCaseView` ships empty today. A curated Bethesda term set is genuinely wanted.

The problem is that the terms are being seeded into a **parallel taxonomy that nothing can read.** develop's cytology UI resolves its lists through `DisplayListService`, which maps `ListType` -> `createDictionaryListForCategory("...")` using the existing lowercase category names: `cytology_adequacy_satisfactory`, `cytology_adequacy_unsatisfactory`, `cytology_epithelial_cell_abnomalit_squamous`, `cytology_epithelial_cell_abnomalit_glandular`, `cytology_non-neoplastic_cellular_variations`, `cytology_reactive_cellular_changes`, `cytology_diagnosis_organisms`, `cytology_diagnosis_other`. This PR creates twelve *new* uppercase categories (`CYTOLOGY_SPECIMEN_ADEQUACY`, `CYTOLOGY_SQUAMOUS_ABNORMALITY`, ...), none of which appear in `ListType`. Net effect: 1,019 lines of data that no endpoint can serve, and the live dropdowns stay empty.

What I'd suggest:
1. **Retarget the entries at the eight existing categories** — adequacy -> `cytology_adequacy_satisfactory`/`_unsatisfactory`, squamous -> `cytology_epithelial_cell_abnomalit_squamous`, glandular -> `..._glandular`, non-neoplastic -> `cytology_non-neoplastic_cellular_variations`, organisms -> `cytology_diagnosis_organisms`, other -> `cytology_diagnosis_other`. Those immediately light up the existing UI with no Java changes.
2. The genuinely new axes (HPV result/genotype/method, risk level, recommendation, hormonal evaluation) have no home yet. Each needs a `ListType` entry + a `createDictionaryListForCategory` registration + UI to consume it — worth splitting into its own PR rather than seeding data ahead of the code.
3. **The i18n keys are in the wrong bundle** (inline below).
4. **Liquibase placement**: new work shouldn't append `changeSet id="13"`.. to the historical `2.8.x.x/cytology.xml`; it belongs in a new numbered file under `3.5.x.x/` wired into `base.xml` (develop is at `067`).
5. `Cytology_Bethesda_DataDictionary.csv` at the repo root is a second source of truth for the same data and will drift. Either generate the changeset from it under `src/main/resources/`, or drop it (it's also missing a trailing newline).

One note so you don't chase a ghost: `Build + Test` is red on a spotless violation in `.specify/oe/skills/overview/reference/canvas-design.md`, which this PR doesn't touch — it's stale base contamination from June and should clear on a rebase.

PR #3688, opened 2026-06-03
feat(analyzer): push identifier_pattern to the bridge for identity corroboration

pmanko:
## Why
The bridge corroborates an inbound message's two identity signals (connection source IP × in-message sender). To do that authoritatively it needs the same regex OE identifies senders with — `Analyzer.identifierPattern`. Without it the bridge falls back to a display-name substring, which false-mismatches ASTM caret-delimited senders.

## What
Pass `Analyzer.getIdentifierPattern()` through `registerTcp` into the bridge registration payload — all three callers (create, startup re-registration, QC-rule registration). No behaviour change in OE itself.

> Stacked on #3629 (`feat/analyzer-order-dispatch`). Pairs with the bridge PR that consumes the pattern. OE2 compiles clean.


pmanko:
Folded into #3629 directly — the `identifier_pattern` plumbing belongs in the analyzer PR it supports, not a satellite stacked on it. Commit 1103632db is now on `feat/analyzer-order-dispatch`.

Issue #3699, opened 2026-06-04
Refactor: Standardise `BaseWebContextSensitiveTest` — Phase 1: Bug Fixes


Agaba-derrick:
`BaseWebContextSensitiveTest` is the root of every integration test in OpenELIS.
It currently contains several bugs and code-quality issues that need to be fixed
before a two-layer hierarchy (Phase 2) can be introduced cleanly.

This issue covers **Phase 1 only**  five safe, self-contained fixes. No existing
subclass should need to change its `extends` declaration.

#### 1. Annotate `setUp()` with `@Before`

`setUp()` currently has no JUnit annotation. `MockMvc` is only initialised in
subclasses that explicitly call `super.setUp()`. Any subclass that forgets this
call gets a silent `null` `MockMvc` at runtime with no indication of why the test
fails.

**Fix:** Add `@Before` so the method is guaranteed to run before every test
automatically, regardless of what the subclass does.

#### 2. Fix the connection leak in `cleanRowsInCurrentConnection`

The current implementation opens a DBUnit `IDatabaseConnection` wrapper around
the raw JDBC connection but only closes the inner JDBC connection inside the
`try-with-resources` block. The outer wrapper is never closed, leaking a connection
back to the pool on every dataset load. In a large test suite running hundreds of
tests this silently exhausts the connection pool.

**Fix:** Remove the DBUnit wrapper entirely from this method. Plain JDBC is
sufficient for truncation and a single `try-with-resources` block closes everything
correctly.

#### 3. Cache `ObjectMapper` instead of recreating per call

Both `mapToJson` and `mapFromJson` instantiate a new
`MappingJackson2HttpMessageConverter` and extract its `ObjectMapper` on every
invocation. Creating an `ObjectMapper` triggers a full Jackson module scan each
time this is documented as expensive in the Jackson docs and is unnecessary when
the configuration never changes between calls.

**Fix:** Declare a single `private static final ObjectMapper` field initialised
once, shared across all calls and all test instances.

#### 4. Replace the magic literal in `buildFhirRequest`

`buildFhirRequest` hardcodes the integer `1` as the system user id despite the
class already defining the constant `TEST_SYS_USER_ID` for exactly this purpose.

**Fix:** Replace the literal with `Integer.parseInt(TEST_SYS_USER_ID)`.

#### 5. Extract `buildDbUnitConnection()` as a private helper

[...]

Issue #3665, opened 2026-05-29
 Refactor: Standardise `BaseWebContextSensitiveTest` — bug fixes and two-layer hierarchy


Agaba-derrick:
`BaseWebContextSensitiveTest` is the root of every integration test in
OpenELIS. It currently mixes DB wiring, dataset loading, security context
management, web layer setup, and JSON helpers all in one class. This makes
it harder to read, harder to maintain, and forces every test — including
simple service and DAO tests — to load the full web context unnecessarily.

This issue proposes a two-phase clean-up modelled on the approach OpenMRS
used when they standardised their own test base classes.

---

### Phase 1 — Fix bugs and clean up the existing class

These changes are safe to land in a single PR. No existing test class needs
to change its `extends` declaration.

#### 1. Annotate `setUp()` with `@Before`

`setUp()` currently has no JUnit annotation. MockMvc is only initialised
in subclasses that explicitly call `super.setUp()`. Any subclass that
forgets this call gets a silent `null` MockMvc at runtime with no
indication of why the test fails.

**Fix:** Add `@Before` so the method is guaranteed to run before every
test automatically, regardless of what the subclass does.

**Risk:** None. Subclasses already calling `super.setUp()` are unaffected.

---

#### 2. Fix connection leak in `cleanRowsInCurrentConnection`

The current implementation opens a DBUnit `IDatabaseConnection` wrapper
around the raw JDBC connection but only closes the inner JDBC connection
inside the try-with-resources block. The outer wrapper is never closed,
leaking a connection back to the pool on every dataset load. In a large
test suite running hundreds of tests this silently exhausts the connection
pool.

**Fix:** Remove the DBUnit wrapper entirely from this method. Plain JDBC
is sufficient for truncation and a single try-with-resources block closes
everything correctly.

**Risk:** None. Identical truncation behaviour, no connection leak.

---

#### 3. Cache `ObjectMapper` instead of recreating per call

Both `mapToJson` and `mapFromJson` instantiate a new
`MappingJackson2HttpMessageConverter` and extract its `ObjectMapper` on
every invocation. Creating an `ObjectMapper` triggers a full Jackson module
scan each time — this is documented as expensive in the Jackson docs and is
unnecessary when the configuration never changes between calls.

**Fix:** Declare a single `private static final ObjectMapper` field
initialised once, shared across all calls and all test instances.

**Risk:** None. Pure performance improvement, identical serialisation
output.

---

[...]

PR #3700, opened 2026-06-04
test: harden test framework with lifecycle, connection and ObjectMapper improvements

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

This PR implements **Phase 1** of the `BaseWebContextSensitiveTest` clean-up proposed in issue #3665.
All five changes are confined to a single file. **No existing subclass needs to change its `extends` declaration.**

### 1. Annotate `setUp()` with `@Before`

`setUp()` previously had no JUnit lifecycle annotation. Any subclass that did not
explicitly call `super.setUp()` received a silent `null` `MockMvc` at runtime with
no indication of why tests failed.

`@Before` guarantees the method runs before every test automatically, regardless
of what the subclass does. Subclasses that already call `super.setUp()` are
unaffected — `@Before` methods chain normally in JUnit 4.

### 2. Fix connection leak in `cleanRowsInCurrentConnection`

The previous implementation opened a DBUnit `IDatabaseConnection` wrapper around
the raw JDBC connection but only closed the inner `Connection` inside the
`try-with-resources` block. The outer wrapper was never closed, leaking a
connection back to the pool on every dataset load. In a large test suite running
hundreds of tests this silently exhausts the connection pool.

Plain JDBC is sufficient for `TRUNCATE`. A single `try-with-resources` block now
covers everything correctly.

### 3. Cache `ObjectMapper` as `private static final`

Both `mapToJson` and `mapFromJson` previously instantiated a new
`MappingJackson2HttpMessageConverter` (and thus triggered a full Jackson module scan)
on every invocation. Jackson documents this as expensive. A single
`private static final ObjectMapper OBJECT_MAPPER` field is now initialised once
in a static block and shared across all calls and all test instances.
Serialisation output is identical.

### 4. Replace magic literal in `buildFhirRequest`

`buildFhirRequest` hardcoded the integer `1` as the system user id despite the class
already defining `TEST_SYS_USER_ID = "1"` for exactly this purpose.
Replaced with `Integer.parseInt(TEST_SYS_USER_ID)` so the intent is visible and
consistent at every call site.

### 5. Extract `buildDbUnitConnection()` as a private helper

The DBUnit connection configuration (`FEATURE_ALLOW_EMPTY_FIELDS`,
`FEATURE_CASE_SENSITIVE_TABLE_NAMES`, `PROPERTY_DATATYPE_FACTORY`) was written
inline inside `executeDataSetWithStateManagement`. Extracted into a focused private
`buildDbUnitConnection()` helper, mirroring the `setupDatabaseConnection()` pattern
from the OpenMRS `BaseContextSensitiveNonTransactionalTest` reference implementation.

Phase 2 (two-layer hierarchy) will be tracked in a separate PR.

## Screenshots

## Related Issue
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3699


mherman22 (commented):
Good work on the five Phase 1 items the direction is right. Leaving a few inline comments on things to tighten up before merge, mainly around how `executeDataSetWithStateManagement` handles its database connection.

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -166,11 +186,7 @@ protected void executeDataSetWithStateManagement(String datasetFileName) throws
         InputStream inputStream = null;
 
         try {
-            connection = new DatabaseConnection(dataSource.getConnection());
-            DatabaseConfig config = connection.getConfig();
-            config.setProperty(DatabaseConfig.FEATURE_ALLOW_EMPTY_FIELDS, true);
-            config.setProperty(DatabaseConfig.FEATURE_CASE_SENSITIVE_TABLE_NAMES, true);
-            config.setProperty(DatabaseConfig.PROPERTY_DATATYPE_FACTORY, new PostgresqlDataTypeFactory());
+            connection = buildDbUnitConnection();
```
mherman22: The TRUNCATE and REFRESH run on separate connections/transactions. If REFRESH fails, the truncation can't be rolled back and the next test gets an empty database. Use a single `jdbcConn` with `setAutoCommit(false)` / `commit()` / `rollback()` for both operations.

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -209,15 +225,45 @@ protected void executeDataSetWithStateManagement(String datasetFileName) throws
...
-     * @param tableNames The names of the tables to truncate.
-     * @throws SQLException If an error occurs during truncation.
+     * @return a fully configured {@link IDatabaseConnection}
+     * @throws DatabaseUnitException if DBUnit fails to wrap the JDBC connection
+     * @throws SQLException          if the underlying data-source fails to open a
+     *                               connection
      */
-    protected void cleanRowsInCurrentConnection(String[] tableNames) throws SQLException, DatabaseUnitException {
+    private IDatabaseConnection buildDbUnitConnection() throws DatabaseUnitException, SQLException {
         IDatabaseConnection connection = new DatabaseConnection(dataSource.getConnection());
```
mherman22: This opens its own connection instead of accepting the caller's. Change the signature to `buildDbUnitConnection(Connection jdbcConn)` so the caller can share its transactional connection with this method.

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -209,15 +225,45 @@ protected void executeDataSetWithStateManagement(String datasetFileName) throws
...
+     *
+     * <p>
+     * Plain JDBC is all that is needed for {@code TRUNCATE}. A single
+     * try-with-resources block now closes everything correctly.
+     *
+     * @param tableNames the names of the tables to truncate
+     * @throws SQLException if an error occurs during truncation
+     */
+    protected void cleanRowsInCurrentConnection(String[] tableNames) throws SQLException {
+        try (Connection conn = dataSource.getConnection(); Statement stmt = conn.createStatement()) {
```
mherman22: Extract the truncation loop into a private `truncateTablesInConnection(Connection, String[])` helper. Then both `executeDataSetWithStateManagement` (inside the transaction) and this method (ad-hoc cleanup) can share the same logic.

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -209,15 +225,45 @@ protected void executeDataSetWithStateManagement(String datasetFileName) throws
...
             for (String tableName : tableNames) {
                 String truncateQuery = "TRUNCATE TABLE " + tableName + " RESTART IDENTITY CASCADE";
                 logger.info("Truncating table: {}", tableName);
```
mherman22: This prints one line per table for every fixture load, adding hundreds of lines of noise to every CI run. Change it to `logger.debug`.

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -96,6 +96,22 @@ public abstract class BaseWebContextSensitiveTest extends AbstractTransactionalJ
...
     protected MockMvc mockMvc;
 
+    /**
+     * Shared {@link ObjectMapper} instance, initialised once from the Spring MVC
+     * message-converter stack. Creating a new
+     * {@code MappingJackson2HttpMessageConverter} (and thus triggering a full
+     * Jackson module scan) on every {@link #mapToJson} / {@link #mapFromJson} call
+     * is documented as expensive in the Jackson docs and is unnecessary when the
+     * configuration never changes between calls.
+     */
```
mherman22, line 104:
```suggestion
   /** 
    * Reuses a shared {@link ObjectMapper} to avoid expensive repeated jackson init 
    */
```

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -107,7 +123,16 @@ public void clearTestAuthentication() {
...
 
-    protected void setUp() throws Exception {
+    /**
+     * Initialises {@link MockMvc} for the full web application context. Annotated
+     * with {@code @Before} so the method is guaranteed to run before every test
+     * automatically, regardless of whether a subclass calls {@code super.setUp()}.
+     * Previously this method had no JUnit annotation, which meant subclasses that
+     * forgot the {@code super} call silently received a {@code null} MockMvc at
+     * runtime.
+     */
```
mherman22, line 127:
```suggestion
    /**
     * Initializes MockMvc before each test to prevent null instances when
     * subclasses omit super.setUp().         
     */
```

src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -162,70 +176,106 @@ protected void executeDataSetWithStateManagement(String datasetFileName) throws
...
+                logger.debug("Truncating table: {}", tableName);
             }
         }
     }
 
+    /**
+     * Ad-hoc truncation helper for tests that need to clean specific tables outside
+     * of a fixture load. Tables in {@link #PROTECTED_SEED_TABLES} are silently
+     * skipped so the Liquibase seed cannot be wiped. Delegates to
+     * {@link #truncateTablesInConnection(Connection, String[])}.
```
mherman22, line 264:
```suggestion
    /**
     * Truncates specified test tables and protected Liquibase seed data.
```

mherman22 (changes requested):

HerbertYiga:
thanks @Agaba-derrick the major thing here is handling the Connection leaks  for the db which you have handled and now the connection should be auto closed, @pmanko we can merge off this

mherman22 (approved):

mherman22:
@Agaba-derrick are you able to make a followup fix for the failing tests on ci?

Agaba-derrick:
@mherman22 thanks for the review, I'm on it !

PR #3644, opened 2026-05-22
Add SampleTypeRequestServiceTest

Agaba-derrick:
# Pull Requests Requirements

- [x] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [x] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [x] The changes include tests or are validated by existing tests.
- [x] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary
Adds comprehensive integration tests for `SampleTypeRequestService` and `SampleTypeRequestServiceImpl` using a real Spring context and database-backed fixtures.

### What was added

- Full CRUD integration test coverage for `SampleTypeRequest`
- Tests for request lifecycle transitions:
  - REQUESTED
  - COLLECTED
  - CANCELLED
- Coverage for:
  - request fulfillment
  - cancellation workflows
  - ordering and filtering queries
  - persistence operations
  - invalid state handling
  - exception scenarios
- Fixture validation tests for dataset consistency and ordering behavior
- Real database interaction testing with DBUnit datasets and Spring-managed transactions
- Verification of entity relationships between:
  - `Sample`
  - `SampleTypeRequest`
  - `TypeOfSample`
  - `SampleItem`

### Additional improvements

- Added reusable helper methods to reduce test duplication
- Expanded negative-path and edge-case coverage
- Ensured tests run without mocks to validate real Hibernate mappings and persistence behavior
## Screenshots
Screenshot from 2026-05-22 15-51-46

## Related Issue
Closes #3643

PR #3701, opened 2026-06-04
issue-3642: Bind sampleId as String in SampleTypeRequestDAOImpl

### LINKED ISSUE #3642: fix(sample-type-request): DAO binds Integer sampleId but Sample.id is String in HQL
`SampleTypeRequestDAOImpl` parses `sampleId` as `Integer` and binds it to HQL parameter `:sampleId` in queries against `str.sample.id`. In OpenELIS, `Sample.id` is mapped as **`String`** (`org.openelisglobal.sample.valueholder.Sample`). Hibernate therefore rejects the parameter type at query execution time.

This breaks all sample-scoped list queries for the decoupled order workflow (OGC-356 / `sample_type_request` table).

## Affected code

- `src/main/java/org/openelisglobal/sampletyperequest/daoimpl/SampleTypeRequestDAOImpl.java`
  - `getRequestsBySampleId(String sampleId)`
  - `getPendingRequestsBySampleId(String sampleId)`
  - `getFulfilledRequestsBySampleId(String sampleId)`

Called from `SampleTypeRequestServiceImpl` and REST:

- `GET /rest/sample-type-requests/sample/{sampleId}`
- `GET /rest/sample-type-requests/sample/{sampleId}/pending`

(`getFulfilledRequestsBySampleId` is implemented in the service/DAO but not yet exposed on the controller.)

## Steps to reproduce

1. Start the app with Liquibase changeset `034-create-sample-type-request-table` applied.
2. Ensure at least one row exists in `clinlims.sample_type_request` for a valid `sample_id`.
3. Call the service or REST with a normal numeric sample id string, e.g. `"2"`:

```java
sampleTypeRequestService.getRequestsBySampleId("2");
```

Or:

```http
GET /rest/sample-type-requests/sample/2
```

## Actual behavior

Runtime failure when Hibernate executes the query, e.g.:

```text
IllegalArgumentException: Parameter value [2] did not match expected type [java.lang.String (n/a)]
```

## Expected behavior

Query returns `SampleTypeRequest` rows for the given sample, ordered by `sortOrder`.

Null, empty, or whitespace-only `sampleId` should continue to return an empty list (current early-return in `parseSampleId` is fine).

## Root cause

```java
// Current (wrong for HQL str.sample.id)
Integer id = Integer.valueOf(sampleId.trim());
query.setParameter("sampleId", id);

// Sample entity
public String getId() { return id; }
```

HQL compares `str.sample.id` (String) to an Integer bind value.

## Proposed fix

Bind `sampleId` as **String** (trimmed), not Integer. For example:

```java
private String normalizeSampleId(String sampleId) {
    if (sampleId == null || sampleId.trim().isEmpty()) {
        return null;
    }
    return sampleId.trim();
}

// In each query method:
String id = normalizeSampleId(sampleId);
if (id == null) {
    return Collections.emptyList();
}
query.setParameter("sampleId", id);
```

No API signature changes required.

## Verification

After the fix:

1. Add or extend integration tests in `SampleTypeRequestServiceTest` that call `getRequestsBySampleId`, `getPendingRequestsBySampleId`, and `getFulfilledRequestsBySampleId` with fixture sample ids (tests were intentionally limited to null/blank inputs until this bug is fixed).
2. Manually hit the REST endpoints above with a sample that has requests.

## Context

Discovered while adding integration tests for `SampleTypeRequestService` (`SampleTypeRequestServiceTest`). `fulfillRequest` / `cancelRequest` / CRUD paths work because they use request primary key (`Integer`), not sample-id HQL.

COLLINS-WAGIMA256:
Closes #3642.

  `parseSampleId` converted `sampleId` to an Integer and bound it to the HQL
  parameter for `str.sample.id`. `Sample.id` is a String (Sample →
  EnumValueItemImpl → BaseObject<String>), so Hibernate 6's strict parameter-type
  validation rejected it with "Parameter value [..] did not match expected type
  [java.lang.String]", breaking getRequestsBySampleId, getPendingRequestsBySampleId
  and getFulfilledRequestsBySampleId.

  The helper (now `normalizeSampleId`) trims/null-checks the id and binds it as a
  String, matching the solution proposed in the issue. No API signatures change.

  Pairs with #3644 (@Agaba-derrick's SampleTypeRequestServiceTest) — this fix makes
  those integration tests pass. Verified locally that the project compiles and
  SampleTypeRequestRestControllerTest (21 tests) passes; DB-level coverage comes
  from #3644 in CI.


src/main/java/org/openelisglobal/sampletyperequest/daoimpl/SampleTypeRequestDAOImpl.java
```diff
@@ -16,20 +16,19 @@ public SampleTypeRequestDAOImpl() {
         super(SampleTypeRequest.class);
     }
 
-    private Integer parseSampleId(String sampleId) {
+    private String normalizeSampleId(String sampleId) {
+        // Sample.id is a String, so the HQL parameter must be bound as a String.
```
Agaba-derrick: can we get rid of these unwanted comments

Agaba-derrick (approved):
LGTM

src/main/java/org/openelisglobal/sampletyperequest/daoimpl/SampleTypeRequestDAOImpl.java
COLLINS-WAGIMA256: All done .Thanks for the feedback

mozzy11 (commented):
Verified against current develop, @COLLINS-WAGIMA256 — parseSampleId still binds an Integer to the String-typed str.sample.id in all three query methods, so this fix is needed and your String-binding normalizeSampleId approach matches what the issue proposed. I like that this deliberately pairs with #3644 for the integration coverage; let's coordinate landing them together (fix first, then the tests prove it in CI). Next steps: rebase to freshen the branch and I'll approve the CI run — with #3644 green on top of this, both should merge quickly.

COLLINS-WAGIMA256:
Thanks @mozzy11, i have rebased onto the latest `develop` to freshen the branch (history is now just the two commits: the `normalizeSampleId` fix + the comment cleanup). Clean and mergeable. Ready to go for the CI approval whenever you're ready.I am also happy to coordinate with #3644 so the fix lands first .

mozzy11 (approved):
Approving — seconding @Agaba-derrick's earlier approval. Re-verified against current develop (blob 03dc5fe83): `parseSampleId` still binds an Integer to the String-typed `str.sample.id` in all three query methods, so the Hibernate 6 type-validation failure from #3642 is still live, and this branch's diff base matches develop exactly (clean rebase, two focused commits). `Sample.id` is a String (`EnumValueItemImpl` -> `BaseObject<String>`), so binding the trimmed String via `normalizeSampleId` is the correct fix; non-numeric input now simply yields an empty result instead of a silent null-out, and no API signatures change.

Still happy to land this ahead of #3644 so its integration tests go green on top of it.

PR #3710, opened 2026-06-11
Fix:fhir qc and storage dbunit errors

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
This PR resolves two distinct integration testing issues: test data isolation failures in the FHIR QC pipeline, and CI-specific execution order failures in the Storage location tests.

### 1. FHIR QC Pipeline Test Refactoring
* **The Problem**: `FhirQCPipelineIntegrationTest` was experiencing test failures because the `AnalyzerTestNameCache` (an in-memory singleton cache) retained state from previous tests. It failed to map the test codes, marking valid incoming QC FHIR bundles as `readOnly` and dropping them. Additionally, the test was manually injecting control lots via Java code rather than relying entirely on DBUnit XML datasets, which goes against OpenELIS test standards.
* **The Fix**: 
  * Forced a `reloadCache()` execution in the test `setUp()` method immediately after DBUnit loads the XML, ensuring the cache always syncs with the fresh database state.
  * Removed the dynamic `insertEstablishmentLot` Java helper method.
  * Moved the runtime test control lots directly into `testdata/fhir-qc-pipeline.xml` as static rows.
  * Standardized the test approach to exactly match `MenuServiceTest` and `PatientServiceTest` while fully preserving all test assertions.

### 2. Storage Location DBUnit Execution Order Errors
* **The Problem**: `StorageLocationServiceIntegrationTest` and `LabelManagementServiceIntegrationTest` were failing unpredictably on the `develop` CI branch with `DatabaseUnit Exception processing table name='sample'`. This happened because the `storage-location.xml` dataset lacked a `system_user`. It passed locally because prior tests (like Patient tests) seeded the database with a user, but it failed via Foreign Key Constraint violations on CI when run in isolation or a different order.
* **The Fix**: 
  * Inserted the required `<system_user id="1" ... />` row into `testdata/storage-location.xml`.
  * The storage datasets are now self-sufficient and completely isolated from test execution order pollution.

## Changes Made
* `src/test/java/org/openelisglobal/analyzer/integration/FhirQCPipelineIntegrationTest.java`
* `src/test/resources/testdata/fhir-qc-pipeline.xml`
* `src/test/resources/testdata/storage-location.xml`

## Related Issue
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3699https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3699

## Other


mherman22 (approved):
LGTM
cc: @reagan-meant

mozzy11:
Thanks Derrick — your diagnosis here was spot on: the stale AnalyzerTestNameCache and the missing system_user were exactly the causes of these flaky failures, and you found them independently. In the meantime develop caught up with you: PR #3773 added the same reloadCache() call in setUp (line 89 now), and the test-infra hardening added ensureAuditSystemUser() to the base loader, which re-seeds system_user id=1 after every fixture load. So both fixes are already live and merging this would mostly duplicate them.

On the fixture-vs-helper question, I'd actually keep insertEstablishmentLot — putting all four lots in the shared XML loads them for every test in the class, while the helper keeps each scenario isolated. I'll close this one, but please don't read that as anything other than good timing by other PRs — the analysis was correct. If you spot any remaining flakiness in this suite on latest develop, a fresh PR would be very welcome.

Issue #3649, opened 2026-05-26
Change PractitionerProvider Search To Query from OpenELIS


Isabirye1515:
Search operations eg. The PractitionerProvider Search sync result from the externalFhirStore which is not the obliged role of the facade, Build a Search operation that handle search by using OpenELIS native services

joelss-tech:
hello @Isabirye1515 ,does this point to practitioner provider only?  another thing is that  will providers from incoming  orders be searchable? cant we have a hybrid approach on this?: query OpenELIS native services as you mentioned at first, then optionally check the external FHIR store. wont that maintain interoperability while ensuring reliability?

Isabirye1515:
Thanks @joelss-tech , Though this approach has performance issues,The “best” server is the one your team can run, support and afford over time, we are drawing to wards the facade entirely to take up interoperability

All providers r going to be implemeted like this. Every provider persisted in the system will be searchable

joelss-tech:
hello @Isabirye1515 , does this fit this ticket requrements? else i would appreciate your review when you have time.https://github.com/DIGI-UW/OpenELIS-Global-2/pull/3874

PR #3720, opened 2026-06-15
Built practitioner search facade querying the database

Isabirye1515:
# Practitioner FHIR Search Implementation

## Summary

Implemented FHIR Practitioner search support with parameter extraction utilities and search service integration.

### Practitioner Search

Added support for Practitioner search by:

- `identifier`
- `given`
- `family`
- `_id`
- `city`
- `state`
- `zipCode`
- `country`
- `_lastUpdated`

### Search Parameter Refactoring

- Introduced common search parameter handling in `BaseSearchParam`.
- Added reusable parameter extraction methods in `FhirProviderUtils`:
  - `stringValueFromStringAndListParam(...)`
  - `uuidValueFromTokenAndListParam(...)`
  - `dateRangeValueFromDateRangeParam(...)`
- Centralized conversion of HAPI FHIR search parameter types into OpenELIS-compatible values.

### Practitioner Search Service

- Added `PractitionerSearchService`.
- Converts FHIR search parameters into a flat search criteria map.
- Returns all practitioners when no search parameters are provided.
- Supports filtered searches through `ProviderService#getAllMatching(...)`.

### Testing

Added and updated integration tests covering:

- Search without parameters
- Search by family name
- Search by given name
- Search by identifier
- Search with no matching results

Improved assertions to verify:

- HTTP response status
- Bundle resource type
- Bundle type (`searchset`)
- Total result count
- Returned practitioner details

### Cleanup

- Removed temporary debugging statements and console logging used during search troubleshooting.
- Fixed package naming and component scanning issues affecting `PractitionerSearchService`.
- Simplified search parameter extraction logic by moving common functionality into shared utilities.

## Benefits

- Reduces duplication across FHIR search implementations.
- Provides a reusable foundation for implementing search on additional FHIR resources.
- Improves maintainability through centralized parameter extraction.
- Aligns Practitioner search behavior with standard FHIR search conventions.

## Future Enhancements

- Add support for additional FHIR search modifiers and prefixes.
- Extend reusable search infrastructure to Organization, Patient, Location, and other FHIR resources.
- Add sorting and pagination support through the shared search framework.


mozzy11 (commented):
Thanks for this — moving Practitioner search off the remote FHIR store and onto our own database is exactly where we want the facade to go, and the integration tests through the servlet (checking bundle type, totals, and actual practitioner content) are the right way to test this. The reusable BaseSearchParam/SearchParameterMap idea is a good foundation for other resources too.

Before we go further, one coordination item and a few code notes:

**Coordination:** PR #3874 implements the same Practitioner local search and creates the same new classes (PractitionerSearchParams, PractitionerSearchService, FhirConstants). Please sync with @joelss-tech and the maintainers so we land one combined approach — I'd hate for either of you to lose work to a merge conflict.

**Code notes:**
- FhirProviderUtils has two identical methods: StringValueFromStringAndListParam and stringValueFromStringAndListParam. Please delete the capitalized one.
- In PractitionerProvider, the @IncludeParam revIncludes block is commented out. Commented-out code shouldn't ship — either keep supporting _revinclude or remove the block and note the behavior change in the PR description.
- _lastUpdated puts the raw DateRangeParam object into the criteria map, so it gets compared for equality against a timestamp column — that can't work. Either implement a real date-range filter or drop the parameter for now, and add a test either way.
- A non-UUID identifier will make UUID.fromString throw, and the catch-all turns that into a 500. Client input errors should be a 400 (InvalidRequestException) or an empty result.
- city/state/postalCode/country exist in PractitionerSearchParams but the provider always passes null for them — either wire them up as @OptionalParams or trim them from this PR.
- SearchQueryInclude is unused, and the TypeToken filter in SearchParameterMap.getParameters(String) doesn't do what it looks like — I'd simplify both.

Once the #3874 question is settled and _lastUpdated is fixed or removed, this is close. Nice work overall — the test coverage especially stands out.

mozzy11 (commented):
Hi @Isabirye1515 — thanks for this, and I want to start by saying the **instinct here is exactly right**. You didn't just bolt a query onto `PractitionerProvider`; you reached for a framework — `BaseSearchParam`, `SearchParameterMap`, a dedicated search service, returning `IBundleProvider` so HAPI handles the Bundle protocol. That OpenMRS-FHIR2-style shape is precisely the direction we want for #3649, and the integration tests through the servlet are the right kind of tests. Nice work on the skeleton.

**The bar this needs to clear:** #3649 will eventually convert *all* our ResourceProviders to local search, so the search plumbing must live in **one shared engine**. A per-resource contribution should be limited to (a) a param-mapping declaration and (b) an entity→FHIR transform (which mostly already exists in `FhirTransformService`) — nothing else. Right now the reusable pieces are the extractors and the id/identifier/lastUpdated base, but the actual engine (build map → query → transform loop → bundle) lives inside `PractitionerSearchService`, so a future `LocationSearchService` would copy ~35 of its ~45 lines. Let's fix that while the framework is young.

**Concrete findings, file by file:**

1. **`_lastUpdated` is broken** — `baseSearchParameterMap()` puts the raw HAPI `DateRangeParam` object into the map, and it flows into `providerService.getAllMatching(flatMap)`, where `BaseDAOImpl` builds `criteriaBuilder.equal(path, DateRangeParam)` — an equality check against a HAPI object. Also the entity property is `lastupdated` (lowercase — see `BaseObject`), not `lastUpdated`, so the path doesn't even resolve. Any `?_lastUpdated=` search fails. Date ranges need to become GE/LE criteria (next-steps below).
2. **`SearchParameterMap.getParameters(String)`** — the `TypeToken` filter compares T's raw type against `PropParam.class` (the container, not the payload), so it never filters what you intend — and the method is never called anywhere. I'd delete it.
3. **Silent truncation of multi-values** — all extractors take `.get(0).get(0)`, so `given=John,Jane` silently searches only "John". Either expand Or-lists into IN criteria or reject multi-values with a 400 — silent truncation is the one option we can't ship.
4. **Duplicate extractor** — `StringValueFromStringAndListParam` and `stringValueFromStringAndListParam` in `FhirProviderUtils` are byte-identical; drop the capitalized one.
5. **Malformed UUID → 500** — `UUID.fromString` throws on any non-UUID token, and the provider's `catch (Exception)` converts it to `InternalErrorException` (HTTP 500). Malformed input should be `InvalidRequestException` (400); well-formed-but-unknown should be an empty bundle.
6. **`identifier` conflated with `_id`** — `FhirConstants.IDENTIFIER_SEARCH_HANDLER = "fhirUuid"` makes `?identifier=` a synonym for `?_id=`; real Provider identifiers (externalId, npi) become unsearchable. They need distinct mappings.
7. **Case-sensitive exact match** — `getAllMatching` is `criteriaBuilder.equal`, so `?family=doe` won't match "Doe". FHIR string search defaults to case-insensitive starts-with; `BaseDAO` already has case-insensitive LIKE machinery (`getAllLike`) to build on.
8. **No paging** — `providerService.getAll()` loads and transforms the whole table on every search. `SimpleBundleProvider` gives correct protocol, but a lazy `IBundleProvider` backed by a count query + paged fetch would give DB-level paging to every resource at once.
9. **`_revinclude` regression** — the `@IncludeParam(reverse)` block is commented out, so existing `_revinclude` consumers (`Encounter:participant`, `ServiceRequest:requester`) silently lose data. Until local include support exists, keep routing those through the store or reject explicitly.
10. **Dead params + fragile constructor** — `PractitionerSearchParams` declares name/city/state/postalCode/country the provider never exposes, and the 10-arg positional constructor with six `null`s is fragile; a builder with only-what's-exposed would tighten this.
11. **Missing `@Transactional(readOnly=true)`** on `PractitionerSearchService` — the Person access during transform should happen inside the transaction per our service-layer rule.

**Coordination:** @joelss-tech's #3874 tackles the same provider and creates `PractitionerSearchService` / `PractitionerSearchParams` at *identical file paths*, so the two PRs can't both merge. My proposal: **your architecture becomes the base** — the SearchParams hierarchy, `IBundleProvider`, and package layout stay — and we fold in the two things #3874 got right: correct `lastupdated` GE/LE range handling, and DB-level pagination — but generalized into `BaseDAO` (add GE/LE to `DBComparison` in `BaseDAOImpl.addWhere`, plus a `getCountMatching`) instead of per-entity HQL, so every future resource inherits them.

**Suggested next steps for this PR:**
1. Add GE/LE (and starts-with-ignore-case) to `DBComparison`/`BaseDAOImpl.addWhere`, plus `getCountMatching` — small, isolated, hugely leveraged.
2. Replace the flatMap-of-Object with a typed criterion (propertyPath, comparison, value) so range and string semantics survive the trip to the DAO.
3. Extract the engine into an abstract `BaseFhirSearchService<Entity, Resource>` (`@Transactional`, two hooks: entity service + transform); `PractitionerSearchService` shrinks to ~5 lines.
4. Back it with a lazy `IBundleProvider` (count + paged slice) instead of `getAll()`.
5. Fix UUID 400-vs-500, the identifier mapping, delete the dead TypeToken method and duplicate extractor, restore `_revinclude` handling.

Happy to pair on the `BaseDAO` piece — once that lands, this framework becomes the pattern every other provider converts onto, which is exactly what #3649 needs. Really glad you took the framework route here.

src/main/java/org/openelisglobal/fhir/FhirConstants.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static final String ADDRESS = "address";
+    public static final String FIRST_NAME = "firstName";
+    public static final String LAST_NAME = "lastName";
+    public static final String CITY = "city";
+    public static final String STATE = "state";
+    public static final String POSTAL_CODE = "zipCode";
+    public static final String COUNTRY = "country";
+
+    public static final String IDENTIFIER_SEARCH_HANDLER = ID_PROPERTY;
```
mozzy11, line 50: `IDENTIFIER_SEARCH_HANDLER = ID_PROPERTY` still makes `?identifier=` a synonym for `?_id=` (both hit fhirUuid), so real Provider identifiers (externalId, npi) remain unsearchable — flagged in the previous review and still open.

src/main/java/org/openelisglobal/fhir/FhirConstants.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static final String FIRST_NAME_SEARCH_HANDLER = PERSON + "." + FIRST_NAME;
+
+    public static final String LAST_NAME_SEARCH_HANDLER = PERSON + "." + LAST_NAME;
+
+    public static final String CITY_SEARCH_HANDLER = PERSON + "." + ADDRESS + "." + CITY;
```
mozzy11: `person.address.city` (and state/zipCode/country) can't resolve — Person has no `address` association; those are direct fields on Person. `FhirPropertyResolver` will throw the moment these are wired. Map to `person.city` etc., or drop them with the dead params.

src/main/java/org/openelisglobal/common/fhir/dao/BaseFhirDao.java
```diff
@@ -0,0 +1,975 @@
...
+@Transactional(readOnly = true)
+public abstract class BaseFhirDao {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(BaseFhirDao.class);
+
+    private static final Locale NORMALIZATION_LOCALE = Locale.ROOT;
+    private static final char LIKE_ESCAPE_CHARACTER = '\\';
+    private static final int DEFAULT_MAX_RESULTS = Integer
+            .parseInt(ConfigurationProperties.getInstance().getPropertyValue("page.defaultPageSize"));
+    private static final int MAX_ALLOWED_RESULTS = Integer
```
mozzy11: `DEFAULT_MAX_RESULTS` and `MAX_ALLOWED_RESULTS` are identical (both parse `page.defaultPageSize` = 20), and the static `Integer.parseInt` with no fallback risks `ExceptionInInitializerError` if the property is missing. The max cap should be a separate, larger limit with a safe fallback.

src/main/java/org/openelisglobal/common/fhir/dao/BaseFhirDao.java
```diff
@@ -0,0 +1,975 @@
...
+     *
+     * @param context     query context
+     * @param firstResult zero-based result offset
+     * @param maxResults  maximum number of results
+     * @throws IllegalArgumentException if pagination parameters are invalid
+     */
+    protected <R> List<R> list(FhirCriteriaContext<?, R> context, int firstResult, int maxResults) {
+        validateContext(context);
+        validatePagination(firstResult, maxResults);
+        return executeQuery(context, firstResult, Math.min(maxResults, MAX_ALLOWED_RESULTS));
```
mozzy11: `Math.min(maxResults, MAX_ALLOWED_RESULTS)` silently under-fills pages: with `_count=100`, `getResources(0,100)` yields 20 entities but HAPI's next link starts at offset 100 — records 20–99 are skipped. Cap `_count` at the provider/bundle level so HAPI knows the real page size, instead of truncating inside the DAO.

src/main/java/org/openelisglobal/fhir/providers/PractitionerProvider.java
```diff
@@ -219,30 +223,38 @@ public MethodOutcome delete(@IdParam IdType theId, HttpServletRequest request) {
...
         String methodName = "searchPractitionerBundle";
-        LogEvent.logDebug(this.getClass().getSimpleName(), methodName,
-                "Searching for Practitioners (returning Bundle)");
 
-        try {
+        LogEvent.logDebug(this.getClass().getSimpleName(), methodName, "Searching for Practitioners");
 
-            Bundle bundle = util.forwardSearchToFhirStore(request);
+        try {
+            PractitionerSearchParams params = new PractitionerSearchParams(identifier, name, given, family, null, null,
```
mozzy11: The 10-arg positional constructor passes `null` for city/state/postalCode/country that are never exposed as `@OptionalParam` — the dead-params + fragile-constructor issue from the earlier review (builder suggested). Also `revIncludes` is accepted but never reaches the search service (`_revinclude=ServiceRequest:requester` silently ignored), and `Encounter:participant` was dropped from the allow-list without mention.

src/main/java/org/openelisglobal/fhir/search/bundleProviders/BaseFhirBundleProvider.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * Reads the configured default page size while protecting application startup
+     * from a missing or invalid configuration value.
+     */
+    private static int resolveDefaultPageSize() {
+
+        try {
+            String configuredPageSize = ConfigurationProperties.getInstance().getPropertyValue("default.page.size");
```
mozzy11: Property key `default.page.size` doesn't exist — SystemConfiguration defines `page.defaultPageSize` (which BaseFhirDao reads). This always falls through to the hard-coded 20; use the same key in both places.

src/main/java/org/openelisglobal/fhir/search/bundleProviders/BaseFhirBundleProvider.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        List<E> entities = loadEntities(fromIndex, pageSize);
+
+        if (entities == null || entities.isEmpty()) {
+
+            return Collections.emptyList();
+        }
+
+        return entities.stream().filter(Objects::nonNull).map(this::transformEntity).filter(Objects::nonNull)
```
mozzy11, line 94: `transformEntity` runs when HAPI calls `getResources`, outside the `@Transactional` search service. Provider.person is `lazy="false"` so Practitioner is safe today, but as a *generic* base class this hits LazyInitializationException on the first entity with a lazy association — run load+transform inside one transactional boundary or document the eager-fetch requirement.

mozzy11 (commented):
Reviewed as a draft — this has become the stronger of the two competing practitioner-search implementations (vs #3874): the shared criteria-based `BaseFhirDao` + `FhirPropertyResolver` engine, lazy `BaseFhirBundleProvider` (count query + paged fetch → correct `Bundle.total` and HAPI paging), proper AND/OR semantics, LIKE-escaped case-insensitive string search, real `_lastUpdated` range predicates, and 400-vs-500 UUID handling address most of the earlier review's findings.

Still open before undrafting (inline): `identifier` is still conflated with `_id`; the address params map to a `person.address.*` path that doesn't exist on Person; `_revinclude` declared but ignored; `MAX_ALLOWED_RESULTS` == default page size silently skips records for `_count` > 20; a config key mismatch. CI backend failure looks unrelated (cold-cache spotless on `SECURITY.md`) — a rebase + `mvn spotless:apply` should clear it. Please coordinate with #3874 so the date-range/paging fixes land once, in one PR.

mozzy11 (commented):
Status check after today's develop sync (8408b79): the merge is clean, the full pipeline is now green (backend Build + Test, frontend, E2E), so the earlier spotless failure is resolved. This remains the stronger of the two competing practitioner-search implementations vs #3874.

All five items from my 2026-07-23 inline review are still present in the current diff, so they remain the pre-undraft checklist:
1. `IDENTIFIER_SEARCH_HANDLER = ID_PROPERTY` still conflates `?identifier=` with `?_id=` (FhirConstants.java).
2. The address params still map to a non-existent `person.address.*` path.
3. `_revinclude` is declared but ignored.
4. `MAX_ALLOWED_RESULTS` == default page size silently under-fills pages for `_count` > 20 (BaseFhirDao).
5. `default.page.size` config key mismatch vs `page.defaultPageSize` (BaseFhirBundleProvider).

No new inline comments — the existing threads cover these. Once they're addressed and the search framework consolidation with #3874 is agreed, this looks ready to undraft.

mozzy11 (commented):
Re-reviewed against the current head (`94bfca6`), focusing on whether the items from my 2026-07-23 review are resolved — the branch was reworked on 07-29 (`d4de990`) and 08-04 (`f61205ad`), i.e. after my 07-28 note. Verified by tracing the actual control flow, not just names.

**Resolved (verified):**
- **Address search** now resolves to real `Person` fields — `CITY/STATE/POSTALCODE/COUNTRY_SEARCH_HANDLER` compose to `person.city` / `person.state` / `person.zipCode` / `person.country`, all of which exist on `Person`; the old `person.address.*` (no such association) is gone, so it no longer throws. (`FhirConstants.java:56-62`)
- **Pagination truncation fixed** — `executeQuery` does `setFirstResult/setMaxResults` with the real page size (no `Math.min(_, 20)` cap), and the count uses `countDistinct` while the paged fetch is `distinct`, so `Bundle.total` and HAPI's next-link offsets agree; `_count=100` now returns 100. (`BaseFhirDao.java:940`, `:814`)
- **Config key fixed** to `page.defaultPageSize`, with startup guarded. (`BaseFhirBundleProvider.java:205`)
- **Dead params + revInclude** — `city/state/postalCode/country` are now real `@OptionalParam`s (no more `null` positional args), and `revIncludes` is passed through to `PractitionerSearchParams`. (`PractitionerProvider.java:240-269`)
- CI is green and the search paths have real servlet integration tests (search by none / family / given / identifier / `_id`).

**Still open / follow-ups:**
- **`identifier` search still resolves to `fhirUuid`** (identical to `_id`); `Provider.externalId` and `npi` exist but are excluded — now documented as intentional in `PractitionerSearchDao`. Per FHIR semantics `?identifier=` should match business identifiers, so this remains a real gap even if deliberately deferred; worth a tracked follow-up. (`FhirConstants.java:50`)
- **Transform runs outside a transaction in the generic base.** `BaseFhirBundleProvider` loads entities inside the DAO's `@Transactional` call, then runs `transformEntity` *after* it returns — HAPI calls `getResources` on a plain object once the search-service tx has closed. Safe for Practitioner only because `Provider.person` is `lazy="false"` and the transform reads eager scalars, but as the base for future resources the first subclass whose entity dereferences a genuinely lazy association will hit `LazyInitializationException`. Recommend load+transform within one transaction in the base. (`BaseFhirBundleProvider.java:87-94`)
- **Minor:** `BaseFhirDao:76` static `Integer.parseInt("page.defaultPageSize")` has no fallback (`ExceptionInInitializerError` if the property is ever absent); `resolveDefaultPageSize()` can return `null` → NPE-on-unbox when `page.defaultPageSize <= 0`. And the 15-arg positional `PractitionerSearchParams` constructor is getting fragile — a builder would be safer as params grow.

Overall this is solid, well-structured work on the right architecture (criteria-based DAO + lazy bundle provider + reusable search params) — nothing here is a live bug or regression for the shipped Practitioner path.

mozzy11 (approved):
Approving per my re-review comment above: the substantive items from the July review (address resolution, pagination, config key, dead params/revInclude) are all genuinely resolved in the current head, CI is green, and the search paths carry real integration tests. The remaining points — `identifier` search scoped to fhirUuid (documented as intentional) and the generic-base transform-outside-tx note — are follow-ups, not blockers for the Practitioner path. Nice, well-structured work.

PR #3747, opened 2026-06-18
Test: Result calculation service integration Test

### LINKED ISSUE #3746: Test: Implement Integration Tests for ResultCalculationService
## Objective
Implement a robust integration testing suite for `ResultCalculationService` to validate all core service methods against a real database state using the `BaseWebContextSensitiveTest` framework.

### 1. Test Setup & Seeding
- **DBUnit Dataset**: Use the XML fixture `testdata/result-calculation.xml` to load reference database state.
- **Programmatic Seeding**: Populate the join table `test_operations` (which represents the `@ElementCollection` of tests on `ResultCalculation` and lacks a primary key) programmatically in the test setup via the service layer.

### 2. Entity Management & Hibernate 6 Guidelines
- **Managed Parameters**: When querying by associations, query parameters for entity types (e.g., `Test`, `Calculation`, `Patient`) must be managed/persistent entities retrieved via their respective services (e.g., `TestService.get()`, `TestCalculationService.get()`), rather than transient stub instances, to avoid parameter-binding errors.
- **Eager vs. Lazy Assertions**: Verify direct properties and eagerly loaded `@ManyToOne` associations. Avoid asserting on lazy-loaded collections outside active transaction contexts.

### 3. Coverage Targets
Ensure the following operations are thoroughly covered:
- **CRUD Operations**: Test `getAll()`, `get()`, `getCount()`, `insert()`, `update()`, and `delete()`.
- **Query Methods**: Validate custom query methods:
  - `getResultCalculationByPatientAndTest`
  - `getResultCalculationByTest`
  - `getResultCalculationByPatientAndCalculation`
- **Edge Cases**: Validate behavior with null, non-existent, and ghost parameters.

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

### `ResultCalculationServiceTest.java`
- Added sequence reset to prevent primary key conflicts from hardcoded DBUnit XML IDs
- Added managed entity lookups via `PatientService`, `TestService`, `TestCalculationService`, and `ResultService` to resolve `TransientObjectException` errors
- Added lazy-load-safe assertions and corrected exception expectations to match service layer behaviour
- Covers 28 test cases across CRUD operations, query methods, and null/non-existent edge cases

### `result-calculation.xml`
Added fixtures including test 103 and additional calculations and results to support full edge case coverage.

## Result

All 28 integration tests pass.

## Screenshots
Screenshot from 2026-06-19 01-39-54

## Related Issue
Closes #3746


mozzy11 (commented):
Derrick, this is impressively thorough — 28 tests, every seeded row has a positive lookup plus wrong-patient/null/non-existent negatives, and you found a real bug (getResultCalculationByTest), filed it as #3748, and pinned the current behavior with a link. That find-file-pin workflow is exactly what this testing project is about.

A few cleanups before merge:

- **Use the base-class helper for the sequence reset.** BaseWebContextSensitiveTest now has resyncSequence(sequence, table) which sets the sequence from MAX(id)+1 — please replace the hard-coded setval('result_calculation_seq', 100, false) with it, so the fixture can grow without anyone remembering to bump 100.
- **No commented-out code, please.** Turn the commented block for the intended getResultCalculationByTest behavior into an @Ignore("Blocked by #3748") test — that keeps it compiling and ready to switch on when the bug is fixed.
- **Collapse the 4 duplicate exception pins.** They all exercise the same broken code path; one pin (plus the ignored intended test) tells the same story with less noise.
- The CI run here is from June 19 — please rebase/re-run so we see it green against current develop.

The fixture comment explaining the DBUnit NoPrimaryKeyException workaround was a thoughtful touch. With those cleanups this is a strong merge.

src/test/java/org/openelisglobal/testcalculated/service/ResultCalculationServiceTest.java
```diff
@@ -0,0 +1,373 @@
...
+    private TestService testService;
+
+    @Autowired
+    private PatientService patientService;
+
+    @Autowired
+    private TestCalculationService testCalculationService;
+
+    @Autowired
+    private ResultService resultService;
```
mozzy11: `resultService` is autowired but never used. Remove the field and its import.

src/test/java/org/openelisglobal/testcalculated/service/ResultCalculationServiceTest.java
```diff
@@ -0,0 +1,373 @@
...
+    @Autowired
+    private DataSource dataSource;
+
+    private JdbcTemplate jdbcTemplate;
+
+    @Before
+    public void init() throws Exception {
+        jdbcTemplate = new JdbcTemplate(dataSource);
+        executeDataSetWithStateManagement("testdata/result-calculation.xml");
+        jdbcTemplate.execute("SELECT setval('result_calculation_seq', 100, false)");
```
mozzy11: Use the base-class helper `resyncSequence("result_calculation_seq", "clinlims.result_calculation")` instead of a hand-rolled `setval(...,100,...)` — it syncs to MAX(id)+1 so it can't collide with fixture ids, and removes the JdbcTemplate/DataSource boilerplate.

src/test/java/org/openelisglobal/testcalculated/service/ResultCalculationServiceTest.java
```diff
@@ -0,0 +1,373 @@
...
+        Patient nonExistentPatient = new Patient();
+        nonExistentPatient.setId("9999");
+        List<ResultCalculation> results = resultCalculationService
+                .getResultCalculationByPatientAndTest(nonExistentPatient, labTest("101"));
+
+        Assert.assertTrue(results.isEmpty());
+    }
+
+    @Test
+    public void getResultCalculationByPatientAndTest_rcWithNoTestAssociation_shouldReturnEmpty() {
```
mozzy11: `rcWithNoTestAssociation_shouldReturnEmpty` runs byte-for-byte the same query as `wrongPatientForTest_shouldReturnEmpty` (patient 2 + test 101) — it never touches rc4, the row that actually has no test association. Delete it or make it exercise rc4's scenario for real.

src/test/java/org/openelisglobal/testcalculated/service/ResultCalculationServiceTest.java
```diff
@@ -0,0 +1,373 @@
...
+        Patient patient = patientService.get("2");
+        List<ResultCalculation> results = resultCalculationService.getResultCalculationByPatientAndTest(patient,
+                labTest("101"));
+
+        Assert.assertTrue(results.isEmpty());
+    }
+
+    // BUG: getResultCalculationByTest always throws LIMSRuntimeException due to a
```
mozzy11: Remove the commented-out test block — `@Ignore` (as in your other PRs) preserves an intended-behavior test visibly; commented code rots.

src/test/java/org/openelisglobal/testcalculated/service/ResultCalculationServiceTest.java
```diff
@@ -0,0 +1,373 @@
...
+        resultCalculationService.getResultCalculationByTest(labTest("102"));
+    }
+
+    @Test(expected = LIMSRuntimeException.class)
+    public void getResultCalculationByTest_test103_shouldThrowLIMSRuntimeException() {
+        resultCalculationService.getResultCalculationByTest(labTest("103"));
+    }
+
+    @Test(expected = LIMSRuntimeException.class)
+    public void getResultCalculationByTest_returnedRowsMustHaveAllRelationshipsLoaded() {
```
mozzy11: `returnedRowsMustHaveAllRelationshipsLoaded` asserts nothing about relationships — it duplicates `getResultCalculationByTest_test101_shouldThrowLIMSRuntimeException` under a misleading name. One pinning test for #3748 is enough (the test102/test103 copies add no branch coverage either).

mozzy11 (changes requested):
Thanks @Agaba-derrick — broad, real coverage of `ResultCalculationService` (CRUD, count, query methods, null/non-existent branches), and I appreciate the documented `test_operations` unique-constraint workaround in the fixture. I verified bug #3748 is still on develop (the HQL lacks `select r`), so the `expected = LIMSRuntimeException` tests do pass today.

Main asks before merge: four near-identical tests pin that bug — one (`returnedRowsMustHaveAllRelationshipsLoaded`) with a name that asserts nothing it claims — plus a commented-out block, a duplicated empty-result test, an unused field, and a hand-rolled sequence reset where a base-class helper exists. Also please coordinate fixture naming with your #3886 (`result-calculation.xml` here vs `result-calculation-service.xml` there — the latter actually serves TestCalculationServiceTest) so neither PR clobbers the other on merge. Details inline.

src/test/java/org/openelisglobal/testcalculated/service/ResultCalculationServiceTest.java
```diff
@@ -0,0 +1,293 @@
...
+    private TestService testService;
+
+    @Autowired
+    private PatientService patientService;
+
+    @Autowired
+    private TestCalculationService testCalculationService;
+
+    @Before
+    public void init() throws Exception {
```
mozzy11, line 34: Please restore the two insert tests here and add `resyncSequence("result_calculation_seq", "clinlims.result_calculation")` to `init()` — the helper exists on `BaseWebContextSensitiveTest` and syncs the sequence to MAX(id)+1, so the fixture ids can't collide with generated ones.

mozzy11 (changes requested):
Thanks @Agaba-derrick — the cleanup round landed well: the unused `resultService` field is gone, the commented-out block is now a proper `@Ignore("Blocked by ... #3748")` intended-behavior test, the four duplicate exception pins are collapsed to a single `getResultCalculationByTest_throwsUntilIssue3748Fixed`, the redundant empty-result test was removed, and the fixture was renamed to `result-calculation-integration-test.xml` with a comment coordinating with #3886. CI is green.

One regression from the cleanup keeps this from an approve: the `setval` removal took both insert tests with it (`insert_shouldPersistAndIncrementCount`, `insert_newRowWithTestShouldBeReturnedBySubsequentQuery`), so the service's insert path is now completely uncovered. The ask was to swap the hand-rolled reset for the base-class `resyncSequence(...)` helper, not to drop the tests that needed it. Restore insert coverage using the helper and this is ready to merge.

HerbertYiga (dismissed):

src/test/java/org/openelisglobal/testcalculated/service/ResultCalculationServiceTest.java
HerbertYiga: @Agaba-derrick did you respond to the above request?

Agaba-derrick: @HerbertYiga @mozzy11  I have resolved these comments and addressed the required reviews
Are we able to take another look and see if we can land this one too?

Issue #3725, opened 2026-06-16
Week 4: Migrate Analyzer Forms, Configuration, and Field Mapping to TypeScript


manroop-2005:
## Overview

This week's goal is to complete the remaining TypeScript migration work for the Analyzer feature area by converting the complex analyzer workflows, configuration screens, and field mapping components from JavaScript to TypeScript while preserving existing functionality and UI behavior.

## Target Folders

* `frontend/src/components/analyzers/AnalyzerForm`
* `frontend/src/components/analyzers/AnalyzerTypeManagement`
* `frontend/src/components/analyzers/FieldMapping`
* `frontend/src/components/analyzers/CustomFieldTypes`
* `frontend/src/components/analyzers/SerialConfiguration`
* `frontend/src/components/analyzers/ErrorDashboard`

## Objectives

* Complete migration of remaining analyzer-related UI components to TypeScript.
* Introduce strong typing for analyzer configuration workflows.
* Improve type safety for field mapping and analyzer setup forms.
* Preserve all existing API interactions and validation behavior.

### Analyzer Form Migration

* Convert analyzer form components from `.js/.jsx` to `.ts/.tsx`.
* Add interfaces/types for:

  * Form values
  * Analyzer configuration payloads
  * Validation objects
  * Modal state
  * Callback signatures

### Analyzer Type Management

* Migrate analyzer type management screens to TypeScript.
* Add typing for:

  * Analyzer type records
  * Form submissions
  * CRUD callbacks
  * API responses

### Field Mapping

* Convert field mapping components to TypeScript.
* Add typing for:

  * Mapping rows
  * Analyzer fields
  * LIS fields
  * Mapping payloads
  * Table state and callbacks

### Configuration Modules

* Migrate:

  * File Import Configuration
  * Serial Configuration
  * Custom Field Types

* Type:

  * Default configuration payloads
  * Protocol settings
  * Import settings
  * Configuration forms
  * Validation states

### Error Dashboard

* Convert dashboard components to TypeScript.
* Add typing for:

  * Error records
  * Status values
  * Filters
  * Table rows
  * Dashboard state

## Acceptance Criteria

* All target analyzer folders are migrated to TypeScript.
* Form values, mapping data, configuration payloads, and dashboard records are strongly typed.
* Existing analyzer workflows continue to function without behavioral changes.
* No analyzer API contracts are modified.
* Targeted linting, tests, and TypeScript validation pass or blockers are documented.

[...]

Issue #3646, opened 2026-05-23
Modernizing OpenELIS Frontend with TypeScript and Performance Optimization( parent issue)


manroop-2005:
# Overview

This umbrella issue tracks the modernization of the OpenELIS frontend through incremental TypeScript adoption and targeted performance optimization improvements.

The primary goals of this initiative are to:

- Improve type safety and maintainability
- Reduce runtime errors
- Enhance developer experience and code consistency
- Establish scalable frontend architecture patterns
- Optimize frontend performance and rendering efficiency
- Standardize TypeScript migration practices across modules

This issue will serve as the parent tracking issue for all related migration, refactoring, testing, and optimization tasks.

---

# Child Issues

PR #3705, opened 2026-06-09
Week 3: migrate analyzer components and services to TypeScript

manroop-2005:
# Pull Requests Requirements

- [X] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [X] The PR title follows conventional commit label standards.
- [X] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [X] The changes include tests or are validated by existing tests.
- [X] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary

This PR continues the TypeScript migration effort for the analyzer module by converting analyzer-related services and components from JavaScript to TypeScript while preserving existing functionality, API contracts, and styling.

#### Services

* `frontend/src/services/analyzerService.jsx` → `analyzerService.ts`
* `frontend/src/services/serialService.js` → `serialService.ts`

#### Components

* `frontend/src/components/analyzers/AnalyzersList/AnalyzersList.jsx` → `AnalyzersList.tsx`
* `frontend/src/components/analyzers/DeleteAnalyzerModal/DeleteAnalyzerModal.jsx` → `DeleteAnalyzerModal.tsx`
* `frontend/src/components/analyzers/TestConnectionModal/TestConnectionModal.jsx` → `TestConnectionModal.tsx`

#### Shared Types and Constants

* `frontend/src/components/analyzers/constants.js` → `constants.ts`
* Added `frontend/src/components/analyzers/types.ts`

## Changes Made

* Added TypeScript typings for analyzer-related services.
* Created reusable shared types for analyzers, status values, and API responses.
* Updated analyzer components to use strongly typed props, state, and service responses.
* Centralized shared analyzer type definitions to reduce duplication across components.
* Preserved existing API payload structures and UI behavior.
* Kept existing styling and functionality unchanged.

### Tests

*  Targeted tests passed:

  * `AnalyzersList`
  * `DeleteAnalyzerModal`
  * `TestConnectionModal`

### Type Checking

* Analyzer-related TypeScript errors were resolved.
* `tsc --noEmit` still reports pre-existing unrelated issues in:

  * Playwright tests
  * Admin calculated value tests
  * Patient results viewer
  * Result page file upload
  * `vite.config.ts`

## Related Issue

[...]

PR #3753, opened 2026-06-22
Week 4 GSoC: Complete Analyzer Module TypeScript Migration

manroop-2005:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

# Summary

This PR completes **Week 4** of the GSoC 2026 project **"Modernizing OpenELIS Frontend with TypeScript and Performance Optimization"** by migrating the remaining Analyzer feature modules from JavaScript to TypeScript.

### TypeScript Migration

Converted analyzer-related components from `.js/.jsx` to `.ts/.tsx` across the following areas:

* Analyzer Form
* Analyzer Type Management
* Field Mapping
* Custom Field Types
* Serial Configuration
* Error Dashboard

### Shared Type Definitions

Added a new shared type module:

* `analyzerTypes.ts`

This file centralizes reusable interfaces and types for analyzer configuration, form data, field mappings, serial settings, validation rules, error records, and dashboard state.

### Strong Typing Improvements

Introduced TypeScript interfaces and type definitions for:

* Analyzer form values
* Configuration payloads
* Validation objects
* Modal state
* Callback signatures
* Analyzer type records
* API request/response structures
* Mapping rows and table data
* Analyzer and LIS field definitions
* Serial configuration settings
* Custom field validation rules
* Error dashboard records and filters

### Testing

Added and updated tests for migrated components, including:

* AnalyzerForm
* ValidationRuleEditor
* SerialConfiguration
* ErrorDashboard
* FieldMapping components

These tests help validate the migrated TypeScript implementations while preserving existing behavior.

## Results

* All targeted analyzer folders have been migrated to TypeScript.
* Strong type safety has been introduced across analyzer configuration workflows.
* Existing analyzer functionality and UI behavior remain unchanged.
* No analyzer API contracts were modified.
* Type definitions are now reusable and easier to maintain across analyzer modules.

## Validation

* Targeted component tests pass.
* Linting and formatting checks pass for migrated files.
* Existing project-wide TypeScript issues outside the analyzer scope remain unchanged and are not introduced by this PR.

## Related

* Parent GSoC Project: #3646
* Week 4 Issue: #3725


frontend/src/components/analyzers/SerialConfiguration/SerialConfiguration.tsx
```diff
@@ -216,6 +237,18 @@ const SerialConfiguration = ({
...
   };
 
+  const getNumberInputValue = (
+    event:
+      | React.ChangeEvent<HTMLInputElement>
+      | React.MouseEvent<HTMLButtonElement>
+      | React.KeyboardEvent<HTMLInputElement>
+      | React.FocusEvent<HTMLInputElement>,
+  ): number => {
+    return event.target instanceof HTMLInputElement
```
pmanko: This changes the old Carbon `NumberInput` behavior. The previous code read `e.imaginaryTarget?.value || e.target.value`; Carbon uses `imaginaryTarget` for stepper button changes, so this helper can return `0` when users click increment/decrement. Please preserve `imaginaryTarget?.value ?? target.value` and then coerce safely to a number.

pmanko (changes requested):
Thanks for the Week 4 analyzer migration. The PR is close to the stated goal of finishing the analyzer TypeScript migration while preserving behavior: I verified locally that the analyzer test set passes and `npm run build` succeeds.

Requesting changes for one behavior regression in `SerialConfiguration.tsx`: the migration dropped Carbon `NumberInput` `imaginaryTarget` handling, so stepper interactions can write `0` into baud/data-bit state. Please restore the prior value extraction behavior.

Also note this PR conflicts with #3705 in `frontend/src/components/analyzers/constants.ts`; during rebase, please consolidate the overlapping analyzer type modules rather than keeping both `types.ts` and `analyzerTypes.ts`.

mozzy11 (commented):
Thanks for the Week 4 update — the NumberInput imaginaryTarget regression flagged in the earlier review is now fixed (`state?.value ?? event.imaginaryTarget?.value ?? event.target.value` in SerialConfiguration.tsx), and I want to call out two genuinely good fixes you folded in: the Carbon v11 Checkbox onChange signatures (`(_, { checked })` — the old code was reading the event as the boolean) and the intl.formatMessage defaultMessage corrections in ErrorDashboard.tsx. The new 152-line analyzers/types.ts is also a solid shared foundation.

Remaining items before this can merge:
- **CI: the '03 Checkpoint - E2E' job is failing** — please triage; we can't merge on a red E2E checkpoint.
- **AnalyzerForm.defaultConfigs.test.tsx**: the local `const waitFor = async (cb) => { await Promise.resolve(cb()); }` is not a real waitFor — it runs the assertion once with no retry, so the async assertions no longer actually wait. Please import waitFor from @testing-library/react (your AnalyzerForm.test.tsx already does this correctly).
- **Coordination with #3705**: both PRs create analyzers/types.ts and rename constants.js → constants.ts with different content. I've approved #3705 (smaller and green) to land first — please rebase this one on top and fold its types into your richer types.ts.
- Smaller: AnalyzerForm declares AnalyzerFormProps but ignores them (`_props`) — wire up open/onClose or drop the interface; and `AnalyzerProtocol = ... | string` collapses the union — drop the `| string` to keep autocomplete.
- Please list the intentional UI-adjacent changes (ModalFooter → explicit Buttons, List/ListItem → ul/li, removed required props) in the PR description so reviewers know they're deliberate.

Next steps: fix the test helper, triage the E2E failure, and rebase after #3705 lands — then this is ready.

mozzy11:
This one is blocked three ways: @pmanko's CHANGES_REQUESTED is still open, the branch is **CONFLICTING** with develop, and backend CI fails. Please address the review feedback and rebase; happy to re-review then. The `tsc --noEmit` recommendation from #3884/#3895 applies to this migration too.

frontend/src/components/analyzers/types.ts
```diff
@@ -0,0 +1,152 @@
...
+  type?: string;
+  pluginTypeId?: string;
+  ipAddress?: string | null;
+  port?: number | string | null;
+  importDirectory?: string | null;
+  protocol?: AnalyzerProtocol;
+  protocolVersion?: string | null;
+  communicationMode?: string | null;
+  testUnitIds?: string[];
+  status?: string;
```
mozzy11: Three divergences from develop's `types.ts` to resolve deliberately, not by whichever side wins the merge: `status` and `lifecycleStage` are `string` here but `AnalyzerStatus` on develop (widening loses the union), and `testUnitIds` is `string[]` here but `Array<string | number>` on develop (narrowing may reject existing numeric ids). I'd keep develop's types for these three and yours for everything else.

frontend/src/components/analyzers/AnalyzerForm/__tests__/AnalyzerForm.defaultConfigs.test.tsx
```diff
@@ -8,6 +8,10 @@ import AnalyzerForm from "../AnalyzerForm";
 import messages from "../../../../languages/en.json";
 import * as analyzerService from "../../../../services/analyzerService";
 
+const waitFor = async (callback: () => void | Promise<void>) => {
```
mozzy11, line 11: This shadows the real `waitFor`: it awaits the callback exactly once with no polling or timeout, so any assertion that isn't already true passes/fails immediately rather than waiting. Please import `waitFor` from `@testing-library/react` as `AnalyzerForm.test.tsx` already does.

mozzy11 (commented):
Status check — no new commits since Jul 3, so the three blockers from my last review all still stand, but one has changed shape and is now actionable.

**#3705 has merged.** develop now carries `analyzers/types.ts` (49 lines) and `analyzers/constants.ts`. Your branch independently adds a 152-line `types.ts` and its own `constants.js` -> `constants.ts` conversion, so the rebase is a real content conflict in both files rather than a rename — this is exactly @pmanko's third point, and it's now unblocked. Your `types.ts` is mostly a superset, so taking yours as the base is right, but three fields diverge in ways that would quietly change typing; see inline.

Still outstanding:
- **`03 Checkpoint - E2E` is red** (run 28682592855). Needs triage before this can merge.
- The hand-rolled `waitFor` in `AnalyzerForm.defaultConfigs.test.tsx` is still there — it runs the assertion once with no retry, so those async assertions aren't actually waiting.
- CONFLICTING against develop.

On @pmanko's review: the `imaginaryTarget` regression he requested changes for **is** fixed (`state?.value ?? event.imaginaryTarget?.value ?? event.target.value` in `SerialConfiguration.tsx`). His review is still formally CHANGES_REQUESTED though, and I'm not going to clear it on his behalf — after you rebase and consolidate the type modules, please re-request review from him so he can sign off on both points himself.

One forward-looking note: @pmanko's #3929 adds TypeScript 6.0.3 and a manual `npm run typecheck`. CI doesn't typecheck `.ts` today, so please run it once #3929 lands and confirm the analyzer files are clean. (Also worth knowing: the `AnalyzerProtocol = ... | string` union collapse I flagged is now on develop via #3705, so that cleanup belongs in a separate PR, not this one.)

frontend/src/components/analyzers/AnalyzerForm/__tests__/AnalyzerForm.defaultConfigs.test.tsx
manroop-2005: resolved

pmanko (changes requested):
Thanks for the Aug 3 rebase — and before the blockers, some credit where it's due, because the discipline in this migration is better than the headline issues suggest.

Across 1433 added lines there is **not one** `any`, `as any`, non-null `!`, or `@ts-ignore`. That is the single thing most JS→TS migrations get wrong and you avoided it completely. You also fixed two real bugs in passing: the Carbon v11 Checkbox callback in `TestMappingModal` (the old code read the event as the boolean, making two options impossible to untick and putting a non-boolean in an API payload) and a misused `react-intl` call in `ErrorDashboard`. And of the 23 Carbon callbacks in these files, 22 use the correct v11 signature — `({ selectedItem })`, `(_, { checked })`, `onToggle={(checked)}`. You clearly know this API.

**On my own two points, so you know exactly where you stand:**

**P1 (stepper writes 0) — still open, and this one is my fault as much as yours.** My original comment quoted `e.imaginaryTarget?.value || e.target.value` as "the previous behavior" to restore, and you restored it faithfully. But `imaginaryTarget` is a Carbon **v10** (`carbon-components-react`) mechanism. We're on `@carbon/react` 1.104.1, whose `NumberInput` never sets it — it calls `onChange(event, state)` and puts the value in the second argument. So at `SerialConfiguration.tsx:250` the `imaginaryTarget` read is dead code and it falls through to `event.target.value`, where on a stepper click `target` is the `<button>`, whose `.value` is `""` → `Number("") === 0` → `Number.isFinite(0)` is true → the helper returns **0**. The exact outcome we were trying to prevent. `tsc` flags the same line independently (`TS2339: Property 'value' does not exist on type 'EventTarget | (EventTarget & HTMLInputElement)'`). The fix is the shape you already use correctly 22 times:

```tsx
onChange={(_e, { value }) => handleFieldChange("baudRate", Number(value))}
```

Please also add the regression test — `SerialConfiguration.test.tsx` has four tests and none touches the stepper, so the suite passes unchanged with the bug present. That's why this got reported as fixed twice while still broken.

**P2 (type consolidation) — still open, and now it has teeth.** Taking develop's `types.ts` as the base and deriving from it was the right instinct, and `analyzerTypes.ts` correctly extends rather than forks it. But the split has produced a real contract mismatch: there are now two `SerialPortConfiguration` shapes — yours at `analyzerTypes.ts:89` and the pre-existing one at `services/serialService.ts:20` — and `tsc` rejects the create/update call sites at `SerialConfiguration.tsx:234` and `:236` (`TS2345: '…analyzerTypes.SerialPortConfiguration' is not assignable to '…serialService.SerialPortConfiguration'`). This is a non-strict error, so it isn't strict-mode noise. Either reuse the service's type at the boundary or make yours structurally compatible.

**Also blocking:**

**1. `vitest-globals.d.ts:12` weakens DOM typing across the whole frontend.** The `interface HTMLElement { checked?, disabled?, value? }` is inside `declare global`, and our `tsconfig.json` has no `include`, so it applies to every file in the program — production code included, despite the filename. After this, `someDiv.value` typechecks anywhere in the app. It's also why hundreds of pre-existing errors vanish: they're suppressed, not fixed. Please scope it to test files, or cast at the call sites instead.

**2. Test theater, and a new lint suppression that enables it.** `AnalyzerForm.test.tsx:25` adds `/* eslint-disable @typescript-eslint/no-unused-vars, jest/expect-expect */` — the only eslint-disable in the entire analyzers tree, absent from develop, and `jest/expect-expect` is exactly the rule that flags assertion-less tests. Three tests can't fail:
- `FieldMapping.test.tsx:672` `testValidationDashboard_DisplaysInValidationStage` — its only assertion is inside `if (validationDashboard) { expect(validationDashboard).not.toBeNull() }`, which the guard already entails. Absent → zero assertions; present → tautology. `ValidationDashboard` is never rendered in the file.
- `testSubmitForm_WithValidData_CallsAPI` — the name promises the API is called with valid data; the body asserts `expect(createAnalyzer).not.toHaveBeenCalled()`. The happy path has no positive test.
- `testAnalyzerTypeDropdown_DisplaysAllOptions` — asserts only that the dropdown isn't null (a tautology after `getByTestId`), while its comment claims it "would have caught the issue where analyzer type dropdown wasn't rendering." It wouldn't.

Please drop the `jest/expect-expect` suppression and make these assert their names. Where Carbon makes something awkward in JSDOM, that's a signal to assert on rendered output differently — not to add a comment explaining why the assertion is missing. `MappingPanel`'s draft-vs-activate tests are the right pattern; use those.

**3. `AnalyzerForm.defaultConfigs.test.tsx:11` still shadows `waitFor`** (mozzy's M2). The local `const waitFor = async (cb) => { await Promise.resolve(cb()) }` runs once with no retry, across 20 call sites, so none of those async assertions actually wait. Every other analyzer test file imports the real one. Please delete the local and import it.

**4. `AnalyzerForm.tsx:44/50` — `AnalyzerFormProps` is declared and then ignored via `_props`** (mozzy's M4). Either wire up `open`/`onClose` or drop the interface; as written it lets inert props typecheck in tests.

**5. 658 lines of net-new, unreferenced components.** `FieldMapping/QualitativeMappingModal.tsx` (317 lines) and `UnitMappingModal.tsx` (341) don't exist on develop and have zero importers anywhere. They're not mentioned in the PR body. In a PR titled as a migration these should come out and land separately, wired up and reviewed on their own merits.

**Two things that are explicitly not your problem:**
- **The red `03 Checkpoint - E2E` is infrastructure, not your code.** In run 30843323331 the job `Playwright Harness / harness 2/2` failed at "Install Playwright browsers" with `E: Could not get lock /var/lib/dpkg/lock-frontend` on all three attempts; every later step is `skipped` and zero tests ran. `harness 1/2` passed. I've re-run the failed jobs and filed the retry-hardening separately.
- **The typecheck error count is mostly ours, not yours.** `npm run typecheck` reports 410 errors in your files, but #3929 bumped us to TypeScript 6.0.3, which is strict-by-default, and our `tsconfig.json` sets no `strict` key. With `--strict false` your files drop to 79, and only 22 of those are in source. Please fix the ones called out above (the `SerialPortConfiguration` mismatch, `:250`, and the unparameterized `useState` cluster in `FieldMapping.tsx`) and ignore the implicit-any noise — settling the repo-wide baseline is on us. M5 (`AnalyzerProtocol`'s `| string`) is likewise in a file you don't touch; it came in via #3705 and belongs in a separate PR.

**To merge:** fix P1 with the `(_, { value })` shape plus a stepper regression test; resolve the `SerialPortConfiguration` mismatch; scope `vitest-globals.d.ts` to tests; drop the `jest/expect-expect` suppression and fix the three tests; restore the real `waitFor`; and pull the two unwired modals into their own PR. Then re-request review and I'll clear my CHANGES_REQUESTED.

manroop-2005:
@pmanko , i have made new clean pr for week 4 : https://github.com/DIGI-UW/OpenELIS-Global-2/pull/3974

PR #3754, opened 2026-06-22
fix: allow multi-line Provisional Clinical Diagnosis on Add Order

HerbertYiga:
Replace the single-line TextInput with a Carbon TextArea (3 rows) on the SamplePatientEntry Add Order panel so clinicians can capture longer diagnoses without character/visual truncation.


mozzy11 (approved):
Thanks @HerbertYiga

Issue #3743, opened 2026-06-18
Cannot delete device in Cold Storage Dashboard + inconsistent timestamps in Active Alerts


yiyucangqiong:
I am using an ESP32 development board with a DS18B20 temperature sensor to monitor refrigerator temperature.After adding the device in the Cold Storage Dashboard settings, I cannot delete it. Even after restarting the server, the device still remains. Additionally, in the Active Alerts section, there is no delete button for the alerts. Could a delete button be added?I have also noticed inconsistent timestamps (the time I circled in red in the screenshot). What could be the cause of this time mismatch?I will upload my ESP32 code for reference. Please help me check it.Thank you!

<img width="1485" height="824" alt="Image" src="https://github.com/user-attachments/assets/1cbb8180-5694-4063-8adb-e4518bf6bed3" />

[esp32+ds18b20.txt](https://github.com/user-attachments/files/29082833/esp32%2Bde18b20.txt)

Agaba-derrick:
CC: @reagan-meant  @samuelmale

COLLINS-WAGIMA256:
Hi @Agaba-derrick @yiyucangqiong @reagan-meant , I'd like to pick this up : starting with the "cannot delete device" issue (happy to take the Active Alerts delete buttons and timestamp inconsistency as separate PRs if you'd prefer to keep things focused).

  I reproduced it on the latest develop and traced the root cause.

  **What happens**

  Deleting a device doesn't remove it — it stays in the Settings device list permanently (even after a server restart), shown as inactive.

(screen recording attached below)
  https://github.com/user-attachments/assets/c8e669fe-35d4-48f9-bc80-195c08f300df



  **Root cause**

  - Delete is a soft delete: FreezerServiceImpl.deleteFreezer() just sets active = false.
  - But the device-management list — GET /rest/coldstorage/devices → "FreezerDAOImpl.getAllFreezers() / searchFreezers()" — has no filter on active, so it keeps returning "deleted" devices. (The status/dashboard view uses "findActiveFreezers()" with WHERE active = true, 
so it does hide them — only the management list is affected.)

  **It also resurrects**

  Because the "deleted" device is still listed, its enable/power toggle remains usable. Pressing it calls POST …/devices/{id}/toggle-status
  → "setDeviceStatus()", which sets active = true with no "is this deleted?" guard — fully bringing the deleted device back.

  **Verified at the DB level**

[...]

PR #3763, opened 2026-06-24
issue-3743: separate Cold Storage device delete from enable/disable

COLLINS-WAGIMA256:
Addresses item 1 (device delete) of #3743

  ## Problem
  Deleting a device in the Cold Storage Dashboard didn't remove it thus it stayed
  in the device list permanently even after a server restart, and its
  enable/power toggle could bring it back.

  ## Root cause
  Delete was a *soft delete that reused the `active` flag* (`deleteFreezer`
  set `active = false`). But `active` is also the enable/disable toggle, and
  the device-management list query (`FreezerDAOImpl.getAllFreezers` /
  `searchFreezers`) never filtered on it. So "deleted" devices kept showing,
  and `setDeviceStatus` (the toggle) could flip `active` back to true.

  ## Fix
  Introduce a dedicated `deleted` flag, distinct from `active`:
  - **Liquibase `046-add-freezer-deleted-column.xml`** thus adds `freezer.deleted`
    (`BOOLEAN NOT NULL DEFAULT false`), idempotent precondition, backfills existing rows.
  - **`Freezer.deleted`** field.
  - **`FreezerDAOImpl`** — `findByName`, `findActiveFreezers`, `getAllFreezers`,
    `searchFreezers` now exclude `deleted = true` (`get(id)` left unfiltered).
  - **`FreezerServiceImpl`** — `deleteFreezer` sets `deleted = true` (no longer
    touches `active`); `setDeviceStatus` rejects status changes on a deleted device.
  - **`FreezerServiceTest`** — updated the delete test + added a guard test.

  Chose a `deleted` flag over a hard delete because `freezer_reading` rows FK to
  `freezer`; a hard delete would destroy temperature history. **Happy to switch
  to a hard delete if you'd prefer** — flagged this as an open question on the issue.

  ## Testing
  Verified on the dev stack (build 3.2.1.10): add device → delete → it disappears
  from the list; DB shows `deleted = true`; `getAllFreezers` returns it no longer;
  deleted device can't be re-activated.

https://github.com/user-attachments/assets/7a3d11d6-0f5b-4e72-9d6a-b28b48a97ddd

  ## Notes / follow-ups (not in this PR)
  - A soft-deleted device keeps its (unique) `name`, so a device with that exact
    name can't be re-added. Could be addressed with a partial unique index
    (`... WHERE deleted = false`) if desired.
  - Separate pre-existing bug noticed while testing: creating a device whose
    generated `StorageDevice` code exceeds 10 chars fails with 500
    (`generateDeviceCode` truncates to 50, column max is 10). Can file separately.

### DIFF
--- a/src/main/java/org/openelisglobal/coldstorage/dao/impl/FreezerDAOImpl.java
+++ b/src/main/java/org/openelisglobal/coldstorage/dao/impl/FreezerDAOImpl.java
@@ -21,7 +21,7 @@ public class FreezerDAOImpl extends BaseDAOImpl<Freezer, Long> implements Freeze
 
     @Override
     public Optional<Freezer> findByName(String name) {
-        String hql = "FROM Freezer f WHERE lower(f.name) = lower(:name)";
+        String hql = "FROM Freezer f WHERE lower(f.name) = lower(:name) AND f.deleted = false";
         Query<Freezer> query = entityManager.unwrap(Session.class).createQuery(hql, Freezer.class);
         query.setParameter("name", name);
         query.setMaxResults(1);
@@ -30,20 +30,20 @@ public class FreezerDAOImpl extends BaseDAOImpl<Freezer, Long> implements Freeze
 
     @Override
     public List<Freezer> findActiveFreezers() {
-        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice WHERE f.active = true ORDER BY f.name";
+        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice WHERE f.active = true AND f.deleted = false ORDER BY f.name";
         return entityManager.unwrap(Session.class).createQuery(hql, Freezer.class).list();
     }
 
     @Override
     public List<Freezer> getAllFreezers() {
-        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice ORDER BY f.name";
+        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice WHERE f.deleted = false ORDER BY f.name";
         return entityManager.unwrap(Session.class).createQuery(hql, Freezer.class).list();
     }
 
     @Override
     public List<Freezer> searchFreezers(String search) {
         String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice "
-                + "WHERE lower(f.name) LIKE lower(:search) " + "ORDER BY f.name";
+                + "WHERE lower(f.name) LIKE lower(:search) AND f.deleted = false " + "ORDER BY f.name";
         Query<Freezer> query = entityManager.unwrap(Session.class).createQuery(hql, Freezer.class);
         query.setParameter("search", "%" + search + "%");
         return query.list();
--- a/src/main/java/org/openelisglobal/coldstorage/service/impl/FreezerServiceImpl.java
+++ b/src/main/java/org/openelisglobal/coldstorage/service/impl/FreezerServiceImpl.java
@@ -168,6 +168,9 @@ public class FreezerServiceImpl implements FreezerService {
     @Transactional
     public void setDeviceStatus(Long id, Boolean active) {
         Freezer freezer = requireFreezer(id);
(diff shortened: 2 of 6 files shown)


COLLINS-WAGIMA256:
-  Rebased onto the latest develop and resolved the conflict in liquibase/3.5.x.x/base.xml (the Liquibase changeset numbering had collided with newer changesets on develop). 
My migration is now 049-add-freezer-deleted-column.xml, included after the existing 048-test-reagent-link.xml, and the branch is a single clean commit on top of the current develop tip thus no remaining conflicts.

 Ready for review whenever you have a moment. @Agaba-derrick @reagan-meant 

-   One small thing worth flagging from testing: freezer.name has a UNIQUE constraint, so a soft-deleted device keeps its name and blocks reusing that name on a new device. I left that out to keep this PR focused on the reported delete bug, but happy to follow up with a partial unique index (WHERE deleted = false) if you'd like it addressed here or in a separate issue.

mozzy11 (commented):
This is a well-diagnosed fix, @COLLINS-WAGIMA256 — you correctly identified that delete was overloading the active flag that also serves as the enable/disable toggle, and the dedicated deleted column with an idempotent, backfilling migration plus DAO-level filtering is the right design; keeping get(id) unfiltered was a thoughtful touch. The branch has drifted into conflict, almost certainly on liquibase/3.5.x.x/base.xml and the changeset number — please rebase onto develop and renumber the changeset after the current latest. Once that's pushed I'll approve the CI run; with FreezerServiceTest green this is close to merge-ready, and it slots in well alongside your #3811 timestamp fix.

COLLINS-WAGIMA256:
Thanks for the review @mozzy11, l am glad the dedicated-column approach makes sense. Rebased onto the latest `develop` and renumbered the changeset to `059-add-freezer-deleted-column.xml` (id `3.5.0-059-add-freezer-deleted`) so it no longer clashes with the changesets that landed in the meantime. Branch is clean and mergeable now, it should be good for a CI approval whenever you have a moment.

mozzy11 (commented):
Re-reviewed against current develop: the bug is still live there (`deleteFreezer` still sets `active = false`), so this fix remains needed and the design (dedicated `deleted` flag, DAO-level filtering, guard in `setDeviceStatus`, listing-exclusion tests) is sound.

Two blockers before this can merge:
1. **Rebase required** — the branch is still in conflict. Develop's Liquibase sequence has moved to `067-...`; your `059-add-freezer-deleted-column.xml` now collides with develop's `059-results-r1-unified-worklist.xml`. Please rebase and renumber to the next free slot (068+), updating `base.xml` accordingly.
2. **Coordination with #3828** — the draft "Harden cold storage monitoring" PR from the core team includes an equivalent freezer `deleted`-flag migration (`052-add-freezer-deleted-column.xml`) plus the same service/DAO changes as part of its #3743 fixes. Maintainers should decide which lands first; whichever merges second must drop or reconcile its duplicate migration (both use `columnExists` preconditions, so the schema itself is safe, but the code changes will conflict).

One functional note (you already flagged it in the description): with `findByName` now filtered to `deleted = false`, re-adding a device with a soft-deleted device's name will pass the application-level uniqueness check and then fail on the DB unique constraint with an unfriendly 500. Adding the partial unique index (`WHERE deleted = false`) in this same migration, or keeping `findByName` unfiltered for the uniqueness path, would close that gap now rather than as a follow-up.

COLLINS-WAGIMA256:
Thanks for the thorough re-review, @mozzy11.

1. **Rebase** is done. Rebased onto latest `develop` and renumbered the changeset to `068-add-freezer-deleted-column.xml` (id `3.5.0-068-add-freezer-deleted`), updating `base.xml`. Branch is clean and mergeable.

2. **Coordination with #3828** is agreed. Since our migrations both use `columnExists` preconditions the schema is safe either way; it's the service/DAO changes that'll conflict. Happy to defer to the maintainers' call on ordering , if this lands second I'll drop the duplicate migration and reconcile the code against #3828; if it lands first, #3828 can rebase over it.

3. **Unique-name gap** is a good point. I'd like to close it here rather than as a follow-up by adding the partial unique index (`WHERE deleted = false`) in this same migration (dropping the existing full unique constraints on `name` and replacing them), so a soft-deleted device's name can be reused cleanly.
 Only hesitation is duplicating migration work while the #3828 sequencing is open , happy to add it now if you'd like it in this PR, or fold it into whichever migration ends up landing.

Issue #3759, opened 2026-06-23
Implement, stabilize, and aggressively refactor the integration tests for Sample Workflow


Agaba-derrick:
The Sample Workflow within OpenELIS Global 2 manages the end-to-end lifecycle of a sample—from initial order creation to storage, QA checklist execution, EQA handling, and final sequence transmission. Currently, the integration testing coverage across these services is inconsistent. Some tests are missing, while others rely on outdated testing patterns (e.g., using `assertNotNull` instead of exact database validations, or improperly handling `BaseWebContextSensitiveTest` environment state).

## Objective
Implement, stabilize, and aggressively refactor the integration test suites for all services associated with the Sample Workflow. The objective is to guarantee reliable test execution using isolated database states and `BaseWebContextSensitiveTest`, adhering strictly to OpenELIS Global 2 architectural testing guidelines.

## Scope of Work

The following services must be audited, refactored, and/or written from scratch:

- [ ] **`GenericSampleOrderService`** (`sample/service/`)
  - *Goal*: Verify order persistence, FHIR sync artifacts, and observation creation. 
- [x] **`SampleEditService`** (`sample/service/`)
  - *Note*: Partially completed. Need to ensure no regressions and verify `SampleItem` date modifications and cascading cancellations.
- [ ] **`SampleStorageService`** (`samplestorage/service/`)
  - *Goal*: Validate storage location assignments, capacity limitations, and retrieval workflows.
- [x] **`SampleQaChecklistService`** (`sampleqaevent/service/`)
  - *Goal*: Ensure QA event creation, validation criteria, and status updates on sample items.
- [ ] **`SampleTypeTestAssignService`** (`typeofsample/service/`)
  - *Goal*: Verify correct test-to-sample-type mapping, restriction enforcements, and assignment logic.
- [ ] **`SampleEQAService`** (`sample/service/`)
  - *Goal*: Test External Quality Assessment (EQA) sample identification and specialized reporting workflows.
- [x] **`SampleTransmissionSequenceService`** (`sample/service/`)
  - *Goal*: Validate tracking of transmitted samples to external systems (e.g., analyzers, HIEs) and sequence generation.

## Technical Requirements
1. **Aggressive Assertions**: Do not use `assertNotNull()`. Validate exact data values (e.g., specific `AnalysisStatus` strings, exact millisecond timestamps, expected list sizes, exact enum states).
[...]

Issue #3748, opened 2026-06-19
Bug: `getResultCalculationByTest` Fails with `LIMSRuntimeException` Due to Missing `select r` in HQL Query


Agaba-derrick:
`ResultCalculationDAOImpl.getResultCalculationByTest` always throws a `LIMSRuntimeException` caused by an `IllegalArgumentException` from Hibernate 6. The HQL query performs a `JOIN` without an explicit `select r` clause, causing Hibernate to return multiple results (both `ResultCalculation` and `Test`), which it cannot map to a single `TypedQuery<ResultCalculation>`.

### Steps to Reproduce
Call `resultCalculationService.getResultCalculationByTest(test)` with any valid `Test` entity.

### Expected Behaviour
Returns a `List<ResultCalculation>` containing all result calculations associated with the given test.

### Actual Behaviour
Throws:
```
org.openelisglobal.common.exception.LIMSRuntimeException: Error in ResultCalculationDAOImpl getResultCalculationByPatientAndTest
Caused by: java.lang.IllegalArgumentException: Cannot create TypedQuery for query with more than one return using requested result type [org.openelisglobal.testcalculated.valueholder.ResultCalculation]
```

### Root Cause
In `ResultCalculationDAOImpl.java`, the HQL query in `getResultCalculationByTest` is missing an explicit `select r` clause required by Hibernate 6:

```java
// Current (broken)
String sql = "from ResultCalculation r JOIN r.test t WHERE t.id = :testId";

// Fix
String sql = "select r from ResultCalculation r JOIN r.test t WHERE t.id = :testId";
```

Additionally, the `catch` block in the same method references the wrong method name in its error message:

```java
// Current (misleading)
handleException(e, "getResultCalculationByPatientAndTest");

// Fix
handleException(e, "getResultCalculationByTest");
```

### Environment
- Hibernate 6
- Detected via `ResultCalculationServiceTest` integration tests

### Affected File
`src/main/java/org/openelisglobal/testcalculated/daoimpl/ResultCalculationDAOImpl.java` — `getResultCalculationByTest` method (line 49)

gp07-cmd:
Hey @Agaba-derrick I have created a PR for this issue (3748), do check it out.

Hariom-kr:
@Agaba-derrick Can i Work on this?

Agaba-derrick:
@Hariom-kr there is already an existing pr on this issue, kindly look for another issue  to work upon !

PR #3776, opened 2026-06-26
Feat: Add unassigned sample item service integration tests

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
Implemented comprehensive integration tests for `UnassignedSampleItemService`. The test suite validates the core business logic using the `BaseWebContextSensitiveTest` framework and a carefully crafted DBUnit XML fixture dataset.

### Fixture Layout (`testdata/unassigned-sample-item-service.xml`)
- `sample_item id=1`: ACC-001 / Blood Sample / sort_order=1 → **UNASSIGNED**
- `sample_item id=2`: ACC-001 / Urine Sample / sort_order=2 → **UNASSIGNED**
- `sample_item id=3`: ACC-002 / Serum Sample / sort_order=1 → **ASSIGNED** to box `BOX-2024-001`

*Note: Each sample item has exactly one referral pointing to "National Reference Lab" with status `CREATED`.*

### What Was Done
- **Robust Value-Level Assertions**: Implemented thorough property-level assertions for all `SampleItemDTO` fields and `ReferralTestDTO` associations, matching the aggressive testing style of `PatientServiceTest`.
- **Database Fixture**: Created a dedicated XML dataset seeding samples, sample items, referrals, boxes, and box item assignments to test accurate relational mappings.
- **Working Paths Fully Tested**: The `isAlreadyAssigned(String)` service method utilizes a separate, correctly-typed DAO query. It has been fully tested and confirmed to be working as expected against the fixture.

### Known Bug (Reported)
During testing, a known production bug was encountered and fully documented in the test suite:

Several DAO methods (e.g., `BoxSampleItemDAOImpl.getAllAssignedSampleItemIds` and `ReferralDAOImpl.getReferralsBySampleItemId`) define HQL queries that attempt to match a `String` parameter against `SampleItem.id`. Because `SampleItem.id` is backed by a numeric sequence in the database, Hibernate throws an `IllegalArgumentException` (*"Parameter value [1] did not match expected type [java.lang.String (n/a)]"*).

The service catches this exception, but because the transaction has already been marked as rollback-only by the nested DAO call, the Spring transaction proxy throws an `UnexpectedRollbackException` when the service method returns.

**Test Suite Strategy for Broken Methods:**
This test suite documents "the way things are" by asserting that the following methods currently throw the `UnexpectedRollbackException`:
- `getAllUnassigned()`
- `searchUnassignedByAccessionNumber()`
- `getSampleItemById()`

The intended behavioural tests (with full data assertions) have been written and preserved, but marked with `@Ignore` so they can be easily re-enabled once the underlying production code is fixed.

## Screenshots

Screenshot from 2026-06-26 13-14-10

## Related Issue
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3759
## Other

@mozzy11 @HerbertYiga @pmanko


mozzy11 (commented):
This one shows real maturity, Derrick — instead of only testing what works, you mapped the whole service: isAlreadyAssigned is fully covered against a properly relational fixture (referrals, shipping box, box assignment), and you wrote out the *intended* behavior of the broken methods as @Ignore'd tests that are ready to switch on the moment the DAO bug is fixed. That's a great pattern.

A few asks before merge:

- **Link the bug issue number** in each @Ignore reason and in the 'BROKEN PATHS' comment block (like you did with #3748 in your result-calculation PR). 'Known BoxSampleItemDAOImpl bug' will be hard to trace in six months.
- **Reconsider the exception type** in the three broken-path pins. UnexpectedRollbackException is Spring's transaction wrapper, so these tests will also fail on any unrelated transaction-boundary change. Pinning the root cause (or at least adding a comment that the wrapper type is incidental) makes the intent clearer.
- When the DAO fix lands, remember this PR needs the follow-up: delete the 3 pins, un-ignore the 5 intended-behavior tests.

The fixture design here is genuinely good — happy to approve once the issue links are in.

Agaba-derrick:
@mozzy11 we can kindly  have a look at this too

src/test/java/org/openelisglobal/shipment/UnassignedSampleItemServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+    public void isAlreadyAssigned_nonExistentSampleItem_returnsFalse() {
+        assertFalse("Non-existent sample item must return false",
+                unassignedSampleItemService.isAlreadyAssigned("99999"));
+    }
+
+    // ─────────────────────────────────────────────────────────────────────────
+    // BROKEN PATHS - Documenting Current Behaviour (See Issue #3660)
+    // ─────────────────────────────────────────────────────────────────────────
+
+    @Test(expected = UnexpectedRollbackException.class)
```
mozzy11: Pinning `UnexpectedRollbackException` locks in #3660's incidental symptom (the tx marked rollback-only), not behavior. When the bug is fixed, these three go red while the intended-behavior tests below remain @Ignore'd — the opposite of what you want. The fix is one line (`ReferralDAO.getReferralsBySampleItemId(Integer)` → `String`, dropping the `Integer.parseInt` calls in `UnassignedSampleItemServiceImpl`); consider fixing it in/alongside this PR and enabling the six ignored tests instead.

src/test/java/org/openelisglobal/shipment/UnassignedSampleItemServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.UnexpectedRollbackException;
+
+public class UnassignedSampleItemServiceTest extends BaseWebContextSensitiveTest {
+
+    @Autowired
+    private UnassignedSampleItemService unassignedSampleItemService;
+
+    @Before
+    public void loadFixture() throws Exception {
```
mozzy11, line 22: Add an `@After` calling `cleanRowsInCurrentConnection` for the fixture's tables (referral, box_sample_item, shipping_box, analysis, sample_item, sample). The shared container means rows left by the last test leak into whichever class surefire runs next — this repo has had order-dependent failures from exactly this.

src/test/java/org/openelisglobal/shipment/UnassignedSampleItemServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+        // wrapper.
+        // The root cause is a Hibernate type mismatch on sampleItem.id (Issue #3660).
+        unassignedSampleItemService.getSampleItemById("1");
+    }
+
+    // ─────────────────────────────────────────────────────────────────────────
+    // INTENDED BEHAVIOUR - Ignored until bug is fixed
+    // ─────────────────────────────────────────────────────────────────────────
+
+    @Ignore("Blocked by known BoxSampleItemDAOImpl bug (type mismatch on sampleItem.id) - Issue #3660")
```
mozzy11: If #3660 can't be fixed here, pair each `@Ignore`d intended-behavior test with its exception-pinning counterpart via a comment naming which pinning test must be deleted when the ignore lifts, so the halves can't drift.

mozzy11 (changes requested):
Thanks @Agaba-derrick — good coverage of `isAlreadyAssigned` (both branches + non-existent id), a well-documented fixture, and the `UnassignedSampleServiceTest` Integer→String fix is correct.

The structural problem: three tests pin `UnexpectedRollbackException` — the *incidental symptom* of bug #3660 — while the six tests of intended behavior sit `@Ignore`d. The moment #3660 is fixed (a one-line `Integer`→`String` in `ReferralDAO.getReferralsBySampleItemId`), the pinning tests go red while the real ones stay ignored — the inverted failure mode. The bug is small enough to fix here so the real assertions can run. Also missing `@After` cleanup (shared DB, surefire order matters in this repo). Details inline.

Agaba-derrick:
@mozzy11:

Thanks for catching the test behavior! I completely agree that pinning the UnexpectedRollbackException was masking the actual bug.

I have fixed issue #3660 directly in this PR by changing ReferralDAO.getReferralsBySampleItemId() to accept a String (matching the SampleItem ID type) and dropped all the Integer.parseInt workarounds in UnassignedSampleItemServiceImpl and BoxSampleItemServiceImpl. With the bug resolved, I've deleted the tests that were pinning the broken paths and removed the @Ignore tags from the six tests verifying the intended behavior.

Additionally, I've added an @After cleanup block using cleanRowsInCurrentConnection() for the fixture's tables (referral, box_sample_item, shipping_box, analysis, sample_item, sample) to prevent any cross-test leakage. I also patched an issue in the fixture (unassigned-sample-item-service.xml) where tests and sample types inadvertently shared the same name_localization_id, which was causing Test.getName() to incorrectly fetch the sample type's localized string instead of the test name

PR #3779, opened 2026-06-26
fix: round home dashboard turn-around times to whole hours

HerbertYiga:
Average Reception-To-Validation, Reception-To-Result, and Result-To-Validation times are now returned as Long values rounded to the nearest hour (Math.round), removing the decimal component from the home dashboard JSON payloads.


pmanko (approved):
Reviewed via code-qa lenses (simplicity / test-coverage / hygiene).

Lean, consistent `Double`→`Long` change — all callers are internal and updated in lockstep, so no contract break (JSON serializes `27` and `27.0` identically; frontend `Tile.value: number` is unaffected). Net effect cleans up raw decimal rendering on the home dashboard. CI fully green. Approving.

PR #3780, opened 2026-06-26
i18n: wire header icon aria-labels to React Intl

HerbertYiga:
Replace hardcoded "Search", "Notifications", "User", "Lang", and "Help" aria-labels (which surface as hover tooltips on Carbon HeaderGlobalAction) with i18n lookups. Adds the corresponding keys to en.json so non-English translations can flow in via Transifex.


frontend/src/components/layout/Header.jsx
```diff
@@ -139,7 +139,9 @@ function OEHeader({
   }, [userSessionDetails.authenticated]);
 
   const panelSwitchLabel = () => {
-    return userSessionDetails.authenticated ? "User" : "Lang";
+    return userSessionDetails.authenticated
+      ? intl.formatMessage({ id: "header.icon.user" })
+      : intl.formatMessage({ id: "header.icon.lang" });
```
ibacher, line 144: We could adopt (for consistency) the model from below:


```suggestion
    return intl.formatMessage({
         id:
           userSessionDetails.authenticated
             ? "header.icon.user"
             : "header.icon.lang"
    });
```

Which feels more readable

pmanko (approved):
Reviewed via code-qa lenses (simplicity / test-coverage / hygiene).

Clean i18n wiring — `intl` is in scope in both files (`useIntl()`), the 8 new keys are alphabetically placed, and the change is `en.json`-only (Transifex source-of-truth rule satisfied; duplicate-key and translation-source-of-truth checks both green). CI fully green. Approving.

pmanko:
Optional nit (non-blocking): the menu button repeats the same 3-way ternary for both `aria-label` and `title` in `Header.jsx`. Could hoist to a local `const menuLabel = intl.formatMessage({ id: ... })` and reuse. Fine to merge as-is.

ibacher:
Resolved my example because I did not mean it to be merge blocking

PR #3808, opened 2026-07-01
fix(i18n): Translate CustomDatePicker placeholder by date locale

HerbertYiga:
fix(i18n): translate CustomDatePicker placeholder by date locale


frontend/src/components/common/CustomDatePicker.jsx
```diff
@@ -102,11 +104,13 @@ const CustomDatePicker = (props) => {
       >
         <DatePickerInput
           id={props.id}
-          placeholder={
-            configurationProperties.DEFAULT_DATE_LOCALE == "fr-FR"
-              ? "dd/mm/yyyy"
-              : "mm/dd/yyyy"
-          }
+          placeholder={intl.formatMessage({
+            id: "datepicker.placeholder",
```
mozzy11: This is not right ,it assumes the translatable date format  is always`dd/mm/yyyy` yet it could be mm/dd/yyyy

HerbertYiga: i see, kindly looking through

HerbertYiga: so @mozzy11 the expectation is that the value for datepicker.placeholder will depend on the locale, if its french then a french translation will be picked for datepicker.placeholder , same to other languages.

mozzy11: Based on the `Date Locale config` the date format could either be `dd/mm/yyyy` or `mm/dd/yyyy` regardless of the locale.
Then either of those farmat values could be translated based on the locale.

For now you only hardcoded a sinlge date format ie  `dd/mm/yyyy`  to be translated into multiple languages based on the locale.

But the actual format it'self is not static , its dynamic based on the date locale config

HerbertYiga: @mozzy11 i the two date formats can now be translated with the current changes

mozzy11 (approved):

PR #3811, opened 2026-07-02
issue-3743: fix seconds/millisecond timestamp handling in Cold Storage

COLLINS-WAGIMA256:
## Problem
  Part of #3743 (the "inconsistent timestamps" item):
  - The **Temperature Trends** chart X-axis showed **Jan 1970** (`1月21日` in a non-English locale).
  - The **Active Alerts** "Duration" showed ~29,000,000 min (~56 years).
  - The "Started" column was correct.

  ## Root cause
  The REST API serializes `OffsetDateTime` as a Unix epoch in **seconds**
  (`WRITE_DATES_AS_TIMESTAMPS`), but the chart label and duration calc passed it
  straight into `new Date(...)`, which expects **milliseconds** → the value landed
  ~20 days after 1 Jan 1970. `formatDateTime` ("Started") already ×1000, so only it
  looked right.

  ## Fix
  - New shared helper `coldStorage/shared/timeUtils.js` (`toDate`, `formatDuration`).
  - Unit tests (`timeUtils.test.js`).

  ## Screenshots
  -Below is  the screen record after the fix
https://github.com/user-attachments/assets/90406f7d-0fa6-49e7-9c36-c86e2da24014


COLLINS-WAGIMA256:
@yiyucangqiong @reagan-meant @Agaba-derrick requesting for your review on this PR

mozzy11 (commented):
Great debugging, @COLLINS-WAGIMA256 — the seconds-vs-milliseconds epoch diagnosis is exactly right, and centralizing it in a shared timeUtils with unit tests (instead of patching three call sites independently) is the correct shape for the fix; the screen recording confirming the chart axis and alert durations is appreciated. CI is fully green and the branch is mergeable, so this is ready for maintainer review as-is. One optional thought for review: a brief comment in timeUtils noting the backend's WRITE_DATES_AS_TIMESTAMPS convention would help the next person who hits a raw epoch. Queuing this for review alongside your #3763.

COLLINS-WAGIMA256:
Thanks @mozzy11, i have added a short header comment in `timeUtils.js` documenting the `WRITE_DATES_AS_TIMESTAMPS` epoch-seconds convention and why `toDate()` scales by 1000, so the next person who hits a raw epoch has the context right there. Thanks for the review.

yiyucangqiong:
Sorry, I tried to pull your code from GitHub, but the build failed (because I’m pulling the image files remotely). I hope the PR can be reviewed and merged as soon as possible so I can see the fixed version in the update soon.
I’ve also discovered some new issues and hope you can help fix them. Related to #3904.

mozzy11 (approved):
Approving. The seconds-vs-milliseconds diagnosis is correct, the fix is centralized in a shared `timeUtils` with solid unit tests (epoch-seconds, genuine millisecond epochs, ISO strings, null/garbage inputs), and the `WRITE_DATES_AS_TIMESTAMPS` header comment requested earlier has been added. CI is fully green and the branch is mergeable.

One cross-PR note for whoever merges: draft #3828 fixes the same issue at the source by switching the cold-storage API to ISO-8601 serialization. That does not conflict with this PR — `toDate()` already parses ISO strings, so this frontend normalization remains correct before and after #3828 lands, and the two can merge in either order.

COLLINS-WAGIMA256:
On **item 2** (delete/clear Active Alerts) and how it relates to @mherman22's #3828:

I saw @mozzy11's note on #3811 that #3828 fixes the timestamp issue at the source and the two layers merge fine in either order , I think item 2 may split the same way.
 #3828 looks like it adds the backend `DELETE /rest/alerts/{id}` endpoint, but the actual bug users hit is on the **frontend**: the Active Alerts table has no Clear/Resolve action at all (only View + Acknowledge, and Acknowledge currently 500s because the dashboard hardcodes `userId: 1`).

@mherman22 @mozzy11, if #3828 provides the backend endpoint, I'm happy to do the complementary **frontend** piece: wire a Clear/Resolve action into the dashboard against that endpoint and fix the hardcoded-`userId` acknowledge/resolve call, so removing an alert actually works end to end. If #3828 already covers the UI too, then it's fully handled there and I'll leave it. Whichever is more useful , just point me at the layer you'd like me to take.

mherman22:
> On **item 2** (delete/clear Active Alerts) and how it relates to @mherman22's #3828:
> 
> I saw @mozzy11's note on #3811 that #3828 fixes the timestamp issue at the source and the two layers merge fine in either order , I think item 2 may split the same way. #3828 looks like it adds the backend `DELETE /rest/alerts/{id}` endpoint, but the actual bug users hit is on the **frontend**: the Active Alerts table has no Clear/Resolve action at all (only View + Acknowledge, and Acknowledge currently 500s because the dashboard hardcodes `userId: 1`).
> 
> @mherman22 @mozzy11, if #3828 provides the backend endpoint, I'm happy to do the complementary **frontend** piece: wire a Clear/Resolve action into the dashboard against that endpoint and fix the hardcoded-`userId` acknowledge/resolve call, so removing an alert actually works end to end. If #3828 already covers the UI too, then it's fully handled there and I'll leave it. Whichever is more useful , just point me at the layer you'd like me to take.

@COLLINS-WAGIMA256 no worries, simply add the changes you see that need to be improved on the cold storage. I will follow up and rebase my PR to remove what you have covered.

mherman22 (approved):

Issue #3803, opened 2026-06-30
Implement Integration Tests for Test Configuration


Agaba-derrick:
## Background
A recent test gap analysis identified several critical services related to test configuration, methods, and notifications that lack comprehensive integration test coverage. To ensure the reliability and stability of these components, we need to implement robust integration tests that validate real database interactions, transactional boundaries, and service logic within the OpenELIS Global 2 framework.

## Services to Test

**Test Configuration (`org.openelisglobal.testconfiguration.service`)**
- [x] `TestActivationService`
- [ ] `TestAddService`
- [x] `TestModifyService`
- [x] `TestCalculationService`
- [ ] `TestSectionTestAssignService`

**Methods (`org.openelisglobal.method.service`)**
- [x] `TestMethodService`

**Notifications (`org.openelisglobal.notification.service`)**
- [ ] `TestNotificationConfigService`
- [ ] `TestNotificationService`

## Definition of Done (DoD)
- [ ] Integration test classes are created for each of the listed services.
- [ ] All tests extend the standard `BaseWebContextSensitiveTest` framework to ensure proper Spring context and database state management.
- [ ] Tests validate real database interactions utilizing XML datasets (e.g., using DBUnit/Hibernate via the existing testing architecture) rather than mocking the DAOs.
- [ ] Aggressive assertions are implemented to validate expected state changes, successful operations, and transactional rollbacks (e.g., constraint violations or exception handling).
- [ ] Tests comply with OpenELIS Global 2 development guidelines (JUnit 4, no inline code cleanup hacks, correct use of `TransactionalAwareTestDataSource`).
- [ ] The test suite runs and passes successfully locally and in the CI pipeline.

## Implementation Notes
- When setting up test data, ensure that dependencies like methods, test sections, and configurations exist before calling the respective services to avoid constraint violations.
- Reference recent integration testing patterns (like `UserServiceTest` or `UnassignedSampleItemServiceTest`) for best practices on state management and assertions.

Ariho-Seth:
Hello @Agaba-derrick, is this issue complete according to the PRs linked above?

Agaba-derrick:
@Ariho-Seth this Issue is still an -inprogressive !

PR #3812, opened 2026-07-02
Add Test AddService Integration  Test

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Adds six database-backed integration tests for `TestAddService.addTests(...)`. They cover localization persistence, sample-type links, test results, result limits, panel items, section and panel activation, and ordered-test updates. The class-level transaction rolls back service writes so the tests do not affect other integration suites.

## Screenshots
Screenshot from 2026-07-02 21-03-38

## Related Issue
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3803

## Other

@mherman22 @HerbertYiga @mozzy11


mozzy11 (commented):
Good choice of target, Derrick — TestAddService.addTests had no coverage and it fans out into localizations, sample-type links, test results, result limits and panel items; you verified every one of those side effects with exact-value assertions, including the nice detail that the section and panel flip from inactive to active. The assertion messages will make CI failures easy to read too.

One important fix before merge — **cleanup**:

- The class isn't @Transactional, and the fixture (test-add-service.xml) contains no rows for test, sampletype_test, test_result, result_limits or panel_item — so the state-management helper never truncates those tables. That means the tests your service calls insert ('HIV Load Test', 'Glucose Test', etc.) stay in the shared database after this class runs and can leak into sibling test classes.
- Easiest fix: add @Transactional at the class level so every service write rolls back (see TestModifyServiceTest in #3846 for the same pattern), or alternatively delete the inserted rows in an @After.

With that one change this is ready — the coverage itself is exactly what we want.

src/test/java/org/openelisglobal/testconfiguration/service/TestAddServiceTest.java
```diff
@@ -0,0 +1,294 @@
...
+        org.openelisglobal.test.valueholder.Test inserted2 = tests2.get(0);
+
+        assertEquals("Test Alpha name must be the localization value", "Multi Test", inserted1.getName());
+        assertEquals("Test Beta name must be the localization value", "Multi Test", inserted2.getName());
+
+        assertEquals("Test Alpha description must match", "Multi Test Alpha(whole blood)", inserted1.getDescription());
+        assertEquals("Test Beta description must match", "Multi Test Beta(whole blood)", inserted2.getDescription());
+        assertEquals("Test Alpha LOINC must be 88881-1", "88881-1", inserted1.getLoinc());
+        assertEquals("Test Beta LOINC must be 88882-2", "88882-2", inserted2.getLoinc());
+        assertEquals("Both tests must be ACTIVE", inserted1.getIsActive(), inserted2.getIsActive());
```
mozzy11: **Blocking (Constitution V.6 Inversion Test):** this only asserts the two `isActive` values are equal to each other — if the service inserted both tests inactive ("N" == "N") it would still pass. Assert the expected value for each: `assertEquals("Y", inserted1.getIsActive())` and `assertEquals("Y", inserted2.getIsActive())`.

src/test/java/org/openelisglobal/testconfiguration/service/TestAddServiceTest.java
```diff
@@ -0,0 +1,294 @@
...
+import org.openelisglobal.testresult.service.TestResultService;
+import org.openelisglobal.testresult.valueholder.TestResult;
+import org.openelisglobal.typeofsample.service.TypeOfSampleService;
+import org.openelisglobal.typeofsample.service.TypeOfSampleTestService;
+import org.openelisglobal.typeofsample.valueholder.TypeOfSample;
+import org.openelisglobal.typeofsample.valueholder.TypeOfSampleTest;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+@Transactional
```
mozzy11, line 35: Non-blocking: class-level `@Transactional` overrides the base class's deliberate `NOT_SUPPORTED`, so every test rolls back and assertions read uncommitted session state — commit-time constraint violations and the durability of the activation UPDATEs are never verified. This matches the TestModifyServiceTest precedent (and the rollback is what cleans up the inserted `test` rows, since the fixture doesn't name `test` for truncation) — add a brief comment saying it's intentional and why, so it isn't mistaken for an accident.

src/test/java/org/openelisglobal/testconfiguration/service/TestAddServiceTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+    @Autowired
+    private PanelItemService panelItemService;
+
+    private static final String SECTION_ID = "1";
+    private static final String SAMPLE_TYPE_ID = "1";
+
+    @Before
+    public void init() throws Exception {
+        executeDataSetWithStateManagement("testdata/test-add-service.xml");
```
mozzy11, line 73: Non-blocking coverage gap: no test for the no-flip paths (section/panel already "Y" must NOT be rewritten — a regression that unconditionally rewrites them would pass every current test) nor for the `set.sortedTests` re-order loop. One extra test seeding an active section+panel with a populated `sortedTests` list closes all three.

src/test/java/org/openelisglobal/testconfiguration/service/TestAddServiceTest.java
```diff
@@ -0,0 +1,294 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+@Transactional
+public class TestAddServiceTest extends BaseWebContextSensitiveTest {
+
+    @Autowired
+    private TestAddService testAddService;
+
+    @Autowired
+    private TestAddController testAddController;
```
mozzy11, line 42: Minor: the controller bean is autowired solely for `testAddController.new TestSet()` (non-static inner class) — fine per precedent, but a follow-up making `TestSet` a static nested class/top-level DTO would decouple the service and tests from the controller. Also the PR body's Summary is empty and the checklist unchecked — please fill in what the suite covers.

mozzy11 (dismissed):
Thanks @Agaba-derrick — five solid integration tests for `TestAddServiceImpl.addTests(...)` hitting the branches that matter (localization sharing, inactive-section/panel activation, sampletype_test linkage, TestResult/ResultLimit persistence, multi-set inserts), and both the class-level `@Transactional` and controller-inner-class `TestSet` construction follow the `TestModifyServiceTest` precedent.

One small **blocking** fix: an assertion that fails the Constitution's Inversion Test (V.6) — it compares the two inserted tests' `isActive` values *to each other*, so both being wrongly `"N"` would still pass. One-line fix inline, plus a couple of non-blocking suggestions. Happy to approve as soon as that assertion is corrected.

HerbertYiga (approved):

pmanko (approved):

PR #3814, opened 2026-07-02
OGC-427 M1: Batch workplan foundation

pmanko:
## Summary

This draft PR implements the analyzer-independent M1 foundation for the OGC-427 / OGC-811 batch workplan stream.

- Adds persisted batch workplans with `DRAFT -> ACTIVE -> COMPLETED -> ARCHIVED` lifecycle
- Adds `/rest/batch-workplans` pending-test and batch endpoints using the existing OpenELIS service/DAO layering
- Adds the `/Workplan` Carbon UI with Pending Tests and Batches tabs, selection, batch creation, and lifecycle actions
- Adds Liquibase schema/menu/module-url wiring and English i18n keys
- Adds focused backend service tests and frontend component tests

## Scope Notes

This is intentionally analyzer-independent. It does not add analyzer/file polling, reagent QC gating, inline QC entry, reagent consumption events, controls, NCE, or Results Entry v4 behavior. Those remain later milestones layered on top of this batch foundation.

## Validation

- `mvn spotless:apply`
- `npm run format`
- `mvn -Dtest=org.openelisglobal.batchworkplan.service.BatchWorkplanServiceImplTest test`
- `npm test -- --run src/components/workplan/BatchWorkplan.test.jsx`
- `mvn clean install -DskipTests -Dmaven.test.skip=true`
- `npm run build`

## Known Existing Build Warnings

Frontend production build completes, but still emits existing project warnings for CSS `:global`, large chunks, and direct `eval` in `src/components/admin/calculatedValue/CalculatedValueForm.tsx`.


samuelmale:
@pmanko is it possible to have a short video demonstrating the visual changes?

src/main/resources/liquibase/3.5.x.x/057-batch-workplan-foundation.xml
```diff
@@ -0,0 +1,142 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <changeSet id="OGC-427-batch-workplan-foundation" author="OGC">
```
mozzy11, line 8: develop already has `057-terminology-display-name.xml` (from #3859) and is now through `067` (#3625), so this file needs renaming to `068-batch-workplan-foundation.xml` (or the next free number at rebase time). The changeset ids themselves are unique, so only the filename and the base.xml include need to move.

src/main/resources/liquibase/3.5.x.x/base.xml
```diff
@@ -153,4 +153,8 @@
   <!-- OGC-949 / OGC-1129: analyzer target→component mapping columns -->
   <include relativeToChangelogFile="true" file="056-analyzer-component-mapping.xml"/>
 
+
+  <!-- OGC-427 / OGC-811 M1: batch workplan foundation -->
+  <include relativeToChangelogFile="true" file="057-batch-workplan-foundation.xml"/>
```
mozzy11, line 158: This include is the only real merge conflict with develop — the includes for 057–067 landed after this branched. On rebase, re-anchor the (renumbered) include after `067-sample-type-disposal-instructions.xml`.

src/main/java/org/openelisglobal/batchworkplan/service/BatchWorkplanServiceImpl.java
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    private List<String> workplanPendingStatusIds() {
+        return Arrays.asList(statusService.getStatusID(AnalysisStatus.NotStarted),
+                statusService.getStatusID(AnalysisStatus.BiologistRejected),
+                statusService.getStatusID(AnalysisStatus.TechnicalRejected),
+                statusService.getStatusID(AnalysisStatus.NonConforming_depricated));
+    }
+
+    private List<BatchWorkplanStatus> openStatuses() {
```
mozzy11, line 182: Naming nit with a behavior question: `openStatuses()` includes COMPLETED, so analyses in a completed batch stay locked out of new batches until it is archived. If that lock is intentional, a name like `blockingStatuses()` plus a short Javadoc would make it explicit; if not, COMPLETED should drop out here.

src/main/java/org/openelisglobal/batchworkplan/service/BatchWorkplanServiceImpl.java
```diff
@@ -0,0 +1,314 @@
...
+        String testId = StringUtils.defaultIfBlank(response.getTestId(), "unknown-test");
+        String methodId = StringUtils.defaultIfBlank(response.getMethodId(), "manual");
+        return testId + ":" + methodId;
+    }
+
+    private Integer toUserId(String sysUserId) {
+        if (StringUtils.isBlank(sysUserId)) {
+            return null;
+        }
+        return Integer.valueOf(sysUserId);
```
mozzy11, line 312: `Integer.valueOf(sysUserId)` throws NumberFormatException for a non-numeric sysUserId, which would surface as a 500 from the controller (the same applies to the `setSysUserId` override in BatchWorkplan). Guarding with a tryParse-style helper that returns null would keep bad input on the 400 path.

mozzy11 (changes requested):
Thanks Piotr — this is a well-layered M1: clean DAO/service/controller separation with @Transactional kept in services, guarded Liquibase changesets with rollbacks, a Carbon-only UI with all new intl keys in en.json, and focused tests on both sides.

Requesting changes on three fronts before this can move:
1. **Rebase** — the branch is 44 commits behind develop. #3859 (test catalog) already took the `057` changelog number and #3625 (sample type management) runs through `067`, so the new changelog needs renumbering to `068+`. Good news: `base.xml` is the only textual conflict; App.jsx, en.json, and message_en.properties auto-merge cleanly.
2. **Reconcile with #3890** — R1 of OGC-811 (unified /Results worklist) landed on develop after this branched; please confirm the batch model and pending-test query line up with that work so the two OGC-811 streams don't diverge.
3. **Open bot feedback** — the Codex P1 on lab-unit scoping for `/pending-tests` and batch create is still open and I'd treat it as blocking (existing workplan controllers filter via `userService.filterResultsByLabUnitRoles`). The pre-cap filtering point (P2) and Copilot's notes on the concurrent already-assigned check, the English default batch name, and the `reload()` loading state also remain unaddressed.

Backend CI is red (PatientDAORedirectIntegrationTest setUp constraint violation) — possibly test-ordering fallout, but we need a green run after the rebase. Happy to re-review quickly once rebased; +1 to Samuel's request for a short demo video.

PR #3817, opened 2026-07-03
fix(test): use /api/OpenELIS-Global prefix on LabelMakerServlet URL in spec

pmanko:
## Summary
Follow-up to #3528. The `post-save-print-quantity` Playwright spec was hitting `/LabelMakerServlet` directly instead of going through the nginx `/api` proxy, which silently 200s (SPA fallback) for unrouted paths. The servlet was never actually reached, so the quantity cap appeared not to fire.

Verified locally against the build-flavour stack: spec now passes, servlet returns 400 as expected.

## Test plan
- [x] Verified locally against build-flavour stack — spec passes, servlet returns 400


mozzy11 (commented):
Thanks for the follow-up, @pmanko. Good news: this fix is already on develop — the spec at `frontend/playwright/tests/foundational/core/post-save-print-quantity.spec.ts` (line 58) already uses the `/api/OpenELIS-Global/LabelMakerServlet` prefix, which landed with the #3528 squash-merge (commit 9c34ab57c).

Because #3528 was squash-merged, the three-dot diff shown here re-lists all 20 files from the original branch, but the only genuinely new commit (c8f4f7aa) touches just the spec file, and that change is present on develop verbatim.

Recommendation: rebase onto current develop to confirm — the diff should collapse to empty — and then this can be closed as already delivered. Flagging it as superseded rather than requesting changes, since there is nothing left to fix.

PR #3828, opened 2026-07-08
Harden cold storage monitoring: dead-man's-switch alerting, RBAC, issue #3743/#3904 fixes, and config cleanup

mherman22:
## Summary

Hardens the cold-storage monitoring feature and fixes the defects reported in #3743 and #3904.

- **Dead-man's-switch alerting.** A freezer that stops answering Modbus polling now raises its own `FREEZER_OFFLINE` alert instead of silently logging a warning; previously an offline sensor produced no alert and no notification at all. It fires after 3 consecutive failed polls (not one transient timeout), notifies through the existing **Equipment Alerts** nature — whose Alert Settings description already reads "equipment malfunctions or connectivity issues" — and is **resolved automatically on the first poll the device answers again**, so a later outage raises a fresh alert and a fresh notification rather than silently bumping a duplicate count on a row nobody can clear. Auto-resolution is scoped to `FREEZER_OFFLINE` alerts only: an open temperature-excursion alert is deliberately left untouched when comms return.
- **RBAC across every cold-storage/alert REST endpoint** (only `POST /system-config` and the class-level gate on `AlertNotificationConfigRestController` had authorization before), and corrective-action/alert-acknowledgment attribution now derives from the authenticated session instead of a client-supplied user id. The client hides every ADMIN-only write affordance to match: Device Management's Add/Edit/Enable-disable/Delete, the create-device link inside Corrective Actions' Log Corrective Action modal, Temperature Thresholds' Save, and Alert Settings' Save. The shared `GET /rest/alerts` read also admits RESULTS and VALIDATION, because `CriticalBanner` on the unified results and validation pages reads it.
- **#3743 fixes.** Device delete uses a dedicated `deleted` flag, so a deleted device can no longer be resurrected by the active/inactive toggle. `DELETE /rest/alerts/{id}` clears an alert, restricted to ADMIN **and** to alerts whose `alertEntityType` is `Freezer` — `AlertService.delete` is a hard row delete with no audit trail, and `/rest/alerts` is shared with the critical-result and EQA/referral flows. A missing id answers 404 rather than 500. The timestamp bug is fixed at its source: `OffsetDateTime` fields serialize as ISO-8601 strings instead of ambiguous epoch-seconds numbers.
- **Modbus hardening.** Per-device polling now runs concurrently on a dedicated pool wrapped in a daemon `SecurityContext` (without which every `@Async` alert listener submission is rejected and no cold-storage alert is created at all), with retry backoff, configurable register byte-order/word-count decoding, verified RS-485 driver timing, and `digitalpetri.modbus`/`jSerialComm` bumped to their latest versions.
- **Hysteresis and retention.** `minExcursionMinutes` stops a reading oscillating at a threshold boundary from flapping an alert every poll, while still escalating a genuinely sustained breach **at any poll interval** — the lookback falls back to the most recent prior reading when the profile-derived window is empty, since the poll interval is configured independently of the profile. Plus a `freezer_reading` retention/cleanup job and a uniqueness constraint.
- **Frontend.** A stale-reading indicator (an `Offline` tag when the newest stored reading is more than 15 minutes old, i.e. when polling itself has stopped — a poll that runs and *fails* still stores a reading, and an unreachable device instead shows a red `Critical` tag with `—` for temperature plus a CRITICAL `FREEZER_OFFLINE` row in Active Alerts), live refresh, session-derived user attribution, race-condition guards, corrected role gate, i18n/Carbon token cleanup, and inline corrective-action logging from an alert.

## Follow-up fixes (issue #3904) and config cleanup

- **Device name collision.** `freezer.name`'s unique constraint was unconditional while `findByName` only checked `deleted=false`, so reusing a soft-deleted name hit a raw DB constraint violation the exception handler didn't catch, surfacing as an opaque "internal storage error." Replaced with a partial unique index scoped to non-deleted rows, plus a `LIMSRuntimeException` handler so validation errors return a real message. The same opaque error had a second cause: the auto-derived storage-device code was truncated to 50 characters while the storage code validator caps codes at 10, so any device whose name had more than 10 alphanumeric characters (e.g. "Vaccine Fridge 01") could not be created at all. `createStorageDeviceFromFreezer` now leaves the code unset so `StorageLocationService.insert` derives it through the existing `CodeGenerationService` on its own auto-generate branch — the same path every other auto-created device takes.

  **Neither cause reproduces the reporter's literal sequence** — naming one device `pcr01` and then another `pcr`, with no delete between. That sequence succeeds at this head, measured against a Testcontainers Postgres with the full Liquibase run, and every candidate collision was traced: `generateCodeFromName` yields `PCR01` and `PCR`, `uk_device_code_in_room` is (parent_room_id, code), `short_code` is unmapped by `StorageDevice` so it stays NULL and NULLs do not collide in Postgres, `findByName` is an exact case-insensitive match, and `storage_device.name` carries no unique constraint. What this PR does for that report is make the failure class diagnosable: `handleStorageValidation` turns the opaque 500 into a 409 carrying the real validation message, so a recurrence names its own cause.

  `deleteFreezer` deliberately leaves the auto-created `storage_device` row active. That row is a Sample Storage location, and `SampleStorageServiceImpl.validateLocationActiveForEntity` refuses every sample assignment and move into an inactive device *and everything beneath it*, while `StorageLocationFhirTransform` publishes `active` as the FHIR `Location.status` — neither is a side effect a monitoring-config delete should cause. Deactivating would also buy nothing on screen: `GET /rest/coldstorage/storage-devices` is the only device endpoint that filters on `active`, and its client (`fetchStorageDevices`) is exported but never imported anywhere in `frontend/src`, while Sample Storage's own device list renders inactive devices too, merely tagged. The residual is a `storage_device` row keeping the deleted freezer's name and code — deactivating would not free the code either, since both `uk_device_code_in_room` and `isCodeUniqueForDevice` ignore `active` — which is what `createFreezer_shouldNotReuseADeletedDevicesStorageDeviceCode` pins.
- **Humidity not displaying, and not alerting.** Added a humidity column/metric selector to the dashboard, historical trends chart and device history view — the backend already recorded it, the frontend never surfaced it. Humidity also fed a reading's stored status while raising **no alert**, so a unit could show a red `Critical` tag with nothing behind it: `ReadingIngestionServiceImpl` only ever checked temperature. It now has a humidity counterpart that raises a **new `FREEZER_HUMIDITY` alert type** (`FreezerHumidityThresholdViolatedEvent` → `FreezerAlertService.createFreezerHumidityAlert`), behind the same `minExcursionMinutes` hysteresis gate, since the escalation status already accounts for humidity.

  `FREEZER_HUMIDITY` is a separate type rather than a flag on `FREEZER_TEMPERATURE` because `findDuplicateAlert` keys on `(alertType, entityType, entityId)` — sharing one would collapse a humidity breach and a temperature breach on the same unit into a single row and lose whichever arrived second. **It notifies:** it maps to the existing `FREEZER_TEMPERATURE_ALERT` nature, so a lab that has that nature enabled starts receiving humidity notifications on upgrade, and the Alert Settings row is relabelled *Temperature & Humidity Alerts* to say so. Splitting them would need a new nature, its own constraint value and a new Alert Settings row.
- **Dual temperature probe support.** Added a second, independent temperature register (`temperatureRegister2/Scale2/Offset2`) through read, storage and display: entity, migration, Modbus TCP/RTU read logic, ingestion/persistence, REST DTOs, dashboard column, trends metric, device history and a "Second Temperature Probe (Optional)" form panel. **Threshold evaluation and alerting still run on the first probe only** (`ThresholdEvaluationService` and the excursion summaries read `temperatureCelsius`), so a site that consolidates two registered devices into one dual-probe device gets probe 2 charted but not alerted on; the form's helper text says so.
- **Cross-subnet/VPN disconnections.** Split the TCP connect timeout from the request timeout (connect floors at 5s regardless of the configured request timeout) and wrapped connect failures with host/port/cause detail, so logs distinguish a timeout from a refused or unreachable connection.
- **Config redesign.** Collapsed an 11-property static `FreezerMonitoringProperties` class (4 of them already dead or duplicated elsewhere) down to 2 settings — enabled + Modbus timeout — moved into the same `SiteInformation`-backed, live-editable mechanism the feature already uses for `modbusTcpPort`/`bacnetUdpPort`, so an admin can toggle monitoring or raise the timeout without a restart. Everything else (retries, backoff, pool size, retention, offline-alert threshold) is now a hardcoded internal constant. `freezer.monitoring.enabled` is seeded with `value_type` `boolean` so it renders as a real toggle rather than a free-text box, and changelog `091` repairs that type on databases seeded by an earlier revision of this branch.
- **Cleanup.** Removed the deprecated, read-only "System Settings" tab from cold-storage settings (already banner-flagged "will be removed in a future release") with its now-dead API calls and i18n keys.
- **API surface.** Removed the leftover client-supplied `userId` from `AcknowledgeAlertRequest`/`ResolveAlertRequest` — the controller derives the actor from the session, so the field was silently accepted and ignored. The `notes` field was accepted and ignored the same way: the modal offers a notes box on an OPEN alert and posted it to `/acknowledge`, which returned 200 and discarded it, while the identical field on Resolve persisted as `resolution_notes`. It now persists to `alert.acknowledgment_notes` (changelog `090`), is returned on `AlertDTO`, and renders back in the modal. Because that field is durable and shown back as the operator's own words, the Active Alerts row buttons — which offer no note field — send no note at all instead of a hardcoded English placeholder.

## Test plan

All figures measured at `506ea2c9f`, not carried forward.

- [x] `mvn -o clean install -DskipTests -Dmaven.test.skip=true` — BUILD SUCCESS (Temurin 21)
- [x] `mvn -o test -Dtest='org.openelisglobal.coldstorage.**.*Test,org.openelisglobal.alert.**.*Test'` → **124 run, 0 failures** (87 coldstorage + 37 alert), against a real Testcontainers Postgres with the full Liquibase run
- [x] Adding `org.openelisglobal.storage.**.*Test` → **408 run, 0 failures, 1 pre-existing skip** (`IheMcsdComplianceTest`)
- [x] `npx vitest run src/components/coldStorage` → **22 passing** across 6 files
- [x] `mvn -o spotless:apply` clean; `npm run format` / `npm run build` clean
- [x] `Duplicate key check` and `Translation source-of-truth check` both green (`en.json` had 22 duplicate keys against 0 on `develop`; the stale copies are removed, 8370 keys, 0 duplicates, no value changed)
- [x] Migration check: **four** changelogs — `089-cold-storage-monitoring-hardening.xml`, `090-alert-acknowledgment-notes.xml`, `091-freezer-monitoring-enabled-value-type.xml`, `092-freezer-humidity-alert-type.xml` — included in that order at the end of `3.5.x.x/base.xml` after `088`. **There is a numbering collision with `develop`,** which shipped its own `089-results-unified-route-default-on.xml` while this branch was open; the merge keeps both, since the filenames differ and Liquibase records a changeset as (id, author, filename) — renumbering either would change that identity and re-run it on databases that already applied it. **The ordering invariant is master-changelog-wide, not `base.xml`-wide, and `092` now owns it:** `092`'s `add-freezer-humidity-alert-type` must stay the last changeset anywhere in `base-changelog.xml` that declares `chk_alert_type`, because `base-changelog.xml` includes `liquibase/analyzer/` and `liquibase/qc/` *after* `3.5.x.x/base.xml`. (089 held this invariant before 092 existed; its comment now points at 092.) To re-check rather than trust a count: `git grep -l chk_alert_type -- src/main/resources/liquibase` lists every declaring file, and `092` must be last in include order — neither `analyzer/` nor `qc/` declares it today. `092`'s `IN` list and the `AlertType` enum are the same 16 values, with nothing on either side the other lacks; diff the two rather than counting them by hand.
- [x] Verified at runtime against the `dev.docker-compose.yml` stack: the full offline lifecycle (outage → one OPEN alert → recovery → RESOLVED by the **daemon** user with no new row → second outage → a genuinely new alert → acknowledge-then-recover → resolves with the acknowledgement preserved), the notification captured on the wire at an SMTP sink, the four role gates in both directions as a real Reception user and a real Global Administrator, and the DELETE/acknowledge contracts read back from the database.

### Known limitations

- **`@PreAuthorize` denials now answer 403, and that change is app-wide — read this before merging.** `ControllerSetup`'s `@ExceptionHandler(RuntimeException.class)` was resolving `AccessDeniedException` before Spring Security's `ExceptionTranslationFilter` could translate it, so every denial in OpenELIS surfaced as HTTP 500. This PR adds a dedicated `@ExceptionHandler(AccessDeniedException.class)` returning 403 and lowers the advice's `@Order` to `HIGHEST_PRECEDENCE + 1`. Both are app-wide: `ControllerSetup` is one of two `@ControllerAdvice` beans and covers every controller, so any existing REST client that branches on the 500 for a denied call changes behaviour. Genuine `RuntimeException`s still answer 500, pinned by `ControllerSetupExceptionHandlingTest`. Measured on the running dev stack: a Reception user POSTing `/rest/coldstorage/devices` answered 403 where it previously answered 500, and reads still answered 200.
- `FREEZER_OFFLINE` notifications share the single Equipment Alerts toggle with `EQUIPMENT_FAILURE`; a lab cannot enable one and mute the other without a new nature. A link that flaps with repeated outages now sends one notification per outage where it previously sent none, and nothing rate-limits that.
- Recovery resolution is driven by a single event; if the async resolution throws it is logged and nothing retries, as no reconciling scheduler was added.
- The red `03 Checkpoint - E2E` gate is a pre-existing failure in `esig-result-validation.spec.ts` at the merge base, reproduced on an unrelated branch sharing that merge base. This PR touches no validation or results frontend file.

---

https://claude.ai/code/session_01BTz8nje7SYKawnNs4z2Lq4


src/main/java/org/openelisglobal/coldstorage/service/impl/ThresholdEvaluationServiceImpl.java
```diff
@@ -41,6 +49,112 @@ public ThresholdProfile resolveActiveProfile(Freezer freezer, OffsetDateTime tim
...
+                break;
+            }
+            earliestContinuousBreachTime = reading.getRecordedAt();
+        }
+
+        long breachDurationMinutes = java.time.Duration.between(earliestContinuousBreachTime, timestamp).toMinutes();
+        if (breachDurationMinutes >= minExcursionMinutes) {
+            return instantaneousStatus;
+        }
+        return FreezerReading.Status.NORMAL;
```
mozzy11, line 125: This `NORMAL` is persisted as `freezer_reading.status` by `saveReading`, so an out-of-range reading is stored as in-range until the window elapses. `findExcursions` then groups on that stored status and loses the start of every excursion (and any excursion shorter than `minExcursionMinutes` entirely). Suggest returning the instantaneous status for storage and gating only the alert publication on the hysteresis result.

src/main/java/org/openelisglobal/coldstorage/service/impl/ReadingIngestionServiceImpl.java
```diff
@@ -40,10 +40,19 @@ public void ingest(Freezer freezer, OffsetDateTime recordedAt, BigDecimal temper
...
+        FreezerReading.Status status = determineStatus(freezer, recordedAt, temperature, humidity, transmissionOk,
+                profile);
         FreezerReading savedReading = freezerReadingService.saveReading(freezer, recordedAt, temperature, humidity,
                 status, transmissionOk, errorMessage);
 
+        if (!transmissionOk) {
+            // Dead-man's switch: a failed poll has no temperature to compare against a
+            // threshold, so it can never reach the checks below. It must still surface
+            // as its own alert or a disconnected sensor goes unnoticed indefinitely.
+            publishTransmissionFailedEvent(freezer.getId(), errorMessage, savedReading.getId());
```
mozzy11: Two things here: `handleFreezerTransmissionFailed` is `@EventListener @Async`, not `@TransactionalEventListener`, so the CRITICAL alert can be created even if this `ingest` transaction later rolls back. And unlike the temperature path, offline detection has no persistence window — one transient timeout after retries raises CRITICAL. A consecutive-failure threshold would make the dead-man's switch match the hysteresis you added for temperature.

src/main/java/org/openelisglobal/coldstorage/service/impl/FreezerReadingServiceImpl.java
```diff
@@ -62,4 +66,84 @@ public List<FreezerReading> getRecentReadings(Long freezerId, int limit) {
...
+            return excursion;
+        }
+
+        FreezerReading firstReading = excursionReadings.get(0);
+        FreezerReading lastReading = excursionReadings.get(excursionReadings.size() - 1);
+
+        excursion.put("alertId", firstReading.getId());
+        excursion.put("freezerId", freezer.getId());
+        excursion.put("freezerName", freezer.getName());
+        excursion.put("locationName", freezer.getRoom());
```
mozzy11: `List<Map<String, Object>>` as a service return type puts the contract in string keys — callers can't be type-checked and a renamed key fails silently at runtime. Since this feeds a report, a small `FreezerExcursionData` DTO alongside the other report beans would fit the layered convention better.

src/main/resources/liquibase/3.5.x.x/base.xml
```diff
@@ -135,6 +135,20 @@
   <!-- OGC-949 / OGC-766: cascade test_sample_handling_history when its handling row is deleted -->
   <include relativeToChangelogFile="true" file="050-tsh-history-fk-cascade.xml"/>
 
+  <!-- Cold storage monitoring audit: FREEZER_OFFLINE dead-man's-switch alert type -->
+  <include relativeToChangelogFile="true" file="051-add-freezer-offline-alert-type.xml"/>
```
mozzy11: Numbering collision: develop already ships `051-uom-code-ucum` through `055-seed-molecular-components` and is currently at `067`. These five changesets need renumbering to `068+` so the sequence stays unambiguous — worth doing before the next rebase rather than after.

frontend/src/components/coldStorage/FreezerMonitoringDashboard.jsx
```diff
@@ -846,9 +1158,32 @@ function FreezerMonitoringDashboard({ intl }) {
...
+                                                          "Acknowledge",
+                                                      })}
                                                     </Button>
                                                   )}
+                                                  {isAdminUser && (
+                                                    <Button
+                                                      kind="danger--ghost"
+                                                      size="sm"
+                                                      renderIcon={TrashCan}
+                                                      iconDescription="Delete alert"
```
mozzy11: Hardcoded string — needs an `en.json` key and `intl.formatMessage` like the rest of this file. It's the only one I spotted in the frontend diff.

mozzy11 (commented):
Reviewing as a draft, so comments only — no approval implied.

This is strong work overall. The RBAC sweep is the right shape (`@PreAuthorize` per endpoint matches how `EQAAlertRestController`/`ResultEntryRestController` do it, and method security is enabled in `SecurityConfig`), dropping `createdByUserId`/`updatedByUserId` from the request DTOs in favour of session-derived attribution is exactly right for a compliance record, and the `@JsonFormat(shape = STRING)` fix addresses #3743's timestamp bug at the source rather than papering over it in the UI. Moving `deriveTargetTemperature` out of the controller into `ThresholdEvaluationService` is a good 5-layer correction, the N+1 in `generateDailyLogData` is a real find, and `054` correctly de-dupes before adding the constraint. CI is fully green.

Main thing I'd like reconsidered before this leaves draft: **the hysteresis suppresses the persisted reading status, not just the alert.** `applyMinExcursionHysteresis` returns `NORMAL`, and that value is what `saveReading` writes to `freezer_reading.status` — so a genuinely out-of-range temperature is recorded as NORMAL for the first `minExcursionMinutes`. That propagates into the daily-log report's Status column and, more importantly, into the new `findExcursions`, which groups on `reading.getStatus()`: the opening minutes of every excursion get dropped, and any real excursion shorter than the window becomes entirely invisible in excursion reporting. For a cold-chain audit trail I'd keep the stored status truthful (instantaneous) and apply the hysteresis only to the alert-publishing decision.

Smaller items, all inline: Liquibase numbering (051–055 are already taken on develop, which is at 067), `@EventListener @Async` vs `@TransactionalEventListener`, the offline alert having no consecutive-failure threshold, `List<Map<String, Object>>` crossing the service boundary, and one hardcoded `iconDescription`.

yiyucangqiong:
hello，I have identified the following issues and would like to request one enhancement. Could you please review and address them together?
1、Device Name similarity issue: If device names are similar, for example, first naming one “pcr01” and then naming another “pcr”, it shows an “internal storage error”.
2、Temperature and humidity sensor display issue: I use a thermo-hygrometer to record the room environment, but OpenELIS only displays temperature; humidity does not show up.
Additionally, I would like a modification: I have a temperature recording device with two temperature probes. I hope both temperatures can be read through the temperature registers. Currently, OpenELIS only has one temperature register and one humidity register, so I have to set two different device names to read them separately.
3、Cross-subnet disconnection issue: If the OpenELIS system and the temperature recording device are not in the same network segment, using a virtual server to connect causes frequent disconnections.
Could you please investigate and fix these bugs, and implement the multi-probe support? Unfortunately, I am unable to submit pull requests to your branch. I would appreciate it if you could check whether these issues have already been resolved in the latest version or let me know the planned timeline.Thank you for your support!

yiyucangqiong:
Hi  @mherman22 @mozzy11 ,
As a user, we’re really hoping this PR can be merged soon. It fixes several critical issues for us (device name conflicts, missing humidity, dual temperature probes, and cross-subnet disconnections).
The PR has been ready for review for a while — any idea when it might get final review and be merged into develop?
Happy to help with testing if needed. Thanks a lot!

mherman22 (commented):
This needs work before it merges. The migration's `chk_alert_type` redeclaration drops alert types that already ship on develop, which breaks a shipped feature on upgrade and breaks this PR's own offline alert on a fresh install, and two of the fixes the description claims don't hold end to end.

On scope, so you know what this review is worth: I went through the backend closely (the migration against develop's changelog, ingestion and alerting, the Modbus client, the RBAC sweep, the report paths) and sampled the frontend rather than reading all of it, so treat the frontend as less covered than the rest. The branch is also a long way behind develop now, and that rebase is worth doing before acting on the migration comment rather than after.

src/main/resources/liquibase/3.5.x.x/070-cold-storage-monitoring-hardening.xml
```diff
@@ -0,0 +1,203 @@
...
+    <preConditions onFail="MARK_RAN">
+      <tableExists schemaName="clinlims" tableName="alert"/>
+    </preConditions>
+    <comment>Add FREEZER_OFFLINE to alert_type CHECK constraint so an unresponsive device can raise its own alert.</comment>
+    <sql>
+      ALTER TABLE clinlims.alert DROP CONSTRAINT IF EXISTS chk_alert_type;
+      ALTER TABLE clinlims.alert ADD CONSTRAINT chk_alert_type
+      CHECK (alert_type IN (
+        'FREEZER_TEMPERATURE', 'FREEZER_OFFLINE', 'EQUIPMENT_FAILURE', 'INVENTORY_LOW', 'SAMPLE_TRACKING', 'OTHER',
+        'EQA_DEADLINE', 'SAMPLE_EXPIRATION', 'STAT_UPCOMING', 'STAT_OVERDUE', 'CRITICAL_UNACKNOWLEDGED'
```
mherman22: This list has to be the full set of alert types, not the set this branch happens to know about. `chk_alert_type` is redeclared wholesale by every changeset that touches it, so whichever one runs last decides what the column accepts, and this one omits four types that already ship.

It breaks in both directions today, before any rebase:

On a fresh install, `base.xml` runs this changeset immediately before `070-critical-result-alert-type.xml`, which redeclares the constraint without `FREEZER_OFFLINE`. The dead-man's-switch alert this PR exists to add then fails its insert for every offline device, and since `handleFreezerTransmissionFailed` swallows into `logger.error`, the only symptom is a log line.

On a deployment that already ran `070-critical-result`, this one runs last instead and drops `CRITICAL_RESULT`, so OGC-1022 critical-value acknowledgment stops being able to raise alerts.

After the rebase the list also needs `REQUIRED_BY_DEADLINE`, `REFERRAL_CRITICAL_RESULT` and `REFERRAL_REJECTED`: develop is at `088` now and carries `057-referral-alert-types` plus `086-merge-alert-type-constraint-union`. `086` exists precisely because two branches did this to each other once already, and its comment is worth reading before rewriting this one.

None of it can surface in CI. The test schema is `src/test/resources/postgre-db-init/OpenELIS-Global.sql` plus `hbm2ddl.auto=update`, that dump contains no `alert` table (only `user_alert_map`), and liquibase never runs in tests, so `chk_alert_type` does not exist in the test database at all. `testOfflineAlertRequiresConsecutiveFailures` passes for that reason, not because the insert would be accepted in production.

src/main/java/org/openelisglobal/alert/form/AcknowledgeAlertRequest.java
```diff
@@ -7,6 +7,5 @@
  */
 @Data
 public class AcknowledgeAlertRequest {
-    private Integer userId;
     private String notes; // Optional acknowledgment notes
```
mherman22, line 10: Dropping `userId` from here and from `ResolveAlertRequest` needs the caller updated too, and it wasn't. `frontend/src/components/coldStorage/api.js` still sends `body: JSON.stringify({ userId, notes })` (line 71) and `{ userId, resolutionNotes }` (line 104).

These bodies are deserialized by the bare `new ObjectMapper()` built in `AppConfig.jacksonMessageConverter()` and pushed to the front of the converter list, so `FAIL_ON_UNKNOWN_PROPERTIES` is left at its default of enabled. I ran it against the exact version the pom pins (databind 2.18.9): that payload throws `UnrecognizedPropertyException`, and Spring turns it into a 400 before the handler method is entered, so the controller's own try/catch never gets a look in.

Acknowledging or resolving an alert from the cold storage dashboard therefore just fails. Dropping `userId` from those two `api.js` call sites is the whole fix. The corrective-action calls in the same file were updated correctly, so it really is only these two.

src/main/java/org/openelisglobal/coldstorage/service/impl/FreezerServiceImpl.java
```diff
@@ -176,8 +206,10 @@ public void setDeviceStatus(Long id, Boolean active) {
     @Transactional
     public void deleteFreezer(Long id) {
         Freezer freezer = requireFreezer(id);
-        // Soft delete by setting inactive
-        freezer.setActive(false);
+        // Soft delete via a dedicated flag, distinct from the active enable/disable
+        // toggle, so a deleted device stays out of every list query and its toggle
+        // can no longer resurrect it (issue #3743).
+        freezer.setDeleted(true);
```
mherman22, line 218: Reusing a deleted device's name still fails, one layer further down than the partial index fixes. Deleting only flips `freezer.deleted`; nothing touches the `storage_device` row that `createStorageDeviceFromFreezer` inserted alongside it. Recreating a device with that name in the same room runs `generateDeviceCode` again, which is deterministic, and inserts a second `storage_device` with the same `code` under the same `parent_room_id`, hitting `uk_device_code_in_room` from `001-create-storage-hierarchy-tables.xml` (added there, never dropped anywhere since). `DeviceManagement.jsx` always posts `storageDevice: { type }` with no id plus a `roomId`, so the UI only ever takes that auto-create path, which is the path the reported bug arrived on.

So issue #3904's first item is still reproducible for a delete-then-recreate, and the new `LIMSRuntimeException` handler turns it from an opaque 500 into an opaque 409.

`createFreezer_shouldAllowReusingNameFromADeletedFreezer` passes for two reasons unrelated to the real flow: it builds the `Freezer` with no `storageDevice`, so no device row is created at all, and `StorageDevice`'s `@Table` declares no `uniqueConstraints`, so `hbm2ddl.auto=update` never creates `uk_device_code_in_room` in the test schema either way.

Deactivating or renaming the device row on delete would close it, as would scoping the code uniqueness the way `074` scopes the freezer name. Whichever route, the test needs to go through the auto-create path to be worth anything.

src/main/java/org/openelisglobal/coldstorage/dao/impl/FreezerDAOImpl.java
```diff
@@ -30,20 +30,20 @@ public Optional<Freezer> findByName(String name) {
...
     public List<Freezer> findActiveFreezers() {
-        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice WHERE f.active = true ORDER BY f.name";
+        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice WHERE f.active = true AND f.deleted = false ORDER BY f.name";
         return entityManager.unwrap(Session.class).createQuery(hql, Freezer.class).list();
     }
 
     @Override
     public List<Freezer> getAllFreezers() {
-        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice ORDER BY f.name";
+        String hql = "SELECT DISTINCT f FROM Freezer f LEFT JOIN FETCH f.storageDevice WHERE f.deleted = false ORDER BY f.name";
```
mherman22, line 39: Filtering deleted rows out of `getAllFreezers()` also hides their history from the two report paths that fan out over every device. `FreezerReportServiceImpl.generateDailyLogData` and `FreezerReportDataController.getExcursions` both call `getAllFreezers("")`, so once a device is deleted, a daily log or excursion report covering a period when that device was in service silently stops including it. Before this change the delete only set `active = false` and the readings still appeared.

The rows are all still in `freezer_reading`, just unreachable from reporting, which is an awkward property for a cold chain audit trail: someone tidying up the device list quietly rewrites what an inspector can pull. Keeping deleted devices out of the settings and dashboard lists but leaving the report queries unfiltered would give you both.

src/main/java/org/openelisglobal/coldstorage/service/impl/FreezerReadingServiceImpl.java
```diff
@@ -62,4 +67,83 @@ public List<FreezerReading> getRecentReadings(Long freezerId, int limit) {
...
+        List<FreezerExcursionData> excursions = new ArrayList<>();
+        if (readings.isEmpty()) {
+            return excursions;
+        }
+
+        List<FreezerReading> currentExcursion = new ArrayList<>();
+        FreezerReading.Status currentStatus = null;
+
+        for (FreezerReading reading : readings) {
+            if (reading.getStatus() == FreezerReading.Status.WARNING
```
mherman22: This groups purely on stored status, and a failed poll is persisted as `CRITICAL` by `determineInstantaneousStatus`, so a comms outage comes back as a temperature excursion with null min and max temperatures.

The logic moved here unchanged from `FreezerReportDataController`, so it isn't a regression, but the offline alerting this PR adds makes those runs a normal, expected part of the data rather than a rarity, which is what makes it worth fixing now. `applyMinExcursionHysteresis` already skips readings where `Boolean.FALSE.equals(getTransmissionOk())`; the same guard here would keep the excursion report about temperature.

src/main/java/org/openelisglobal/coldstorage/service/impl/ModbusPollingService.java
```diff
@@ -3,64 +3,155 @@
...
+        this.pollingExecutor = pollingExecutor;
+        LOGGER.info("Freezer Modbus polling service registered");
     }
 
+    /**
+     * Polls every active freezer concurrently on a small dedicated pool (see
+     * {@link org.openelisglobal.coldstorage.config.FreezerPollingExecutorConfig}).
+     *
+     * <p>
+     * Previously this loop was sequential: one slow/unreachable device (up to
```
mherman22: Several of the new comments record what the change did rather than what the code does, and those go stale the first time somebody edits around them. This one opens with "Previously this loop was sequential"; `convertScaledValue` has "Defaults ... match the original hardcoded behavior"; `isWarningTemperature` explains what "an earlier version" did; `FreezerDeviceController` line 181 explains why the math is no longer in the controller. Git holds all of that already. The two `TODO` blocks in `FreezerReportServiceImpl` are the same shape.

Entirely your call, and I'd rather flag it than not: most of the comments in this PR are the good kind that record why, and these ones dilute them.

src/main/resources/liquibase/3.5.x.x/base.xml
mherman22: This came back with the rebase: the branch now carries `070-cold-storage-monitoring-hardening.xml` alongside develop's `070-critical-result-alert-type.xml`, and develop has moved on to `088`. Renumbering to `089+` on the next rebase would settle it. Worth doing before the constraint rewrite in the comment on the changeset itself, since the correct alert-type list depends on which of develop's changesets end up ahead of it.

mherman22:
## Hardening summary — what changed since the last review round

This has been through several clean-context review passes since @mozzy11 last looked at it, each one by a reviewer that hadn't seen the code being written, with every finding either fixed or declined on the record and the result verified against a running stack. The change has moved far enough that an overview seemed more useful than asking anyone to diff seven commits.

### The dead-man's switch didn't actually work — three separate defects

Each of these would have shipped, and none of them failed a test:

- **No cold-storage alert of any kind was being created.** The new per-device polling pool was a plain `Executors.newFixedThreadPool`, so it carried no `SecurityContext`, and `UserContextPropagatingTaskDecorator` rejects an `@Async` submission whose calling thread has no authenticated context. Every alert-listener submission threw and was swallowed into a log line — so not just the new offline alert, but the **pre-existing temperature alerts too**. The pool is now wrapped in a daemon context exactly the way `SchedulerConfig` wraps the `@Scheduled` pool.
- **`FREEZER_OFFLINE` had no notification path.** `mapAlertTypeToNotificationNature` had no case for it, so it fell to `default: return null` and `handleAlertCreated` returned before building a message. The alert row appeared on the dashboard and no email or SMS was ever sent. It now notifies through the existing **Equipment Alerts** nature, whose Alert Settings description already reads "equipment malfunctions or connectivity issues".
- **Nothing ever ended the alert's lifecycle.** Worse, because `findDuplicateAlert` matches OPEN *or* ACKNOWLEDGED, the operator's natural move — the row's Acknowledge button — permanently muted that freezer: every later outage only bumped `duplicate_count` on a row still showing the original start time. The alert now resolves automatically on the first poll the device answers again, so a later outage raises a fresh alert and a fresh notification. Auto-resolution is deliberately scoped to `FREEZER_OFFLINE` only, so an open temperature-excursion alert is left alone when comms come back.

@yiyucangqiong that third one is the one that mattered most for your cross-subnet report — with frequent disconnections, the old behaviour would have alerted exactly once, ever, and then gone quiet.

### The migration would have broken alerts on fresh installs *and* upgrades

`chk_alert_type` is redeclared wholesale by every changeset that touches it, so the last one to run decides what the column accepts. This branch's changeset declared an 11-value list that omitted `CRITICAL_RESULT` and the three referral types, and its filename collided with develop's `070`. Fresh installs would have lost `FREEZER_OFFLINE`; upgrades would have lost critical-result and referral alerts. Renumbered to `089`, the list is now the full 15-value union matching the `AlertType` enum, and it's verified from scratch against a real PostgreSQL — 1091 changesets, exit 0, one INSERT per alert type plus two negative controls to prove the constraint is actually enforcing. It's also been made re-runnable, so it doesn't hard-fail application startup on a database that applied an earlier revision of this branch.

### Threshold alerting could go permanently silent

The `minExcursionMinutes` hysteresis looked back over a window of `2 × minExcursionMinutes` and returned `NORMAL` when that window held no prior reading. Since the poll interval is configured independently of the profile, any deployment polling less often than that window got an empty window on *every* poll — so **no temperature alert was ever raised**, with no exception, no log line, and readings still stored with a truthful `CRITICAL` status so the dashboard looked healthy. Reproduced at a 15-minute poll interval with the default profile. The lookback now falls back to the most recent prior reading, so a sustained breach escalates at any poll interval.

### The RBAC sweep had knock-on effects outside cold storage

`/rest/alerts` is shared infrastructure, not cold-storage-only:

- `CriticalBanner` on the unified results and validation pages reads it, so gating the controller to `RECEPTION`+`ADMIN` silently removed the critical-result acknowledgment box for Results and Validation users. Those roles are admitted again.
- The freezer dashboard was requesting `?entityType=Freezer` with no `entityId`, and the controller only filtered when *both* were present — so it was listing **every alert in the system**, labelled `Freezer <id> / Unknown location`, with a new ADMIN hard-delete button on top. Fixed server-side, and `DELETE` is now scoped to `Freezer` alerts as well as to ADMIN.
- Making the device writes ADMIN-only with no matching client gate meant Reception users saw enabled Add/Edit/Toggle/Delete controls that answered HTTP 500. All four ADMIN-only write affordances are now hidden from non-admins: Device Management, the create-device link inside the Log Corrective Action modal, Temperature Thresholds' Save and Alert Settings' Save.

### Smaller but real

- **The alert detail modal crashed for every user.** Four `<Button>` usages had been added without importing `Button`, so it threw `ReferenceError` on render. `npm run build` doesn't catch an undefined identifier and nothing rendered the component under test, so it survived several rounds and CI.
- Corrective-action update/complete/retract resolved their promises on *any* HTTP status, so a Reception user retracting an action got a green "retracted successfully" toast on a 403 with nothing retracted.
- Two `@ControllerAdvice` classes both sat at `HIGHEST_PRECEDENCE`, which establishes no precedence — whether #3904's "internal storage error" was actually fixed depended on bean registration order.
- Recreating a device under a soft-deleted name still failed on `uk_device_code_in_room`; device-code derivation now goes through the existing `CodeGenerationService`. This also fixes the more common cause of that same opaque error: any device name with more than 10 alphanumeric characters (e.g. "Vaccine Fridge 01") previously couldn't be created at all.
- Acknowledgment notes typed in the modal were accepted with a 200 and silently discarded; they now persist to `alert.acknowledgment_notes`.
- 22 duplicate `en.json` keys were failing the `Duplicate key check` gate. Removed, no value changed.

### Verified at runtime, not just in tests

The offline lifecycle was driven end to end against the `dev.docker-compose.yml` stack with a purpose-built Modbus responder and an SMTP sink, reading the `alert` table at every step: outage → one OPEN row; recovery → RESOLVED by the daemon user with no new row; second outage → a genuinely new alert; acknowledge-then-recover → resolves with the acknowledgement preserved. The notification was captured on the wire. The four role gates were checked in both directions as a real Reception user and a real Global Administrator.

Tests: **124** coldstorage + alert, **408** including storage (1 pre-existing skip), **22** frontend coldStorage.

### Known limitations, deliberately not fixed here

- `ControllerSetup`'s app-wide `@ExceptionHandler(RuntimeException.class)` resolves `AccessDeniedException` before Spring Security can translate it, so **any** `@PreAuthorize` denial in OpenELIS surfaces as HTTP 500 rather than 403. The client gates above avoid provoking it; making 403 real is an app-wide status-contract change that doesn't belong in this PR.
- Probe 2 on a dual-probe device is read, stored and charted but **not threshold-evaluated** — the form's helper text says so. Wiring it through the threshold model needs its own columns and migration.
- `FREEZER_OFFLINE` shares the single Equipment Alerts toggle with `EQUIPMENT_FAILURE`, and a link that flaps now sends one notification per outage with no rate limiting.
- Soft-deleting a freezer deliberately leaves its auto-created `storage_device` row active — deactivating it would refuse sample assignments into that device and everything beneath it, and flip the published FHIR `Location.status`. Reasoning is in the PR description.

## Is this ready to merge?

**From a code standpoint, yes** — the last review round returned zero blocking findings, and the behaviour is verified against a running stack rather than only by tests. But it **can't be merged as it stands**, for two reasons that aren't defects in this change:

1. **It needs an approving review** (`REVIEW_REQUIRED`). @mozzy11 your earlier comments are all addressed — the hysteresis point you raised is fixed at the stored-status level *and* the deeper window bug behind it.
2. **`03 Checkpoint - E2E` is red**, and someone needs to decide whether to merge past it. It is **not** this PR's failure: the single failing test is `esig-result-validation.spec.ts:167` ("Navigate to Validation and find the order"), which fails identically on an unrelated branch sharing this merge base, and the merge base is itself the commit that rewrote that spec. This PR touches no validation or results file — outside `components/coldStorage/` its only frontend changes are one role array in `App.jsx` and `en.json`. Every other check is green, including `Duplicate key check` and `Translation source-of-truth check`.

Branch is currently 0 commits behind `develop`.

PR #3831, opened 2026-07-08
feat(results): capture Target Gene + CT on molecular result rows

HerbertYiga:
Molecular Biology (PCR) analyses need per-result capture of the target gene amplified and its Ct value. Add two columns on the results page, gated to only appear when the loaded page contains at least one row from a molecular test section — non-molecular workflows are unaffected.

Backend: two nullable columns on `analysis` (target_gene, ct_value), mapped on Analysis, populated on TestResultItem read, whitelisted + persisted on save (both analysis-only and full-results update paths).

Frontend: two columns inserted after Test Name in SearchResultForm, filtered out when no row's testSectionName contains "molecular" (case-insensitive). Also fixes a pre-existing bug where addRejectResult used a hard-coded slice index that would silently break on any column addition — now uses findIndex against the "notes" column id.


frontend/src/components/resultPage/SearchResultForm.jsx
```diff
@@ -968,16 +968,32 @@ export function SearchResults(props) {
 
     if (configurationProperties.allowResultRejection == "true") {
       if (columns) {
+        // Insert immediately before "notes" — was previously slice(0, 8), a
+        // hard-coded index that silently broke each time a new column was
+        // added upstream of "notes".
```
ibacher: This is one of the things Claude is bad at. It routinely over-comments. Things like "was previously..." really don't need to be there unless it's something that someone would break again, but this implementation is clear enough in what it does without comment.

```suggestion
```

frontend/src/components/resultPage/SearchResultForm.jsx
```diff
@@ -968,16 +968,32 @@ export function SearchResults(props) {
...
+        const notesIdx = columns.findIndex((c) => c.id === "notes");
+        const insertAt = notesIdx === -1 ? columns.length : notesIdx;
         const updatedList = [
-          ...columns.slice(0, 8),
+          ...columns.slice(0, insertAt),
           resultColumn,
-          ...columns.slice(8),
+          ...columns.slice(insertAt),
         ];
```
ibacher: This is the part that actually feels redundant. I think this can be replaced with:

```js
const updatedList = columns.toSpliced(insertAt, 0, resultColumn);
```

frontend/src/components/resultPage/SearchResultForm.jsx
```diff
@@ -968,16 +968,32 @@ export function SearchResults(props) {
...
         columns = updatedList;
       }
     }
   };
 
+  // Molecular Biology gate: only surface target_gene / ct_value columns when
+  // the current result set contains at least one row whose test section name
+  // looks molecular. testSectionName is a deployment-configurable string
+  // ("Molecular Biology", "molecularBio", "Biologie moléculaire", …) so a
+  // case-insensitive substring is deliberately loose.
```
ibacher: This comment is really verbose and highlights exactly why this implementation is wrong. We shouldn't be relying on `testSectionName` if `testSectionName` is localizable. Note that the `"Biologie moléculaire"` mentioned in the value will not match this. If we need to make decisions based on that metadata then we need to capture it in a principled way.

ibacher: And we should _never_ be relying on translated strings, but on some kind of internal string whose properties we can guarantee.

frontend/src/components/resultPage/SearchResultForm.jsx
```diff
@@ -1105,6 +1135,13 @@ export function SearchResults(props) {
     },
   ];
 
+  // Apply the Molecular Biology gate: drop the two extra columns whenever
+  // no row in the current page carries a molecular test section. Keeps the
+  // grid unchanged for chemistry / hematology / etc.
```
ibacher: This could be shortened to:


```suggestion
  // drop the two extra columns whenever no row in the current page carries a molecular test
```

frontend/src/components/validation/Validation.jsx
```diff
@@ -48,7 +48,17 @@ const Validation = (props) => {
     };
   }, []);
 
-  const columns = [
+  // Molecular Biology gate: mirror the results-entry page — only surface the
+  // target_gene / ct_value columns when the currently-loaded validation page
+  // contains at least one row from a molecular test section. Read-only here
+  // (validation confirms captured values, doesn't re-enter them).
```
ibacher: If we're defining the same function twice than maybe we should come up with a shared version. Also, this feels unnecessarily verbose.

src/main/java/org/openelisglobal/result/action/util/ResultsLoadUtility.java
```diff
@@ -756,6 +756,13 @@ private TestResultItem createTestResultItem(Analysis analysis, ResultInventory t
         testItem.setResultDisplayType(resultDisplayType);
         testItem.setAnalysisMethod(analysisService.getAnalysisType(analysis));
         testItem.setTestMethod(analysisService.getMethodId(analysis));
+        // Molecular Biology fields — gated on testSectionName in the frontend
+        // so non-molecular rows never render the extra inputs.
```
ibacher: This seems redundant information

caseyi:
Thanks for tackling this @HerbertYiga — capturing target gene + Ct for molecular work is a real gap that needs to be addressed. Before we go further on this shape, though, I want to flag that I think the underlying data model is the wrong one, for two related reasons: **where this data comes from** and **how many of them there are per test.**

### 1. This data is analyzer-produced, and the target set is fixed per assay — not free-text typed by a tech

Target gene + Ct on a PCR result isn't something a technician hand-enters; it comes off the instrument (GeneXpert, DT-Prime, QuantStudio, FluoroCycler, etc.). And the set of targets is **defined by the cartridge/assay**, not arbitrary:

- **Xpert MTB/RIF** reports five *rpoB* probes (A–E), each with its own Ct, plus SPC/PCC controls — MTB-detected and RIF-resistance are derived from those probe Cts.
- **Xpert Xpress SARS-CoV-2** reports the N2 and E gene targets.
- **Xpert HIV-1 VL / HBV / HCV** report a quantitative value derived from Ct against calibrators.

So the real requirement is **multiple targets per test, each with a Ct**, from a known per-assay list. A single `analysis.target_gene` / `analysis.ct_value` scalar pair structurally can't hold MTB/RIF's five probes or SARS-CoV-2's two targets — it can only keep one, which silently drops the rest.

We already have a rich GeneXpert integration (`projects/analyzer-profiles/astm/genexpert-astm.json` and the HL7/CSV variants) whose `default_test_mappings` are keyed per target with LOINC + result type. That's the pathway that should be populating these values — the profile just doesn't extract Ct yet.

### 2. The unified Test Catalog already has the right home for this: result components

OGC-949 M1 (#3709) introduced the multi-component result model — [`[041-result-components.xml](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/src/main/resources/liquibase/3.5.x.x/041-result-components.xml)`](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/develop/src/main/resources/liquibase/3.5.x.x/041-result-components.xml). A test can have N `test_result_component` rows, each with its own `code`, `label`, `result_type`, `uom_id`, `significant_digits`, ranges (`result_limits.component_id`), and interpretations. That is exactly a per-target/per-probe value with a unit of "Ct" — and it's already wired into the editor's Sample & Results section (#3826) and backfilled for every existing test.

This PR instead adds a parallel mechanism (`analysis.target_gene`, `analysis.ct_value`, plus a `test.supports_target_gene_ct` boolean gate) that:

- can't represent multi-target assays (the cardinality problem above);
- duplicates result-type/label/unit machinery the component model already provides (Ct is just a numeric component; the gene *is* the component label/code);
- introduces a second gating concept (`supports_target_gene_ct`) that lives outside the component model and has no editor UI — it's only settable via the section-136 backfill; and
- adds a bespoke JRXML column path rather than rendering components generically.

This is also the substance behind @ibacher's review notes: gating on `testSectionName` was the visible symptom, but the deeper issue is that per-test/per-result molecular metadata needs to be captured in the principled model we just built, not bolted onto `analysis`.

### Suggested direction

1. Model target gene + Ct as **result components** on the relevant PCR tests (numeric component, unit "Ct"), configured in the Test Catalog editor.
2. Populate them through the **analyzer import path** — extend the GeneXpert profile mapping to extract per-target Ct into those components — with manual entry falling out of the generic component rendering for free (no new columns, no new flag).
3. Drop `analysis.target_gene` / `analysis.ct_value` / `test.supports_target_gene_ct` in favor of the component tables.

Oops ,you are likely waiting on me in Zoom

src/main/java/org/openelisglobal/analysis/valueholder/Analysis.java
```diff
@@ -88,6 +88,8 @@ public class Analysis extends BaseObject<String> implements NoteObject {
     private ValueHolderInterface method;
     private ResultFile resultFile;
     private String analyzerId;
+    private String targetGene;
```
mozzy11, line 91: i think there was a misunderstanding on how this was supposed to be captured and stored

src/main/java/org/openelisglobal/result/action/util/ResultUtil.java
```diff
@@ -235,10 +235,21 @@ public static void createAnalysisOnlyUpdates(ResultsUpdateDataSet actionDataSet,
...
     }
 
+    // Molecular Biology capture: persist target_gene / ct_value alongside the
+    // other per-analysis metadata. Nulls are permitted so non-molecular rows
+    // that come back through the save path do not overwrite unrelated data.
+    // Package-private so unit tests can verify the mapping without a Spring
+    // context (the enclosing methods depend on SpringContext-injected beans).
+    static void applyMolecularFields(Analysis analysis, TestResultItem testResultItem) {
+        analysis.setTargetGene(testResultItem.getTargetGene());
+        analysis.setCtValue(testResultItem.getCtValue());
```
mozzy11, line 250: Data-loss risk: this copies unconditionally, so callers that build a `TestResultItem` without these fields wipe stored values. `ObservationProvider` (both the FHIR write and delete paths) and `ResultEntryRestController.saveSingleAnalysisResult` do that today. `ResultUtilMolecularFieldsTest.applyMolecularFields_nullsAreCarriedThrough` locks the behaviour in as intended — it should instead only assign when the DTO actually carries a value.

src/main/resources/liquibase/3.5.x.x/053-test-supports-target-gene-ct.xml
```diff
@@ -0,0 +1,48 @@
...
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                SELECT count(*) FROM clinlims.test WHERE supports_target_gene_ct = true
+            </sqlCheck>
+        </preConditions>
+        <comment>Seed the PCR flag on for every test currently in the Molecular Biology section (id=136).</comment>
+        <sql>
+            UPDATE clinlims.test
+               SET supports_target_gene_ct = true
+             WHERE test_section_id = 136;
```
mozzy11, line 44: Hardcoded `test_section_id = 136` is deployment-specific — section ids are not stable across installs or distros, so this either backfills the wrong section or nothing. If a backfill like this is kept it needs to resolve the section by a stable key. The `sqlCheck` precondition above is also redundant: Liquibase already runs a changeset once.

src/main/java/org/openelisglobal/test/valueholder/Test.java
```diff
@@ -105,6 +105,13 @@ public class Test extends EnumValueItemImpl {
 
     private Boolean orderable;
 
+    // Molecular Biology (PCR) per-test opt-in for target_gene + ct_value
+    // capture on the results grid, validation grid, and patient PDF. Gating
+    // per-test (not per-section) keeps the invariant local to the object
+    // whose semantics actually change, and avoids coupling to a display
+    // string. Not nullable at the DB level; treated as boolean here.
+    private boolean supportsTargetGeneCt = false;
```
mozzy11, line 113: Nothing in the test-catalog admin (`TestRestController` / the Test Catalog editor) exposes this flag, so after the one-off backfill it can only be changed by hand-written SQL. A per-test capability gate needs an editor surface to be usable.

src/main/java/org/openelisglobal/result/controller/rest/LogbookResultsRestController.java
```diff
@@ -104,7 +104,10 @@ public class LogbookResultsRestController extends LogbookResultsBaseController {
             "testResult*.resultFile", "testResult*.resultFile.fileName", "testResult*.resultFile.fileType",
             "testResult*.resultFile.base64Content", "testResult*.refer", "testResult*.referralItem.referralReasonId",
             "testResult*.referralItem.referredInstituteId", "testResult*.referralItem.referredTestId",
-            "testResult*.referralItem.referredSendDate" };
+            "testResult*.referralItem.referredSendDate",
+            // Molecular Biology per-analysis capture (frontend renders these
+            // only when the row's testSectionName matches a molecular section).
```
mozzy11, line 109: Comment is stale — gating moved to `supportsTargetGeneCt`, not `testSectionName`. Worth noting too that whitelisting these with no server-side check means a client can write `target_gene`/`ct_value` onto any analysis, including tests where the flag is false; the gate is frontend-only.

frontend/src/components/resultPage/SearchResultForm.jsx
```diff
@@ -1238,6 +1258,34 @@ export function SearchResults(props) {
...
 
+      case "targetGene":
+        return (
+          <TextInput
+            id={"targetGene" + row.id}
+            name={"testResult[" + row.id + "].targetGene"}
+            labelText=""
+            value={row.targetGene ?? ""}
+            onChange={(e) => {
+              row.targetGene = e.target.value;
```
mozzy11, line 1269: Direct mutation of the row object before `handleChange` — every other editable cell in this file (lines 1229, 1257, 1299, 1310, 1374) just calls `handleChange(e, row.id)` and lets state flow through. Mutating `row` in place bypasses React's update path and can leave the grid and form state disagreeing.

src/main/resources/reports/PatientReportCDI_vreduit.jrxml
```diff
@@ -1058,7 +1081,7 @@
 					<textFieldExpression><![CDATA[$F{result}]]></textFieldExpression>
 				</textField>
 				<textField textAdjust="StretchHeight">
-					<reportElement stretchType="RelativeToTallestObject" x="0" y="0" width="153" height="12">
+					<reportElement stretchType="RelativeToTallestObject" x="0" y="0" width="95" height="12">
```
mozzy11, line 1084: This narrowing (testName 153 -> 95, and result 110 -> 78 at line 1068) is unconditional, but the two new fields are wrapped in `printWhenExpression`. On a non-molecular report the columns still shrink and leave a ~120px blank gap, so this regresses every existing patient PDF on this template.

mozzy11 (changes requested):
Thanks for pushing this forward, Herbert — and the per-test `supports_target_gene_ct` flag is the right answer to @ibacher's point about gating on a localizable `testSectionName`.

Unfortunately I have to ask for a change of direction rather than a tweak, because develop has moved under this branch. The model @caseyi described is now merged: `041-result-components.xml` gives N typed `test_result_component` rows per test (each with its own uom), `055-seed-molecular-components.xml` already seeds a **`Ct` unit of measure** plus the N2/E SARS-CoV-2 target components on COVIDPCR, `056-analyzer-component-mapping.xml` adds `analyzer_test_map.component_id` so a GeneXpert target/channel binds straight to a component on import, and the frontend renders them generically (`resultPage/unified/PolymorphicResultCell.tsx`). "Target gene + Ct" is now just "a numeric component with unit Ct" — no new `analysis` columns, no new flag, and multi-target assays (MTB/RIF's five rpoB probes) actually fit.

Mechanically the branch also can't land as-is:
- **CONFLICTING** — a rebase is required, and note `052-`/`053-` are both taken on develop now (`052-component-is-primary.xml`, `053-terminology-component-scope.xml`), so the changesets need renumbering regardless.
- **Data loss (inline below):** `applyMolecularFields` copies unconditionally, so any save path whose `TestResultItem` doesn't carry these fields silently nulls previously captured values — `ObservationProvider` (FHIR write *and* delete) and `ResultEntryRestController.saveSingleAnalysisResult` both do exactly that.
- Backfill hardcodes `test_section_id = 136`; nothing in the test-catalog admin ever sets the flag; and the JRXML narrows testName/result unconditionally so non-molecular reports lose width.

Suggested retarget: configure the target components on the relevant PCR tests and extend the GeneXpert analyzer profile to emit per-target Ct into them. Happy to pair on where the analyzer mapping hooks in — the plumbing you already understand here is most of the work.

mherman22:
Sweep note: this branch is **CONFLICTING** with `develop` and has been since the conflict bot flagged it on 2026-07-11, so it can't be reviewed or merged as-is. Please rebase onto the latest `develop`.

Separately, @caseyi's 2026-07-08 comment raised a question about the **shape** of this change (how Target Gene + Ct are modelled on molecular result rows) rather than the implementation details. That's still unanswered and it's the thing that decides whether the rebase is worth your time — worth replying to that first, then rebasing once the data model is agreed.

HerbertYiga:
@ibacher @caseyi Closing this as its now implemented with a different approach

PR #3840, opened 2026-07-09
feat(inventory): storage integration + admin-managed item types + code-based item PK (OGC-657, OGC-658)

mherman22:
## Summary

Wires inventory lots into the shared sample Storage model, migrates `inventory_item`'s primary key to a human-readable code, and makes the Inventory Reports tab actually work.

- **OGC-657** — storage-assign inventory lots the same way sample items are assigned, by generalizing the existing storage assignment/movement tables instead of building a parallel inventory-only storage system.
- **OGC-658 Part C** — `inventory_item`'s PK moves from an opaque auto-increment integer to a server-generated code.
- **OGC-658 Part A — not in this PR (superseded).** Earlier revisions of this branch added an admin-managed `inventory_item_type` table + CRUD UI. Per the product decision in the review thread, item "type" becomes a free-form tag in the upcoming inventory redesign, so that slice was stripped back out: **the net diff contains no `inventory_item_type` code at all**, and item types are back to the enum-backed `GET /rest/inventory/items/types` exactly as on develop. No migration ships for it: the table never reached develop, so there is nothing to undo there (see the Liquibase note below).

Liquibase: just two changesets — `069-generalize-storage-assignment-occupant` and `070-inventory-item-code-pk`. Both carry an explicit `logicalFilePath`, and their `<include>`s sit next to the `048-test-reagent-link` anchor rather than at the end of `base.xml`: that is the earliest point satisfying the dependency, and it keeps them away from the append point every branch uses, which is what made `base.xml` conflict on every single rebase.

The two corrective changesets that dropped `inventory_item_type` and its orphaned menu row are **gone**. That table was only ever created by this branch's own since-deleted changesets — it never existed on develop (the `idx_inventory_item_type` there is an index on `inventory_item.item_type`, not a table), so shipping a drop for it would have meant every future deployment running no-op migrations forever. Dev or demo databases that ran an earlier revision of this branch keep the stale table and FK until they are reset (`/restart-dev-env --full-reset`).

### OGC-657: Inventory lots in Storage Management

- `sample_storage_assignment` / `sample_storage_movement` gain an `occupant_type` discriminator (`SAMPLE_ITEM` / `INVENTORY_LOT`) with partial unique indexes and exclusivity CHECKs, so both occupant kinds share one set of tables and one audit trail. The separate, non-functional `inventory_storage_location` skeleton is deleted (entity, DAO, service, controller, IT).
- `LocationPickerModal` is generalized from sample-specific to occupant-agnostic (`occupantType` + `occupant` props) and reused by both flows; the Order Entry call site in `SearchResultForm.jsx` is updated to the new prop shape.
- New `InventoryLotStorageRestController` + assignment/movement forms; lots can be assigned and moved with history recorded like sample items.
- **i18n rename (translator-visible):** 5 `storage.picker.sample.*` keys are replaced by occupant-neutral `storage.picker.occupant.*` equivalents, and `storage.picker.heading.moveSample` → `storage.picker.heading.move`.
- The acting user for assignment/movement resolves from the centralized `UserContextHolder` (daemon-user fallback) rather than a hardcoded id — per review, using #3356's new mechanism.

### OGC-658 Part C: Code-based Item primary key

- PK becomes a server-generated `UPPER_SNAKE` code derived from the item name (collision-suffixed `_2`, `_3`, …) via the new shared `CodeGenerator` util, or an explicit code supplied on create.
- Clean-break migration (Inventory has no production data yet): existing rows are cleared and recreated, not backfilled. **The clean break uses explicit `DELETE`s in FK dependency order, never `TRUNCATE … CASCADE`** — per review, CASCADE would have reached `sample_storage_assignment`/`sample_storage_movement` through the new FKs and wiped shipped sample-storage data on upgrade. Only inventory-lot occupant rows are removed from the storage tables; sample-item data is untouched. The rollback block gets the same treatment.
- The Java property stays named `id` (typed `String`) so `BaseObject`'s CRUD contract, the JSON wire format, and the reflection-based reagent-link lookup are unaffected; only the DB column is renamed to `code`.
- Item form gains an optional Code field (auto-generated when blank, locked once created); Catalog gains a Code column and includes it in search.
- Duplicate/malformed codes return a stable `errorCode` (an `en.json` message id) + interpolation `params` in the 400 body via the new `LocalizedValidationException`, so the UI shows a translated message instead of a raw backend string.

### Inventory Reports tab: implemented the backend

The tab had a complete UI — 6 report types, 3 export formats, date/grouping filters — calling `/rest/inventory/reports/generate`, which did not exist; every "Generate" 404'd.

- New `InventoryReportService`/`Impl` builds all 6 types (Stock Levels, Expiration Forecast, Usage Trends, Lot Traceability, Low Stock, Transaction History) over a plain `ReportTable`, reusing existing DAOs/services plus one genuine gap (`InventoryUsageDAO/Service.getByDateRange`).
- `InventoryReportWriter` exports CSV/PDF/XLSX using dependencies already in the repo (iText, Apache POI), following `AuditTrailReportRestController`'s CSV-injection-safe pattern.
- Fixed two real frontend bugs found while wiring it up: the date pickers sent raw `Date.toString()` through `URLSearchParams` instead of ISO dates (would have 400'd every request — extracted a tested `toIsoDate` helper), and the Export Format dropdown was missing `itemToString`, so every option rendered blank.

### Inventory Reports: correctness and content pass

- **Stock Levels / Low Stock counted dead inventory as available** — both summed `currentQuantity` across *every* lot, so `EXPIRED`/`DISPOSED`/`QUARANTINED`/QC-failed stock counted as usable. Added an **Available Quantity** column (via the existing `InventoryLot.isAvailableForUse()`) beside **Total Quantity**, and redefined low stock against available quantity. `InventoryItemService.getLowStockItems()` — which also backs the Dashboard's low-stock tile and `/low-stock` — got the same fix (its N+1 trade-off is documented in the javadoc per review).
- **"Usage Trends" was a raw per-transaction log**, not a trend. Reshaped into a real per-item aggregation (Total Used, Events, Avg/Event, First/Last use), sorted heaviest-first; `TRANSACTION_HISTORY` already covers the raw log.
- **"Expiration Forecast" silently ignored the date range** the UI collects. Now honors it, and adds an Urgency column (`EXPIRED`/`THIS_WEEK`/`THIS_MONTH`/`LATER`).
- Added a Category column and a TOTAL summary row where they were missing.

### Lot numbers: auto-generate or type one

The Receive Lot form required a lot number. It's now optional: leave it blank and the server generates `<ITEM_CODE>_<yyyyMMdd>` (collision-suffixed against that item's existing lots), or type one — kept exactly as entered, since lot numbers are usually manufacturer-provided and shouldn't be forced into `UPPER_SNAKE`. Editing an existing lot still requires one.

### UI fixes from live testing

- **The Dashboard's low-stock tile and badge never worked.** Both compared a lot's quantity against `item.minimumStockLevel` — a field that does not exist on `InventoryItem` (it's `lowStockThreshold`), so the threshold was always `0`, the "Low Stock" metric tile always read 0, and the per-row "Low Stock" badge could never render. Rather than just renaming the field, both now derive from the backend's `/items/low-stock`, which judges low stock against *usable quantity summed across lots* — something a per-lot comparison in the browser cannot reproduce. The Dashboard, that endpoint, and the Low Stock report now agree by construction. The same change collapses the Dashboard's N+1 item fetch (one `getById` per lot) into a single batched `getAll`.
- **Downshift controlled/uncontrolled warnings:** 11 Carbon `Dropdown`s passed `selectedItem={list.find(...)}`, which flips from `undefined` to an object once data loads. All now coalesce to `null`.
- **Missing i18n keys:** `notification.success` and `notification.warning` were referenced ~60 times across the app (inventory *and* coldStorage) but never existed in `en.json` — every success toast logged a `MISSING_TRANSLATION` error and rendered the raw key. Added.
- **React unmount warning:** `LotEntryModal` set state after `await` when the parent closed it mid-flight (`onSave()` → unmount → `finally { setSaving(false) }`). Guarded with an `isMountedRef` cleanup, and the same guard applied to the sibling views that share the pattern (Dashboard, Catalog, ItemForm).

### The QC gate was invisible

Receiving stock left it unusable with no indication why. FEFO only consumes lots whose `qcStatus` is `PASSED`, but the Receive form defaults to `PENDING` — so a freshly received lot showed quantity on hand and then failed consumption with "No available lots for item: X", which reads as *there is no stock*. The QC requirement itself is correct lab governance and is unchanged; what was missing was any signal. Now:

- the lot table has its own **QC Status** column, and a lot awaiting QC reads **Pending QC** in the stock-status column instead of a reassuring "In Stock";
- the Receive form's QC field says lots stay unusable until QC passes;
- the FEFO error names the cause and counts the blocked lots ("N lot(s) with stock are still awaiting QC").

Setting QC status was already possible per lot from the row menu and is already audited as a `QC_TEST` transaction; there is still no dedicated QC worklist, which is worth a follow-up.

### Dead code removed

Alongside the Part A strip, code with zero call sites anywhere in main or test: `InventoryItemDAO.getLowStockItems()` (orphaned by the low-stock fix above), `InventoryLotDAO.getByBarcode/getByQCStatus/getByStatus`, `InventoryLotService.isLotExpired/isLotAvailable`, the unused `InventoryEnums.LocationType` (left over from the deleted storage skeleton), and 24 unused wrapper methods in the frontend `InventoryService.js`. REST endpoints were left alone — that's public API surface, not internal dead code.

## Breaking changes

- `inventory_item.id` → `inventory_item.code` (`VARCHAR(64)`, PK). `inventory_lot.inventory_item_id`, `inventory_usage.inventory_item_id`, and `test_reagent_link.reagent_id` change from `BIGINT` to `VARCHAR(64)` to match.
- `TestReagentLink.reagentId` changes from `Long` to `String` through the DAO/service/controller stack.
- Changeset `070` clears `inventory_item` and its dependents (`inventory_lot`, `inventory_usage`, `inventory_transaction`, `test_reagent_link`, plus `INVENTORY_LOT`-occupant rows in the storage tables). Acceptable because Inventory has no production data; **sample-item storage data is explicitly preserved**.
- 5 `storage.picker.sample.*` i18n keys are replaced by `storage.picker.occupant.*` equivalents (net +21 / −5 keys in `en.json`).

## Test plan

- [x] Backend: full `inventory.**` + `storage.**` + `testreagentlink.**` sweep — **47 suites green** (incl. new `InventoryReportServiceTest`, `InventoryReportRestControllerIT` proving CSV/PDF/XLSX round-trip over real HTTP, `InventoryItemServiceCodeGenerationTest`, `InventoryItemServiceLowStockTest`, `InventoryManagementServiceLotNumberTest`, `SampleStorageServiceInventoryLotIT`/`Test`, `InventoryItemRestControllerIT`, `InventoryLotRestControllerIT`).
- [x] Frontend: **31 inventory vitest tests green** across 6 suites (Catalog, Dashboard, ItemForm, LotDetailsPanel, LotEntryModal, Reports) — including 4 new Dashboard tests for the low-stock fix, verified to fail against the old behaviour before being kept.
- [x] `mvn clean install -DskipTests -Dmaven.test.skip=true` packages clean.
- [x] Fresh dev-DB reset — Liquibase 068/069/070/072 run cleanly end to end.
- [x] Driven in a running dev stack (Items catalog, Receive Lot with auto-generated number, storage assign/move, Reports generation) — the three UI fixes above came out of that pass.
- [ ] Manual smoke in a review app (auto vs explicit code, code locked on edit, duplicate/malformed rejected, lot storage assignment end to end).


caseyi:
@mherman22 — first, thank you for the serious work in this PR. The OGC-657 storage generalization, the reports backend, and the correctness pass on the report content are all excellent.

One heads-up so you don't sink more time where the ground has moved: we made a product decision this week that supersedes the OGC-658 Part A approach — inventory item "type" is becoming a free-form tag rather than a managed table, so the `inventory_item_type` CRUD won't be needed (OGC-658 is now closed as superseded). Please pause on that piece for now.

A new inventory feature spec is landing shortly that covers this, and we'll sort out how to split or rework the PR once it does. Nothing wrong with the work itself — the decision just changed underneath it, and that's on me for the timing.

src/main/resources/liquibase/3.5.x.x/066-inventory-item-code-pk.xml
```diff
@@ -0,0 +1,126 @@
...
+                <columnExists tableName="inventory_item" columnName="code" schemaName="clinlims"/>
+            </not>
+        </preConditions>
+        <comment>
+            Migrate inventory_item's PK from BIGINT id to a VARCHAR(64) code.
+            Clean break: pre-existing inventory_item rows (and everything that
+            transitively references them) are truncated, not migrated.
+        </comment>
+
+        <sql>TRUNCATE TABLE clinlims.inventory_item CASCADE;</sql>
```
mozzy11: TRUNCATE ... CASCADE truncates the entire referencing closure, and by this point 064 has added `fk_assignment_inventory_lot`/`fk_movement_inventory_lot` on `inventory_lot` — so this also truncates `sample_storage_assignment` and `sample_storage_movement`, wiping production sample-storage data on upgrade (invisible on a fresh dev DB). Please truncate an explicit inventory-only table list (inventory_item, inventory_lot, inventory_usage, inventory_transaction, test_reagent_link) and `DELETE FROM sample_storage_assignment/movement WHERE occupant_type = 'INVENTORY_LOT'` instead. The rollback block's TRUNCATE (line 91) has the same problem.

src/main/resources/liquibase/3.5.x.x/base.xml
```diff
@@ -168,4 +168,23 @@
   <include relativeToChangelogFile="true" file="062-result-limits-sample-type-scope.xml"/>
   <include relativeToChangelogFile="true" file="063-collapse-variant-groups.xml"/>
 
+  <!-- OGC-657: generalize sample_storage_assignment/movement to support inventory-lot occupants,
+       and drop the non-functional inventory storage skeleton (inventory_storage_location) -->
+  <include relativeToChangelogFile="true" file="064-generalize-storage-assignment-occupant.xml"/>
```
mozzy11: Develop has since merged #3625, which claimed 064–067 (064-add-whonet-code-to-type-of-sample through 067-sample-type-disposal-instructions). Your changeset ids are unique so DATABASECHANGELOG is safe, but please renumber these files to 068–072 on rebase so the directory keeps one file per sequence number — this include block is also the only real merge conflict.

src/main/java/org/openelisglobal/storage/service/SampleStorageServiceImpl.java
```diff
@@ -1406,4 +1410,379 @@ private SampleItem resolveSampleItem(String identifier) {
...
+            inventoryLotService.get(lotId);
+        } catch (org.hibernate.ObjectNotFoundException e) {
+            throw new LIMSRuntimeException("InventoryLot not found: " + inventoryLotId);
+        }
+        return lotId;
+    }
+
+    private Integer resolveActingUserId(String sysUserId) {
+        if (sysUserId == null || sysUserId.trim().isEmpty()) {
+            return 1; // Default to system user, consistent with the sample-item assign/move gap
```
mozzy11: Understood this mirrors the existing sample-item gap, but develop has since merged #3356, which introduces a centralized UserContext and a real daemon/system user. After rebasing, consider resolving the acting user from UserContext here rather than hardcoding 1 (fine to leave the pre-existing sample paths for a follow-up).

src/main/java/org/openelisglobal/inventory/service/InventoryItemServiceImpl.java
```diff
@@ -60,10 +87,31 @@ public List<InventoryItem> searchByName(String searchTerm) {
...
+     * {@code InventoryReportServiceImpl.buildLowStockReport()}, which had this
+     * exact fix already) rather than in the DAO, since replicating
+     * {@code isAvailableForUse()}'s expiration/status/QC logic correctly in SQL is
+     * more error-prone than reusing the one Java implementation.
+     */
     @Override
     @Transactional(readOnly = true)
     public List<InventoryItem> getLowStockItems() {
-        return inventoryItemDAO.getLowStockItems();
+        return inventoryItemDAO.getAllActive().stream().filter(item -> item.getLowStockThreshold() != null)
```
mozzy11, line 115: Minor: this issues one lot query per active item (N+1). Fine at current inventory scale, and I agree reusing isAvailableForUse() beats re-encoding it in SQL — maybe just note it as a known trade-off so the dashboard low-stock follow-up you mentioned can batch it.

mozzy11 (changes requested):
Thank you for this substantial, well-tested piece of work — the OGC-657 occupant generalization is cleanly designed (discriminator + partial unique indexes + exclusivity CHECKs, shared audit trail), the reports backend and its correctness pass are real improvements, and CI is fully green.

One blocking issue: changeset 066's `TRUNCATE inventory_item CASCADE` executes after 064 has added `fk_assignment_inventory_lot`/`fk_movement_inventory_lot`, so Postgres's cascade closure now includes `sample_storage_assignment` and `sample_storage_movement` — silently wiping shipped sample-storage data on every upgraded instance. Details inline; fresh-DB verification wouldn't have caught this.

Rebase guidance: develop has since merged #3625 (Sample Type Management), which claimed liquibase file numbers 064–067 — only `base.xml` textually conflicts, but please renumber your 064–068 files to 068–072 when rebasing. #3356 (centralized UserContext/daemon user) also landed and is relevant to the hardcoded user-id fallback (inline note).

Per Casey's earlier comment, OGC-658 Part A (inventory_item_type) is superseded, so I'm holding detailed review of that slice pending the split/rework — just noting that Part C's 065 FK on `inventory_item.item_type` will need to move with it. Requesting changes for the TRUNCATE data-loss issue and the changeset renumbering; the rest looks in good shape.

src/main/resources/liquibase/3.5.x.x/066-inventory-item-code-pk.xml
mherman22: Fixed in d14d7f5a2 — great catch, this would have been ugly in production. One wrinkle vs the suggested fix: a plain multi-table TRUNCATE of the inventory list still fails because fk_assignment_inventory_lot/fk_movement_inventory_lot are constraint-level references into inventory_lot (Postgres refuses TRUNCATE regardless of row data). So it's now DELETEs in FK dependency order: sample_storage_movement/assignment WHERE occupant_type = 'INVENTORY_LOT' first, then inventory_usage, inventory_transaction, test_reagent_link, inventory_lot, inventory_item. Rollback block gets the same treatment. Side benefit: any future FK into these tables makes the changeset fail loudly instead of silently truncating the new referrer.

src/main/resources/liquibase/3.5.x.x/base.xml
mherman22: Done — rebased onto latest develop and renumbered to 068–072 (068-generalize-storage-assignment-occupant, 069-inventory-item-type, 070-inventory-item-code-pk, 071-fix-inventory-item-type-last-updated, 072-remove-inventory-item-type-menu-entry). Verified one file per sequence number across the directory and that every base.xml include resolves.

src/main/java/org/openelisglobal/storage/service/SampleStorageServiceImpl.java
mherman22: Done in 340f20a9a — blank sysUserId now resolves via UserContextHolder.getCurrentSysUserId() with a daemon-user fallback, no hardcoded id. Left the pre-existing sample-item paths for the follow-up as suggested.

src/main/java/org/openelisglobal/inventory/service/InventoryItemServiceImpl.java
mherman22: Noted in the javadoc in 180780018 — N+1 documented as a known trade-off with batching as the follow-up path.

mherman22:
Superseded by a five-PR stack — same work, split into reviewable slices, with the review feedback from this PR applied:

- #4015 `inventory_item` PK → server-generated string code
- #4016 inventory lots wired into Storage Management
- #4017 Reports tab
- #4018 catalog + item form on string codes
- #4019 post-review fix pass

Merge bottom-up, starting with #4015. Branches live on this repo, so each PR's diff shows only its own slice.

Two notes carried over from review here: the `TRUNCATE ... CASCADE` data-loss risk @mozzy11 caught is fixed in #4015 (DELETE in FK dependency order), and OGC-658 Part A is dropped entirely per the product decision that item type becomes a free-form tag — the `inventory_item_type` table never existed on develop, so no corrective changeset was needed.

Issue #3862, opened 2026-07-15
Constitution amendment: i18n Key Reuse & Hygiene (Principle VII, v1.11.0)


caseyi:
## Constitution Amendment Proposal — Principle VII: i18n Key Reuse & Hygiene

**Version:** 1.10.0 → 1.11.0 (MINOR — materially expanded guidance) · **PR:** #3861

## Principle change rationale

`en.json` grew **2,385 → 7,133 keys between 2026-01 and 2026-07** (~800 keys/month, largely AI-assisted development minting component-scoped keys instead of reusing existing ones). Audit of `develop` @ `06d531e`:

- **1,786 keys (25%)** duplicate an existing English value — 56× "Status", 34× "Active", 28× "Cancel", 22× "Sample Type"
- **~2,500 keys (35%)** are orphans: referenced nowhere in `frontend/src` (after accounting for 207 dynamic-key prefixes) yet still sent to translators
- **386 ids** referenced at react-intl call sites but missing from `en.json` — render as raw key ids to users today

Every redundant key multiplies into ~20 Transifex locales (~34k unnecessary translation slots); translator capacity is the project's scarcest resource, and ~36% of French strings remain untranslated.

## Amendment summary

Adds a mandatory **Key Reuse & Hygiene** section to Principle VII: search-before-minting (`npm run i18n:find`), `common.*` canonical keys for generic UI strings, no cross-feature key references (promote to `common.*` instead — prevents hidden coupling), context exceptions via a committed allowlist (`i18n-context-exceptions.json`), an `// i18n-keys:` pragma declaring dynamic key families, referenced-ids-must-exist (CI-enforced), and a CI ratchet preventing PRs from increasing duplicate/orphan counts.

## Impact analysis

- **Backward compatibility:** none broken. Rules apply to new keys; existing keys untouched until reference-gated sweeps.
- **Templates requiring updates:** spec-template, plan-template, tasks-template, speckit.implement (core + oe), speckit.analyze, CLAUDE.md, AGENTS.md, i18n-check.yml, .githooks/pre-commit, new .claude/settings.json (PostToolUse hook on en.json edits — enforcement reaches the agent while it works, not just at CI).
- **Enforcement:** three layers — agent-time hook, pre-commit, CI. Ratchet baselines to current counts, so nothing existing breaks on day one.

## Migration plan

1. Fix the 386 missing keys (user-visible bugs; independent of this amendment).
[...]

PR #3861, opened 2026-07-15
docs(constitution): add i18n Key Reuse & Hygiene mandate to Principle VII (v1.11.0)

caseyi:
## Summary

Constitution amendment (MINOR, 1.10.0 → 1.11.0) adding a mandatory **Key Reuse & Hygiene** section to Principle VII (Internationalization First), per the Amendment Process.

**Why:** `en.json` grew **2,385 → 7,133 keys in six months**. Audit of `develop` @ `06d531e` found:

- **1,786 keys (25%)** duplicate an existing English value — 56× "Status", 34× "Active", 28× "Cancel"
- **~2,500 keys (35%)** referenced nowhere in `frontend/src` (orphans translators still work on)
- **386 ids** referenced at react-intl call sites but missing from `en.json` (render as raw ids to users)

Each redundant key multiplies into ~20 locales of Transifex translator work (~34k unnecessary translation slots today).

**What the mandate adds:** search-before-minting (`npm run i18n:find`), `common.*` canonical keys for generic UI strings, no cross-feature key references (promote instead — prevents hidden coupling), a committed context-exception allowlist, an `// i18n-keys:` pragma for dynamic key families, referenced-ids-must-exist, and a CI ratchet so PRs cannot increase duplicate/orphan counts.

**Transifex safety:** existing translations are never reset. Key consolidation follows a translation-preserving migration: add new keys → copy translations old→new via the Transifex API → codemod references → only then delete old keys. Detailed in the amendment's follow-up plan.

Doc-only change: constitution text + SYNC IMPACT REPORT + version footer (also corrects the stale 1.9.1 footer). Companion tooling (find script, i18n-check.yml jobs, `common.*` seed, `.claude` hook) lands in follow-up PRs listed in the sync impact report; the CI ratchet baselines to current counts so nothing existing breaks.

## Screenshots

N/A — documentation-only change.

## Related Issue

Closes #3862 (constitution-amendment proposal per the Amendment Process).

## Other

Full audit data, consolidation map (758 clusters), and reproducible methodology (`i18n_audit.py`) available from @caseyi.

PR #3863, opened 2026-07-15
fix(i18n): add 328 missing translation keys + seed common.* canonical set

caseyi:
## Summary

Fixes **328 i18n keys referenced at react-intl call sites but missing from `en.json`**, and seeds **111 `common.*` canonical keys** for the highest-frequency duplicated UI strings. Purely additive — no existing key or value is modified, no code changes.

**Missing keys (328).** Ids referenced via `formatMessage`/`<FormattedMessage>` that don't exist in `en.json`. Where the call site had no `defaultMessage`, users see the raw key id; where it did, the string is untranslatable (all locales fall back to English). Sources for the English text:

- **304** extracted verbatim from `defaultMessage` at the call site (authoritative — the text the feature author wrote)
- **8** recovered from backend `message_en.properties`, **2** from `en.json` git history
- **14** authored from key names + code context — worth a human skim

**Deliberately excluded:** 49 additional missing ids under `qc.chart.*` / `qc.*` / `validation.*` — those modules are in-flight; their keys should land with the feature branch (please include `defaultMessage` at the call sites so the text is authoritative). Also excluded: 5 call sites passing English text as the id (need call-site fixes, not keys).

**`common.*` seed (111).** Canonical keys for English values that currently exist under ≥5 different keys each ("Status" ×56, "Active" ×34, "Cancel" ×28…). These are the sanctioned reuse targets under the pending Principle VII amendment (#3861 / #3862). They are additive; existing duplicate keys are consolidated onto them in later, translation-preserving PRs.

Note: the warn-only duplicate-value CI check (tooling PR) will flag many of the missing keys as duplicating existing values — expected: they are referenced in code today and must exist; consolidating their call sites is follow-up work.

<details><summary>14 authored strings for review</summary>

`analyzer.form.validation.port.invalid` = "Port must be a number between 1 and 65535" · `error.loading.metrics` = "Error loading metrics" · `error.required.fields` = "Please fill in all required fields" · `error.save.sample` = "Error saving sample" · `error.validation.failed` = "Validation failed" · `genericSample.import.success` = "Import successful" · `loading.label` = "Loading..." · `loading.metrics` = "Loading metrics..." · `program.name.program` = "Program" · `save.error.msg` = "Error saving" · `saving.label` = "Saving..." · `systemDefaultPayloadTemplate.messageTemplate` = "Message template" · `systemDefaultPayloadTemplate.subjectTemplate` = "Subject template"

</details>

## Review follow-up (rebase, 2026-08-13)

Rebased onto current `develop` (`25d9d7a50`), resolving the append-block conflict. Addresses @mozzy11's blocking review:

- **Dropped 4 keys that develop has since added independently**, all with byte-identical values, so nothing else changes: `button.update` and `label.status` (from #3625, called out in the review), plus `customFieldType.page.hierarchy.management` and `customFieldType.page.hierarchy.root`, which landed after the review was written. Original count 332 → **328**.
- Re-verified against current `develop`: **zero literal duplicate keys** in the merged file (the hard-fail `check-duplicate-keys` script run locally passes over all of `frontend/src/languages/*.json`), all 328 missing keys still have a real call site in `frontend/src`, and only `en.json` is touched so the source-of-truth check stays green.
- `en.json` is prettier-clean.

## Screenshots

N/A — string resource additions only. Before: affected components render raw key ids (e.g. `qc.chart.title`); after: intended English text.

## Related Issue

Part of the i18n hygiene workstream: #3862 (constitution amendment, PR #3861).

## Other

Keys append after existing entries (file is not sorted; append keeps the diff reviewable). Transifex will receive these as new source strings on merge via `tx-push`.


frontend/src/languages/en.json
```diff
@@ -7131,5 +7131,448 @@
...
+  "analyzer.qc.controls.hidden": "QC controls hidden from patient view",
+  "analyzer.qc.controls.section.title": "QC Controls ({count})",
+  "audittrail.table.heading": "Order Information",
+  "barcode.label.info.labnumber": "Accession number",
+  "button.close": "Close",
+  "button.collapse": "Collapse",
+  "button.expand": "Expand",
+  "button.home": "Home",
+  "button.new.search": "New Search",
+  "button.update": "Update",
```
mozzy11: `button.update` was added to en.json by #3625 (f45c5a060, merged after this branch was cut). Keeping this line creates a duplicate JSON key post-merge, which the hard-fail Duplicate key check will flag on every later i18n PR. Drop it on rebase — the value is identical.

frontend/src/languages/en.json
```diff
@@ -7131,5 +7131,448 @@
...
+  "label.button.assign": "Assign",
+  "label.button.create": "Create",
+  "label.button.saving": "Saving...",
+  "label.button.view": "View Report",
+  "label.code": "Code",
+  "label.editing": "Editing",
+  "label.label": "Label",
+  "label.name": "Name",
+  "label.panel": "Panel",
+  "label.status": "Status",
```
mozzy11: Same as `button.update` above: `label.status` now exists on develop via #3625 with the identical value "Status". Drop this line on rebase to avoid a duplicate JSON key.

mozzy11 (changes requested):
Solid PR overall — only en.json is touched (source-of-truth check respected), the additions are 443 keys exactly as claimed (332 missing + 111 common.*), there are no duplicates within the additions, and I spot-checked 22 of the 'missing' keys across analyzer/customFieldType/freezer/nce/storage/validationRule/esig namespaces: every one has a real react-intl call site in frontend/src. The 14 authored strings read sensibly against their key names.

One blocking issue: the branch predates f45c5a060 (#3625, merged to develop today), which independently added `button.update` and `label.status` to en.json. There's no textual conflict, so GitHub reports MERGEABLE and the green 'Duplicate key check' is stale (it ran 2026-07-15, before #3625 landed) — but merging as-is puts two literal duplicate JSON keys into develop, and the existing hard-fail check-duplicate-keys job would then break every subsequent PR touching frontend/src/languages/**. Fix: rebase onto current develop and drop these two additions (values are identical, so nothing else changes). Happy to approve after that.

mozzy11:
Rebased onto current `develop` (`25d9d7a50`) — merge conflict resolved, and the duplicate-key issue from @mozzy11's review is fixed.

**What changed since the review:**

| Key | Now in develop via | Value identical? | Action |
|---|---|---|---|
| `button.update` | #3625 (`f45c5a060`) | yes ("Update") | dropped |
| `label.status` | #3625 (`f45c5a060`) | yes ("Status") | dropped |
| `customFieldType.page.hierarchy.management` | landed after the review | yes ("Custom Field Types") | dropped |
| `customFieldType.page.hierarchy.root` | landed after the review | yes ("System Administration") | dropped |

The review flagged 2; a re-scan against current `develop` found 2 more that landed in the interim. All four had byte-identical values, so dropping them changes nothing but the duplicate. Additions: 443 → **439** (328 missing + 111 `common.*`). PR title/body updated to match.

**Re-verified against `25d9d7a50`:**
- Hard-fail `check-duplicate-keys` script run locally over all of `frontend/src/languages/*.json` → no duplicate keys.
- All 328 "missing" keys still have a literal call site in `frontend/src` — none went stale in the month since the branch was cut.
- Only `en.json` is modified (source-of-truth check) and the file is prettier-clean.
- Diff is still a pure append: 440 insertions, 1 deletion (the trailing-comma line).

Authorship preserved on @caseyi.

mozzy11 (approved):
Approving — the blocking issue from my earlier review is resolved.

Rebased onto current `develop`; the four keys that had landed independently with identical values are dropped (`button.update`, `label.status` from #3625, plus `customFieldType.page.hierarchy.management` / `.root` which landed after that review). Additions: 443 → 439.

Re-verified on the current head:
- Duplicate key check passes fresh over all of `frontend/src/languages/*.json` — no literal duplicate keys post-merge.
- All 328 missing keys still resolve to a real react-intl call site in `frontend/src`.
- The 111 `common.*` seed keys are unreferenced by design — #3862 migration plan step 3 seeds them *before* the duplicate-value check activates so the check can always name a legitimate reuse target. Spot-verified the ≥5 claim: 110 of 111 values already exist under ≥5 distinct keys on develop (Status ×61, Active ×35, Actions ×33, Cancel ×28).
- Only `en.json` modified; prettier-clean.

PR #3864, opened 2026-07-15
chore(i18n): key reuse tooling — i18n:find, warn-only CI checks, Claude Code hook

caseyi:
## Summary

Tooling for the i18n Key Reuse & Hygiene workstream (#3862, constitution PR #3861). Everything here is **warn-only / advisory** — no CI job fails, no behavior changes, zero risk to existing translations.

**`npm run i18n:find -- "<text>"`** — fuzzy-searches `en.json` values and prints matching keys, canonical `common.*` matches first. The "search before minting" tool the amendment mandates.

**`npm run i18n:check`** (`frontend/scripts/i18n-check.js`) — shared checker used by CI and the agent hook:

- `--duplicates <base>`: newly added keys whose English value already exists under another key (`common.*` exempt)
- `--missing`: ids referenced at react-intl call sites but absent from `en.json`
- `--hook`: Claude Code PostToolUse mode — feedback lands *while the agent edits* `en.json`, not after the PR is open

**CI** — two new warn-only jobs in `i18n-check.yml` (duplicate-value, missing-key). They annotate PRs with `::warning` and always pass. Flip `WARN_ONLY` in the script to enforce once the `common.*` consolidation baseline lands.

**`.claude/settings.json`** — repo-level PostToolUse hook wiring `i18n-check.js --hook`, so Claude Code sessions get key-reuse violations as immediate feedback. (`.claude/commands|skills|logs` remain gitignored; only `settings.json` is shared. Local `settings.local.json` files are unaffected.)

## Screenshots

N/A — developer tooling only. Example output:

```
$ npm run i18n:find -- "collection date"
CANONICAL (reuse these):
  100%  common.collectionDate = "Collection Date"
Feature-scoped matches (do NOT cross-reference — promote to common.* if you need one):
  100%  collection.date = "Collection Date"
  100%  sample.collection.date = "Collection Date"
```

## Related Issue

#3862 (constitution amendment proposal; PR #3861).

## Other

Verified locally: find ranks canonical keys first; missing-check reports 0 after the missing-keys PR; duplicate-check correctly flags value duplication; hook exits silently for non-en.json edits and degrades gracefully without a git base.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Edit|MultiEdit|Write",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node frontend/scripts/i18n-check.js --hook"
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/.github/workflows/i18n-check.yml
+++ b/.github/workflows/i18n-check.yml
@@ -70,3 +70,25 @@ jobs:
               sys.exit(1)
           print('i18n source-of-truth OK — only en.json modified')
           "
+
+  check-duplicate-values:
+    name: Duplicate value check (warn-only)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Warn when new keys duplicate existing English values
+        run: |
+          git show origin/${{ github.base_ref }}:frontend/src/languages/en.json > /tmp/base-en.json
+          node frontend/scripts/i18n-check.js --duplicates /tmp/base-en.json
+
+  check-missing-keys:
+    name: Missing key check (warn-only)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Warn for ids referenced in code but missing from en.json
+        run: node frontend/scripts/i18n-check.js --missing
--- a/frontend/scripts/i18n-check.js
+++ b/frontend/scripts/i18n-check.js
@@ -1,0 +1,168 @@
+#!/usr/bin/env node
+/**
+ * i18n-check: duplicate-value and missing-key checks for en.json.
+ * Constitution Principle VII (Key Reuse & Hygiene).
+ *
+ * Modes:
+ *   --duplicates <base-en.json>  Warn for keys added (vs base) whose value
+ *                                already exists under another key. common.*
+ *                                keys are exempt (they are the canonical
+ *                                targets duplicates get consolidated onto).
+ *   --missing                    Warn for ids referenced at react-intl call
+ *                                sites in src/ that do not exist in en.json.
+ *   --hook                       Claude Code PostToolUse hook: reads the hook
+ *                                JSON from stdin; if the edited file is
+ *                                en.json, runs the duplicate check against
+ *                                the git develop version and reports via
+ *                                exit code 2 (feedback to the agent).
+ *
+ * All modes are WARN-ONLY for CI right now (exit 0) except --hook, which
+ * exits 2 on findings so the agent self-corrects. Flip WARN_ONLY to false
+ * once the consolidation baseline lands.
+ */
+const fs = require("fs");
+const path = require("path");
+const { execSync } = require("child_process");
+
+const WARN_ONLY = true;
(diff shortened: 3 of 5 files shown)


caseyi:
Note for review: `.gitignore` line 72 (`.claude/*`) blanket-ignores the whole directory, so `.claude/settings.json` was force-added here. If we want repo-shared Claude Code hooks (the point of this file), the cleaner fix is adding `!.claude/settings.json` to `.gitignore` in this PR — happy to push that if preferred. Local `settings.local.json` / `commands/` / `skills/` stay ignored either way.

frontend/scripts/i18n-check.js
```diff
@@ -0,0 +1,168 @@
...
+ *                                en.json, runs the duplicate check against
+ *                                the git develop version and reports via
+ *                                exit code 2 (feedback to the agent).
+ *
+ * All modes are WARN-ONLY for CI right now (exit 0) except --hook, which
+ * exits 2 on findings so the agent self-corrects. Flip WARN_ONLY to false
+ * once the consolidation baseline lands.
+ */
+const fs = require("fs");
+const path = require("path");
```
mozzy11, line 24: This crashes: frontend/package.json declares `"type": "module"`, so this .js file is parsed as ESM and `require` throws `ReferenceError: require is not defined in ES module scope` (verified on Node 20). Rename to i18n-check.cjs (the existing scripts here are .mjs for the same reason) or convert to ESM imports.

.github/workflows/i18n-check.yml
```diff
@@ -70,3 +70,25 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Warn when new keys duplicate existing English values
+        run: |
+          git show origin/${{ github.base_ref }}:frontend/src/languages/en.json > /tmp/base-en.json
+          node frontend/scripts/i18n-check.js --duplicates /tmp/base-en.json
```
mozzy11, line 85: Because of the ESM crash in i18n-check.js, this job (and check-missing-keys below) exits 1 on every run — a hard-fail check despite the warn-only intent, hitting every PR that touches frontend/src/languages/**. It never ran on this PR itself (paths filter), which is why CI here is green.

.claude/settings.json
```diff
@@ -0,0 +1,15 @@
...
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node frontend/scripts/i18n-check.js --hook"
+          }
+        ]
+      }
+    ]
+  }
+}
```
mozzy11, line 15: The ESM crash happens at require() before the en.json file_path guard, so this hook surfaces a Node stack trace on every Edit/MultiEdit/Write of any file, in every session. Update this path too when renaming the script to .cjs.

mozzy11 (changes requested):
The design is right (shared checker, WARN_ONLY flag, ::warning annotations, hook exit-2 semantics), and I verified the check logic works: renamed to .cjs locally, --missing finds 392 ids on current develop and --duplicates correctly flags value duplication, both exiting 0. But as shipped, none of it runs.

Blocking: `frontend/package.json` has `"type": "module"` (line 130, since the Vite migration — present on this PR's own base), so Node parses `frontend/scripts/*.js` as ESM and both scripts crash at `const fs = require("fs")` with `ReferenceError: require is not defined in ES module scope`, exit 1. Consequences: `npm run i18n:find`/`i18n:check` are dead; the two 'warn-only' CI jobs hard-fail on every run, so once merged, every PR touching frontend/src/languages/** gets two red checks — the opposite of warn-only; and the PostToolUse hook errors on every Edit/Write in every Claude Code session (the crash happens before the en.json file_path guard). Fix: rename both to `.cjs` (matching the existing `.mjs` convention in frontend/scripts) and update the three call sites (package.json scripts, i18n-check.yml, .claude/settings.json), or convert to ESM imports.

Non-blocking: (1) the workflow's `paths: frontend/src/languages/**` filter means check-missing-keys never runs on PRs that add call sites without touching en.json — consider adding `frontend/src/**` so regressions are caught; (2) merge after #3863, otherwise the missing-key job emits ~390 warnings and duplicate suggestions can't prefer common.*; (3) on your .gitignore question: force-adding works since tracked files ignore .gitignore, but adding `!.claude/settings.json` under line 72 would make the intent explicit and let future edits/clones behave normally — worth including here.

PR #3650, opened 2026-05-26
test: Add integration tests for Patient and Practitioner FHIR facade

kavya685:
# Pull Requests Requirements
- [x] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [x] The PR title follows conventional commit label standards.
- [x] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [x] The changes include tests or are validated by existing tests.
- [x] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary
Adds comprehensive integration tests for the `PatientProvider` and 
`PractitionerProvider` FHIR facade endpoints, as requested in #3648.

## Changes
Tests Added (34 Total):
  PatientFacadeTest (19 tests)
  PractitionerFacadeTest (15 tests)

* Implemented ResourceNotFoundException throw/rethrow to ensure a clean 404 Not Found response is returned to the client.

  ---

  Test Results
   * Tests Run: 34
   * Failures: 0
   * Errors: 0
   * Build Status: SUCCESS

## Screenshots
tests (2)

## Related Issue
Closes #3648
[Add a link to the related issue or mention it here if applicable]

## Other

[Add any additional information or notes here]

PR #3874, opened 2026-07-16
Fix PractitionerProvider search to query from OpenELIS local database…

joelss-tech:
Fix PractitionerProvider search to query from OpenELIS local database… #3649

Description
This Pull request changes the PractitionerProvider search to query the OpenELIS local database instead of forwarding requests to an external FHIR store.

What this pull  request does
1. Replaces external forwarding with local database queries
OpenELIS now queries its own database for practitioner searches

   No network calls to external FHIR store

2. Adds _lastupdated search support
Clients can now search practitioners by date range

3. Implements database level pagination
Pagination is done at the database level (not in memory)

4. Adds reusable search architecture
PractitionerSearchService orchestrates search logic

    FhirSearchHelper provides reusable utility methods


mozzy11 (commented):
This is strong work, @joelss-tech — local database queries with DB-level pagination and _lastupdated support is exactly what #3649 asked for, and the reusable PractitionerSearchService + FhirSearchHelper split gives us a template for converting the other passthrough providers. Next steps: (1) per the coordination comments, let's align with @Isabirye1515 on #3720 so we land one implementation — reviewing the two side by side this week; (2) I'll approve the CI run so the new PractitionerFacadeTest and ProviderServiceTest execute on our infra; (3) heads-up that #3650 also adds a PractitionerFacadeTest, so whichever merges first, the other rebases. Keeping this open and prioritized for review.

mozzy11 (commented):
Hi @joelss-tech — thanks for this. Two things here are genuinely valuable and I want to keep them: you handle **`_lastUpdated` correctly** (bounds extracted, the real `lastupdated` property, an actual range query) and you push **pagination into the database** with `setFirstResult`/`setMaxResults` — both of which the competing PR #3720 gets wrong. The integration tests for the date-range cases are solid too.

That said, #3649 will convert *all* ResourceProviders to local search, so we're standardizing on **one shared search engine** rather than per-resource implementations, and this PR's shape doesn't get us there: adding a second resource would mean copying the ~100-line waterfall service, adding `getXxxByLastUpdated` through four layers of that entity's DAO/Service stack, and copying the manual Bundle-building block.

There are also FHIR-semantics issues we'd otherwise have to re-fix in every copy:

- **First-match-wins waterfall breaks AND-semantics** — `?family=Doe&_lastUpdated=ge2025-01-01` silently ignores `family`. FHIR search parameters must combine, not dispatch.
- **`Bundle.total` is wrong** — it's set to the page size after pagination, not the total match count, and there are no self/next links, so clients can't page. (`_page` also isn't a FHIR search parameter — returning `IBundleProvider` lets HAPI run the paging protocol correctly.)
- **`activeOnly` is hard-coded and inconsistent** — the `_id` and `identifier` branches filter inactive providers (a search without `?active=` must return all), while the lastUpdated and name branches don't filter at all.
- **`FhirSearchHelper.filterActiveOnly` is a no-op stub** that returns its input unchanged, while the service has its own private duplicate that *does* filter — a latent bug magnet.
- **`_revinclude` silently ignored** — the `@IncludeParam` argument is kept in the signature but never read, so existing consumers regress without any signal.
- **Uncapped `_count`** — a client can request `_count=10000000`; and `Integer.parseInt` on the config property has no fallback.
- The transform moved out of the service into the provider, so the `@Transactional(readOnly=true)` no longer covers the Person access during transform.
- Hygiene: a stray `@Override()` with a blank line in `ProviderServiceImpl` and trailing whitespace that spotless will flag.

**Coordination outcome:** #3720 and this PR create the same classes at the same paths, so they can't both merge. The plan is to use #3720's framework shape (SearchParams objects + a single shared search service + `IBundleProvider`) as the base, and fold **your two correct pieces into it generically**: GE/LE range comparisons go into `BaseDAOImpl`'s `DBComparison` (so every entity gets `_lastUpdated` for free, replacing per-entity `getProvidersByLastUpdated` HQL), and DB-level paging is implemented once in a lazy `IBundleProvider` backed by a count query + paged fetch.

**Would you be up for contributing exactly that piece** — the `BaseDAO` GE/LE + `getCountMatching` + paged-fetch support — as a focused PR? It's the part of this work you've already proven out, it's the keystone the shared engine needs, and it would carry your date-range tests over largely intact.

src/main/java/org/openelisglobal/search/service/PractitionerSearchService.java
```diff
@@ -0,0 +1,97 @@
...
+                Provider provider = providerService.getProviderByFhirId(UUID.fromString(resourceId));
+                List<Provider> result = new ArrayList<>();
+                if (provider != null && (!activeOnly || provider.getActive())) {
+                    result.add(provider);
+                }
+                return result;
+            }
+        }
+
+        if (params.hasLastUpdated()) {
```
mozzy11, line 42: First-match-wins dispatch breaks FHIR AND-semantics: `?family=Doe&_lastUpdated=ge2025-01-01` takes the lastUpdated branch and silently ignores `family`. FHIR search parameters must combine as conjunctive predicates, not dispatch to exactly one handler.

src/main/java/org/openelisglobal/search/service/PractitionerSearchService.java
```diff
@@ -0,0 +1,97 @@
...
+                if (providers.isEmpty()) {
+                    providers = providerService.getAllMatching("npi", identifierValue);
+                }
+                if (providers.isEmpty()) {
+                    providers = providerService.getAllMatching("id", identifierValue);
+                }
+
+                
+                if (activeOnly) {
+                    providers = filterActiveOnly(providers);
```
mozzy11, line 66: `activeOnly` is applied only in the `_id` and `identifier` branches — the lastUpdated and name branches never filter. And a search without `?active=` should per FHIR return all practitioners, yet the provider hard-codes `activeOnly(true)` — inconsistent in both directions.

src/main/java/org/openelisglobal/search/service/PractitionerSearchService.java
```diff
@@ -0,0 +1,97 @@
...
+            String familyName = FhirSearchHelper.extractString(params.getFamily());
+            String searchTerm = FhirSearchHelper.buildSearchTerm(givenName, familyName);
+
+          
+            return providerService.getPagesOfSearchedProviders(pageStart, searchTerm);
+        }
+
+        
+        List<Provider> allProviders = providerService.getAllMatching("active", true);
+        return FhirSearchHelper.applyPagination(allProviders, pageStart, pageSize);
```
mozzy11, line 84: The no-params branch loads every matching Provider into memory via `getAllMatching("active", true)` and paginates in Java, while the lastUpdated branch pages in the DB — the in-memory path defeats the PR's stated goal of database-level pagination.

src/main/java/org/openelisglobal/fhir/providers/PractitionerProvider.java
```diff
@@ -231,11 +238,27 @@ public Bundle searchPractitionerBundle(
...
-            Bundle bundle = util.forwardSearchToFhirStore(request);
+            PractitionerSearchParams params = PractitionerSearchParams.builder().identifier(identifier).given(given)
+                    .family(family).id(id).lastUpdated(lastUpdated).pageStart(FhirSearchHelper.getPageStart(request))
+                    .pageSize(FhirSearchHelper.getPageSize(request)).activeOnly(true).build();
+
+            List<Provider> providers = practitionerSearchService.searchProviders(params);
+
+            Bundle bundle = new Bundle();
+            bundle.setType(Bundle.BundleType.SEARCHSET);
+            bundle.setTotal(providers.size());
```
mozzy11, line 253: `Bundle.total` must be the total match count, not the returned page size, and there are no self/next links so clients can't page. Returning an `IBundleProvider` and letting HAPI drive paging (as #3720 does) fixes both — note `_page` is not a FHIR search parameter. Also `revIncludes` is declared but never read, so existing `_revinclude` consumers regress silently.

src/main/java/org/openelisglobal/fhir/util/FhirSearchHelper.java
```diff
@@ -0,0 +1,108 @@
...
+
+        List<T> pagedResults = new ArrayList<>();
+        for (int i = fromIndex; i < toIndex; i++) {
+            pagedResults.add(items.get(i));
+        }
+
+        return pagedResults;
+    }
+
+    public static <T> List<T> filterActiveOnly(List<T> items) {
```
mozzy11, line 104: `filterActiveOnly` returns its input unchanged — a no-op stub — while `PractitionerSearchService` carries a private duplicate that actually filters. Delete one; a silent no-op named "filter" is a latent bug.

src/main/java/org/openelisglobal/fhir/util/FhirSearchHelper.java
```diff
@@ -0,0 +1,108 @@
...
+    public static int getPageSize(HttpServletRequest request) {
+        String countParam = request.getParameter("_count");
+        if (countParam != null) {
+            try {
+                return Integer.parseInt(countParam);
+            } catch (NumberFormatException e) {
+
+            }
+        }
+        return Integer.parseInt(ConfigurationProperties.getInstance().getPropertyValue("page.defaultPageSize"));
```
mozzy11, line 81: `_count` is uncapped (a client can request `_count=10000000`), and `Integer.parseInt` on the config property has no fallback — a missing/garbled `page.defaultPageSize` throws NumberFormatException on every search.

src/main/java/org/openelisglobal/fhir/constants/FhirConstants.java
```diff
@@ -0,0 +1,21 @@
+package org.openelisglobal.fhir.constants;
+
+public final class FhirConstants {
```
mozzy11, line 3: These constants are never referenced — `PractitionerSearchService` uses raw strings ("externalId", "npi", "active"). Use them or drop the class; several new files here also miss the newline at EOF.

mozzy11 (changes requested):
Thanks @joelss-tech — the two things this PR gets right are real improvements: correct `_lastUpdated` GE/LE range handling and true DB-level pagination (`setFirstResult`/`setMaxResults`).

Three blockers before this can move:
1. **CI never ran** — only the gate check executed; the Build + Test pipeline is awaiting maintainer workflow approval, so there's no green build to merge on.
2. **Search semantics:** the first-match-wins waterfall breaks FHIR AND-semantics, and `Bundle.total` is set to the page size rather than the match count (details inline).
3. **Duplication with #3720:** both PRs create the same classes at the same paths (`PractitionerSearchParams`, `PractitionerSearchService`, `FhirConstants`) — they cannot both merge. The maintainer reviews on this PR already proposed the coordination plan (use #3720's `IBundleProvider` + criteria-engine shape as the base, fold this PR's date-range and DB-paging correctness into it) — please respond to those threads.

PR #3886, opened 2026-07-20
Add TestCalculationService Test

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
Adds an integration test suite for `TestCalculationService`, covering CRUD operations and its lookup/ordering methods against a seeded dataset.

### What's Covered
- `getAll()` — returns all seeded calculations with correct field values
- `insert()` — creates a new calculation and increases the row count
- `update()` — updates an existing calculation's fields
- `delete()` — removes a calculation and confirms it's no longer present
- `getCount()` — returns the correct total count
- `getAllMatching()` — matching by name, including a no-match case
- `getAllOrdered()` — ascending and descending order by name
- `getMatch()` — single match by name, including a no-match case
- `get()` on a non-existent id — throws `ObjectNotFoundException`

### Fixtures
Added `result-calculation-service.xml` with:
- Localization and locale data
- 2 patients (each with a linked person)
- 1 organization, 1 test section, 2 tests
- 2 `calculation` rows (`Calc 1`, `Calc 2`)
- 2 `result` rows
- 3 `result_calculation` rows

### Notes
- Assertions are centralized in a private `assertCalculationState()` helper used across tests.
- Test and fixture only — no production code changes.
## Screenshots
Screenshot from 2026-07-20 17-52-48

## Related Issue
https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3803


src/test/resources/testdata/result-calculation-service.xml
```diff
@@ -0,0 +1,64 @@
...
+    <localization id="2" description="Test Description 2" />
+    <supported_locale id="1" locale_code="en"
+        display_name="English" is_active="true" is_fallback="true"
+        sort_order="1" />
+    <localization_value id="1" localization_id="1"
+        locale="en" value="TB" />
+    <localization_value id="2" localization_id="2"
+        locale="en" value="Chem" />
+
+    <person id="1" last_name="Doe" first_name="John"
```
mozzy11, line 13: Non-blocking: person/patient/organization/test_section/test/result/result_calculation rows are never exercised — every assertion targets only `calculation`, which has no FK on sample_id/test_id (liquibase 2.8.x.x/calculated_value.xml). Each extra table named here is `TRUNCATE ... CASCADE`d on every test, needlessly cascading into shared tables in an order-sensitive suite. Trim to the two `calculation` rows unless you add `result_calculation` tests.

src/test/java/org/openelisglobal/testcalculated/service/TestCalculationServiceTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test
+    public void getMatch_byName_noMatch_shouldReturnEmpty() {
+        Optional<Calculation> result = testCalculationService.getMatch("name", "Does Not Exist");
+
+        assertFalse("No match should be found for a nonexistent name", result.isPresent());
+    }
+
+    @Test(expected = org.hibernate.ObjectNotFoundException.class)
+    public void get_nonExistentId_shouldThrowObjectNotFoundException() {
+        // BaseDAOImpl.get() propagates Hibernate's ObjectNotFoundException
```
mozzy11, line 154: Non-blocking: comment is slightly off — `BaseDAOImpl.get()` returns an `Optional`; the `ObjectNotFoundException` comes from `BaseObjectServiceImpl.get()`'s `orElseThrow`. Fix or drop so it doesn't mislead about where the contract lives.

mozzy11 (approved):
Approving — clean CRUD suite for `TestCalculationService` (a pure `AuditableBaseObjectServiceImpl` passthrough, so generic CRUD is exactly the right scope). Highlights: correct and genuinely-needed `resyncSequence` on `calculation_seq`, order-insensitive `getAll` assertions keyed by id, real asc/desc ordering checks, audit user id set, and the delete test sensibly targets a fresh row rather than a seeded one that `result_calculation` FKs would block. CI green, no conflicts.

Two **non-blocking** notes inline — mainly trimming the fixture, since 7 of its 9 table types are never asserted against and each named table gets truncated per-test. Also please coordinate fixture naming with your #3747 (`result-calculation-service.xml` here vs `result-calculation.xml` there) so the two PRs don't collide on merge.

PR #3902, opened 2026-07-22
test(systemusermodule): Add  integration tests for Permission Module Service Test

### LINKED ISSUE #3901: Add Integration Tests for `PermissionModuleService`
`PermissionModuleService` is a facade that delegates all permission-related operations to one of two underlying services depending on the `permissions.agent` configuration property:

- `"Role"` → delegates to `RoleModuleService` (uses `RoleModule`)
- `"USER"` → delegates to `SystemUserModuleService` (uses `SystemUserModule`)

The delegation logic lives in `PermissionModuleServiceImpl`. Currently, there are no integration tests verifying this delegation behavior or the correctness of the data returned under each agent configuration.

---

### Goal

Add an integration test class `PermissionModuleServiceTest` that verifies all public methods on `PermissionModuleService` under both `"Role"` and `"USER"` agent configurations.

---

### Test Class Location

```
src/test/java/org/openelisglobal/systemusermodule/service/PermissionModuleServiceTest.java
```

### Dataset Location

```
src/test/resources/testdata/permission-module.xml
```

The dataset must be unified — a single DBUnit XML that seeds all interdependent tables (`system_module`, `system_user`, `role`, `role_module`, `system_user_module`) to avoid FK cascading deletion issues from `CLEAN_INSERT`.

---

### Methods to Cover

For each method below, write **two tests**: one with `permissions.agent=Role` and one with `permissions.agent=USER`.

| Method | Notes |
|--------|-------|
| `getAllPermissionModules()` | Assert count, concrete type (`RoleModule` vs `SystemUserModule`), and field values (module ID, permissions, agent ID) |
| `doesUserHaveAnyModules(int userId)` | Assert `true` for a user with mapped modules, `false` for a user with no mapping |
| `getAllPermittedPagesFromAgentId(int agentId)` | Assert exact set of page names returned |
| `getPageOfPermissionModules(int startingRow)` | Assert non-empty result and correct concrete type |
| `getTotalPermissionModuleCount()` | Assert exact count |
| `getAllPermissionModulesByAgentId(int agentId)` | For `Role` agent: assert results and role ID. For `USER` agent: assert `IllegalArgumentException` is thrown (known bug — hardcoded `role.id` attribute used regardless of active agent) |
| `getData(PermissionModule)` | Annotate with `@Transactional` to keep the Hibernate session open. Assert `getSystemModuleName()` on the loaded module |
| `get(String id)` | Assert returned ID, concrete type, and permission flags |

---

### Implementation Notes

- Use reflection to toggle `permissions.agent` via `ConfigurationProperties`:
  ```java
  DefaultConfigurationProperties.OEProperties props =
      (DefaultConfigurationProperties.OEProperties) ReflectionTestUtils
          .getField(ConfigurationProperties.getInstance(), "finalProperties");
  props.setPropertyValue("permissions.agent", "Role");
  ```
- Restore the original value in `@After` to avoid test pollution.
- **Do not modify `PermissionModuleServiceImpl`** or any other production source file to make tests pass — tests must reflect the current behavior of the codebase.
- Avoid `assertNotNull` / `assertNull` — use concrete value assertions.
- `getData` tests require `@Transactional` on the test method to prevent `LazyInitializationException`.

---

### Known Bug to Document via Test

`getAllPermissionModulesByAgentId` is hardcoded to query `role.id` regardless of the active agent. When `permissions.agent=USER`, this causes:

```
IllegalArgumentException: Unable to locate Attribute with the given name [role]
on this ManagedType [SystemUserModule]
```

Document this with a test:

```java
@Test(expected = IllegalArgumentException.class)
public void getAllPermissionModulesByAgentId_ShouldThrow_WhenAgentIsUser() {
    setPermissionsAgent("USER");
    permissionModuleService.getAllPermissionModulesByAgentId(1003);
}
```

---

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Adds integration tests for `PermissionModuleService`, a facade that delegates permission operations to either `RoleModuleService` or `SystemUserModuleService` based on the `permissions.agent` config property.

### Added

- `PermissionModuleServiceTest.java` — 16 tests covering all public delegation methods under both `"Role"` and `"USER"` agent configurations
- `permission-module.xml` — unified DBUnit dataset seeding `system_module`, `system_user`, `role`, `role_module`, and `system_user_module`

### Tests cover

- `getAllPermissionModules` — verifies correct concrete type and field values per agent
- `doesUserHaveAnyModules` — positive and negative cases per agent
- `getAllPermittedPagesFromAgentId` — verifies exact page name sets
- `getPageOfPermissionModules` — verifies correct concrete type per agent
- `getTotalPermissionModuleCount` — exact row count per agent
- `getAllPermissionModulesByAgentId` — Role agent returns correct results; USER agent asserts `IllegalArgumentException` (documents existing bug: method is hardcoded to `role.id` regardless of active agent)
- `getData` — verifies lazy-loaded `SystemModule` name per agent
- `get` — verifies ID, type, and permission flags per agent

## Screenshots

Screenshot from 2026-07-22 22-11-03

## Related Issue
Closes https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3901

## Other

@HerbertYiga @mozzy11


src/test/java/org/openelisglobal/systemusermodule/service/PermissionModuleServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+        props.setPropertyValue("permissions.agent", agent);
+    }
+
+    @Test
+    public void getAllPermissionModules_ShouldDelegateToRoleModuleService_WhenAgentIsRole() {
+        setPermissionsAgent("Role");
+        List<PermissionModule> modules = permissionModuleService.getAllPermissionModules();
+        assertEquals(3, modules.size());
+
+        RoleModule first = (RoleModule) modules.get(0);
```
mozzy11: `getAllPermissionModules()` runs `from RoleModule` with no ORDER BY, so `modules.get(0)` being row 4001 is an unguaranteed row-order assumption (Postgres heap order can change after updates/vacuum). Fetch deterministically: `modules.stream().filter(m -> "4001".equals(m.getId())).findFirst().orElseThrow(...)`, then assert fields. Same issue in the USER variant below.

src/test/java/org/openelisglobal/systemusermodule/service/PermissionModuleServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+    public void getAllPermissionModulesByAgentId_ShouldDelegateToRoleModuleService_WhenAgentIsRole() {
+        setPermissionsAgent("Role");
+        List<PermissionModule> modules = permissionModuleService.getAllPermissionModulesByAgentId(3001);
+        assertEquals(2, modules.size());
+        assertEquals("3001", ((RoleModule) modules.get(0)).getRole().getId());
+        assertEquals("3001", ((RoleModule) modules.get(1)).getRole().getId());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void getAllPermissionModulesByAgentId_ShouldDelegateToSystemUserModuleService_WhenAgentIsUser() {
```
mozzy11: Misleading name + bug enshrined as spec: named `...ShouldDelegateToSystemUserModuleService...` but asserts the call throws `IllegalArgumentException` because `getAllPermissionModulesByAgentId` hardcodes `getAllMatching("role.id", agentId)`. Preferably fix the one-line dispatch bug and assert real results for user 1003; if out of scope, rename (e.g. `..._currentlyThrowsBecausePropertyIsHardcodedToRoleId`) and add an in-code comment linking the issue — the PR-body explanation is invisible to future readers. The pinned exception type is also incidental (whatever Hibernate throws for an unknown property path) and will break on ORM upgrades.

src/test/java/org/openelisglobal/systemusermodule/service/PermissionModuleServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+        setPermissionsAgent("USER");
+        permissionModuleService.getAllPermissionModulesByAgentId(1003);
+    }
+
+    @Test
+    @org.springframework.transaction.annotation.Transactional
```
mozzy11: The method-level `@Transactional` (here and below) exists only to keep the session open so lazy `getSystemModule().getSystemModuleName()` doesn't throw — i.e. the test passes under conditions real non-web callers of `getData()` don't get. Add a comment saying why the annotation is needed, and use the plain imported annotation rather than fully-qualified.

src/test/java/org/openelisglobal/systemusermodule/service/PermissionModuleServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+public class PermissionModuleServiceTest extends BaseWebContextSensitiveTest {
+
+    @Autowired
+    private PermissionModuleService<PermissionModule> permissionModuleService;
+
+    private String originalPermissionsAgent;
+
+    @Before
+    public void setup() throws Exception {
+        executeDataSetWithStateManagement("testdata/permission-module.xml");
```
mozzy11, line 31: Nit: prefer `assertTrue`/`assertFalse` over `assertEquals(true, ...)`/`assertEquals(false, ...)` throughout — clearer failure messages, matches suite style.

mozzy11 (changes requested):
Thanks @Agaba-derrick — 16 real integration tests over the `PermissionModuleService` facade, exercising both `"Role"` and `"USER"` agent configurations with correct singleton restore in `@After`, and the suite genuinely surfaces the `getAllPermissionModulesByAgentId` hardcoded-`"role.id"` bug. Good work overall.

Two things to fix before merge (inline): a row-order assumption on an unordered HQL result (`modules.get(0)`) — a latent flake in an order-sensitive suite — and a test whose name says "ShouldDelegateToSystemUserModuleService" while it actually pins the accidental exception from the bug, with no in-code explanation. Preferably fix the one-line dispatch bug and assert real results; otherwise rename + comment it as documenting the defect.

src/test/java/org/openelisglobal/systemusermodule/service/PermissionModuleServiceTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+    @Test
+    public void getAllPermissionModulesByAgentId_ShouldDelegateToRoleModuleService_WhenAgentIsRole() {
+        setPermissionsAgent("Role");
+        List<PermissionModule> modules = permissionModuleService.getAllPermissionModulesByAgentId(3001);
+        assertEquals(2, modules.size());
+        assertTrue(modules.stream().allMatch(m -> "3001".equals(((RoleModule) m).getRole().getId())));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
mozzy11, line 149: Non-blocking: `expected = IllegalArgumentException.class` pins whatever Hibernate happens to throw for the unknown property path, so an ORM upgrade could fail this spuriously. `expected = RuntimeException.class` would keep the pin while surviving upgrades. Fine to defer.

mozzy11 (approved):
All four review points are addressed: the row-order assumption is gone (deterministic stream/filter fetch in both agent variants), the defect-pinning test is renamed with an in-code NOTE explaining the hardcoded `role.id` dispatch bug, `@Transactional` is plain-imported with a comment on why the open session is needed, and boolean assertions now use `assertTrue`/`assertFalse`. CI is green (Build + Test, E2E) with the branch merged up to current develop.

One optional hardening note inline — non-blocking. Approving.

PR #3919, opened 2026-07-24
issue-3891: Fix default admin login never created on fresh Windows dev checkout (CRLF bug)

### LINKED ISSUE #3891: Bug: Default admin login is never created on a fresh dev-mode clone (silent CreateAdminUserTask failure)
### Section/page affected
Login page (both Legacy UI and React UI) — root cause is in backend admin
bootstrap, not the login page itself.
- https://localhost/api/OpenELIS-Global/LoginPage
- https://localhost/

### Summary
On a completely fresh clone, following the documented "Running OpenELIS
Global2 with docker compose For Development" steps in README.md exactly,
the default admin account (`admin` / `adminADMIN!`) documented in the same
README is never created. Every login attempt returns "Invalid login" because
no `login_user` row for `admin` exists at all — the account was never
bootstrapped, not just misconfigured.

### Steps to reproduce
1. Fork and clone the repo
2. `git submodule update --init --recursive`
3. `cd dataexport && mvn clean install -DskipTests`
4. `cd .. && mvn clean install -DskipTests -Dmaven.test.skip=true`
5. `docker compose -f dev.docker-compose.yml up -d`
6. Wait for all containers to report `Up` / `healthy`
7. Visit `https://localhost/api/OpenELIS-Global/` and log in with
   `admin` / `adminADMIN!` as documented in the README
8. Observe: "Invalid login. Please try again"

### Root cause
`docker compose -f dev.docker-compose.yml logs oe.openelis.org | grep -i admin`
shows:

ERROR -- Class: org.openelisglobal.config.CreateAdminUserTask, Method: createAdminUser, Line: 67, Message: could not create default admin user as password was not complex enough

Tracing this in `CreateAdminUserTask.java` (lines 62-68):

```java
if (!loginService.isHashedPassword(password)) {
    if (PasswordValidationFactory.getPasswordValidator().passwordValid(password)) {
        loginService.hashPassword(login, password);
    } else {
        throw new LIMSException("could not create default admin user as password was not complex enough");
    }
}
```
isHashedPassword() (LoginUserServiceImpl.java) fails to recognize the
bcrypt hash bundled in src/main/resources/adminPassword.txt as an
already-hashed password. The task then falls through to treating the raw
60-character bcrypt string as a plaintext password candidate, which
correctly fails complexity validation, throwing and aborting admin creation.
The exception is caught and only logged (CreateAdminUserTask.java:37), so
startup continues silently with no admin user ever created.

This is compounded by volume/database/dbInit/OpenELIS-Global.sql (the seed
data dev.docker-compose.yml mounts into Postgres) shipping an empty
login_user table — the demo dataset's credentials appear to have been
stripped, relying entirely on CreateAdminUserTask to bootstrap admin on
first startup. Since that task fails, the system ends up with zero usable
accounts.

Prior art

This isn't a new or isolated report — #2481 ("Fixed dev login docs: default
admin password violates password policy") independently diagnosed the same
root cause back in May 2026, describing it as: "the admin bootstrap fails
silently during first startup, leaving the system with no users and making
login impossible for new contributors." That PR proposed a documentation-only
workaround (removing the credentials table from the README and adding a
troubleshooting section) rather than a code fix, and was closed without being
merged. The underlying bug is still present on develop.

Current workaround

Manually inserting a login_user row using the existing bcrypt hash from
adminPassword.txt restores login. This confirms the hash itself is valid —
the bug is specifically in how CreateAdminUserTask/isHashedPassword
classify it.

Severity

3 — this prevents any new contributor from using OpenELIS at all when
following the documented dev-mode setup from a clean clone.

Environment

- OS: Windows 11 + WSL2 (Ubuntu-24.04)
- Docker Desktop, dev.docker-compose.yml
- Branch: develop, fresh clone as of 2026-07-21

Proposed next step

I'd like to attempt a fix targeting CreateAdminUserTask/isHashedPassword
directly (rather than another docs-only patch) — could this be assigned to
me so I can submit a PR?

Pratyay-Patel:
## Problem
On a fresh dev-mode setup on Windows, the default admin account (`admin` / `adminADMIN!`) is never created, so login fails immediately after first boot.

Root cause: `git core.autocrlf=true` checks out `src/main/resources/adminPassword.txt` with CRLF line endings, even though the committed blob only has LF. `CreateAdminUserTask.processPassword()` only stripped a single trailing `\n`, leaving a dangling `\r` glued to the bundled bcrypt hash. That `\r` broke `LoginUserService.isHashedPassword()`'s match, so the hash was treated as a plaintext password candidate, failed the complexity validator, and `createAdminUser()` threw before the user was ever inserted.

Fixes #3891 

## Fix
`processPassword()` now strips any trailing combination of `\r`/`\n` via `replaceAll("[\r\n]+$", "")` instead of only a single trailing `\n`.

## Testing
- Added `CreateAdminUserTaskProcessPasswordTest` covering: trailing `\n`, `\r\n`, `\r`, no trailing char, `admin:` marker stripping, and `$2y`→`$2a` conversion, each in combination with line-ending variants.
- Verified end-to-end on a genuinely fresh dev DB (`docker compose down -v && up -d`, rebuilt WAR): boot logs no longer show `could not create default admin user as password was not complex enough`; `login_user` table shows the `admin` row with a correctly-formed `$2a$` hash and no stray `\r`; successful login with `admin`/`adminADMIN!` on the Legacy UI confirmed via browser session.
- React UI login was not independently re-verified due to a pre-existing, unrelated local Vite/WebSocket dev-server proxy issue that prevents the React UI from loading in this environment; this is unaffected by and unrelated to this change (the change is backend-only and never touches the frontend/proxy layer). The Legacy UI test exercises the same backend authentication path.
- Full `mvn clean install` (with tests) could not be run locally: it fails at test-fork startup on Windows due to a pre-existing, unrelated case-sensitivity collision between `src/main/java/.../systemuser` and `src/test/java/.../systemUser` (Windows' filesystem is case-insensitive; Linux CI is not, so this does not affect the actual build check on this PR). Verified instead via the isolated test run above plus full manual end-to-end testing.


mozzy11 (approved):
Correct root-cause analysis and a well-scoped fix: `replaceAll("[\r\n]+$", "")` strips exactly the trailing line-ending characters that a `core.autocrlf=true` checkout introduces, without touching any other whitespace a mounted plaintext `adminPassword.txt` might legitimately contain. The relaxation of `processPassword` to package-private for testability is fine, and the test matrix (LF/CRLF/CR/none, `admin:` marker, `$2y`->`$2a`) locks the contract well. Backend Build + Test is green (the new unit tests ran on CI). The E2E checkpoint is still running at review time — since every E2E suite boots a fresh DB and logs in as admin, it exercises this exact path, so please confirm it lands green before merging.

Note: #3925 addresses the same issue (#3891); this PR is the preferred fix (opened first, narrower normalization) and #3925 has been asked to close in its favor.

Optional follow-up (non-blocking): a `.gitattributes` entry such as `src/main/resources/adminPassword.txt -text` would stop Git from rewriting the file's line endings in the first place — worth adding as defense-in-depth, though the runtime normalization here is still the right fix since deployments can mount their own password file.

Pratyay-Patel:
Re: the .gitattributes suggestion
Makes sense as defense-in-depth — I'll follow up with that .gitattributes entry in a separate small PR after this merges. Thanks for confirming the fix and test matrix!

PR #3922, opened 2026-07-24
fix: record actual acting user in sample storage audit trail

### LINKED ISSUE #3921: Sample storage assign/move audit trail always records user ID 1 instead of the actual user
While working on the storage module I noticed that assignSampleItemWithLocation and moveSampleItemWithLocation in SampleStorageServiceImpl hardcode the acting user to 1:
assignment.setAssignedByUserId(1); // Default to system user for tests
...
movement.setMovedByUserId(1); // Default to system user for tests

This happens in both methods, in both the assignment record and the movement audit record (4 places total). Neither method even accepts a user ID parameter, so there's no way for the actual logged-in user to be recorded — every single assign/move in the system gets attributed to user 1, no matter who actually did it.

You can see this makes the audit trail meaningless: if you check getSampleItemMovementsWithUserNames, every row will show the same user regardless of who performed the action.

For comparison, disposeSampleItem in the same class does this correctly — it takes a sysUserId param, validates it's present and numeric, and uses it for both the audit row and the movement record. The assign/move methods never got the same treatment.

Where this needs fixing:
- SampleStorageService interface — both method signatures
- SampleStorageServiceImpl — the 4 hardcoded 1s
- SampleStorageRestController — /assign and /move endpoints need to resolve the real user via ControllerUtills.getSysUserId(request) (same as the dispose endpoint already does)
- SampleEditServiceImpl.persistSampleStorageLocation — already has sysUserId in scope, just needs to pass it through
- SampleStorageAssignmentListener — has updateData.getCurrentUserId() available for the same purpose

kavya685:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
Closes #3921

assignSampleItemWithLocation and moveSampleItemWithLocation hardcoded
assignedByUserId/movedByUserId to 1, so storage assignment and movement
audit rows never reflected who actually performed the action.

- Added sysUserId param threaded through the service interface, impl,
  and all call sites (SampleEditServiceImpl, SampleStorageAssignmentListener)
- Resolved the acting user in SampleStorageRestController via
  ControllerUtills.getSysUserId(request), returning 401 if unresolved
- Validated sysUserId is present and numeric before use
- Updated existing tests; added a mock session to
  SampleStorageRestControllerFlexibleAssignmentTest since the
  assign/move endpoints now require authentication

## Related Issue
Fixes #3921


kavya685:
Hi @pmanko @mozzy11, this PR is ready for review

mozzy11 (dismissed):
Solid fix for the hardcoded `assignedByUserId`/`movedByUserId = 1`. Key points verified:

- The acting user is resolved server-side via `ControllerUtills.getSysUserId(request)` (session `UserSessionData`, falling back to `UserContextHolder`) — never from client-supplied input, so the audit trail can't be spoofed.
- All four call sites of `assignSampleItemWithLocation`/`moveSampleItemWithLocation` are threaded consistently: both REST endpoints, `SampleEditServiceImpl.persistSampleStorageLocation` (using the in-scope `sysUserId`), and `SampleStorageAssignmentListener` (using `updateData.getCurrentUserId()`, which is always populated by the authenticated order flow).
- Returning 401 when no user resolves is the right hardening for the assign/move endpoints, and the service-layer null/numeric validation gives defense-in-depth for non-HTTP callers.
- Tests updated appropriately, including the mock `UserSessionData` session now that the endpoints require authentication.

Backend Build + Test is green (including the updated storage controller/service tests against a real DB). The E2E checkpoint is still running at review time — it exercises order creation with storage auto-assignment, so please confirm it lands green before merging.

One non-blocking nit: the identical sysUserId validate-and-parse block is duplicated in `assignSampleItemWithLocation` and `moveSampleItemWithLocation` — a small private helper in the impl would remove the duplication if you touch this again.

kavya685:
> Solid fix for the hardcoded `assignedByUserId`/`movedByUserId = 1`. Key points verified:
> 
> * The acting user is resolved server-side via `ControllerUtills.getSysUserId(request)` (session `UserSessionData`, falling back to `UserContextHolder`) — never from client-supplied input, so the audit trail can't be spoofed.
> * All four call sites of `assignSampleItemWithLocation`/`moveSampleItemWithLocation` are threaded consistently: both REST endpoints, `SampleEditServiceImpl.persistSampleStorageLocation` (using the in-scope `sysUserId`), and `SampleStorageAssignmentListener` (using `updateData.getCurrentUserId()`, which is always populated by the authenticated order flow).
> * Returning 401 when no user resolves is the right hardening for the assign/move endpoints, and the service-layer null/numeric validation gives defense-in-depth for non-HTTP callers.
> * Tests updated appropriately, including the mock `UserSessionData` session now that the endpoints require authentication.
> 
> Backend Build + Test is green (including the updated storage controller/service tests against a real DB). The E2E checkpoint is still running at review time — it exercises order creation with storage auto-assignment, so please confirm it lands green before merging.
> 
> One non-blocking nit: the identical sysUserId validate-and-parse block is duplicated in `assignSampleItemWithLocation` and `moveSampleItemWithLocation` — a small private helper in the impl would remove the duplication if you touch this again.

I have addressed the nit

mozzy11 (changes requested):
Thanks @kavya685 — the bug is real (the hardcoded `assignedByUserId(1)`/`movedByUserId(1)` are still on develop) and your diagnosis is correct. However, I'm dismissing my earlier approval: develop merged #3356 yesterday, which centralises user identity in `UserContextHolder` and deliberately removed the `getSysUserId(request)` + manual 401 pattern from the sibling `StorageLocationRestController` — this PR re-introduces exactly that pattern and threads `sysUserId` through the service interface by hand.

Please rework on top of the existing mechanism instead — it makes the diff much smaller:

```java
// SampleStorageServiceImpl
@Autowired
private UserContextHolder userContextHolder;

// in assignSampleItemWithLocation(...) and moveSampleItemWithLocation(...),
// replace the hardcoded values:
Integer actingUserId = Integer.valueOf(userContextHolder.requireSysUserId());
...
assignment.setAssignedByUserId(actingUserId);
...
movement.setMovedByUserId(actingUserId);
```

`requireSysUserId()` fails loudly when no context is available (no silent fallback), and since #3356's `UserContextPropagatingTaskDecorator` propagates the security context to async/listener threads, this also covers `SampleStorageAssignmentListener` and the `SampleEditServiceImpl` path with **no signature changes at all**.

Concretely:
1. Revert the `SampleStorageService`/`Impl` signature changes and the new `validateAndParseSysUserId` helper — resolve the user inside the service as above.
2. Revert the `SampleStorageRestController` changes (the `HttpServletRequest` params and 401 blocks) — an unauthenticated call now fails in the service; if you want a friendlier 401 at the controller, use the no-arg `ControllerUtills.getSysUserId()` from #3356 rather than the request-based overload.
3. Revert the `SampleEditServiceImpl` and `SampleStorageAssignmentListener` call-site changes.
4. Tests: instead of the `MockHttpSession`, authenticate the security context (e.g. set a test `Authentication` whose name matches a seeded login user, or keep the session approach — `ControllerUtills` still reads it — but the service-level tests should exercise `UserContextHolder`).

Optional but nice: migrate `disposeSampleItem(..., sysUserId)` (from OGC-738) to the same `UserContextHolder` pattern in this PR so the whole service resolves identity one way.

kavya685:
> Thanks @kavya685 — the bug is real (the hardcoded `assignedByUserId(1)`/`movedByUserId(1)` are still on develop) and your diagnosis is correct. However, I'm dismissing my earlier approval: develop merged #3356 yesterday, which centralises user identity in `UserContextHolder` and deliberately removed the `getSysUserId(request)` + manual 401 pattern from the sibling `StorageLocationRestController` — this PR re-introduces exactly that pattern and threads `sysUserId` through the service interface by hand.
> 
> Please rework on top of the existing mechanism instead — it makes the diff much smaller:
> 
> ```java
> // SampleStorageServiceImpl
> @Autowired
> private UserContextHolder userContextHolder;
> 
> // in assignSampleItemWithLocation(...) and moveSampleItemWithLocation(...),
> // replace the hardcoded values:
> Integer actingUserId = Integer.valueOf(userContextHolder.requireSysUserId());
> ...
> assignment.setAssignedByUserId(actingUserId);
> ...
> movement.setMovedByUserId(actingUserId);
> ```
> 
> `requireSysUserId()` fails loudly when no context is available (no silent fallback), and since #3356's `UserContextPropagatingTaskDecorator` propagates the security context to async/listener threads, this also covers `SampleStorageAssignmentListener` and the `SampleEditServiceImpl` path with **no signature changes at all**.
> 
> Concretely:
> 
> 1. Revert the `SampleStorageService`/`Impl` signature changes and the new `validateAndParseSysUserId` helper — resolve the user inside the service as above.
> 2. Revert the `SampleStorageRestController` changes (the `HttpServletRequest` params and 401 blocks) — an unauthenticated call now fails in the service; if you want a friendlier 401 at the controller, use the no-arg `ControllerUtills.getSysUserId()` from #3356 rather than the request-based overload.
> 3. Revert the `SampleEditServiceImpl` and `SampleStorageAssignmentListener` call-site changes.
> 4. Tests: instead of the `MockHttpSession`, authenticate the security context (e.g. set a test `Authentication` whose name matches a seeded login user, or keep the session approach — `ControllerUtills` still reads it — but the service-level tests should exercise `UserContextHolder`).
> 
> Optional but nice: migrate `disposeSampleItem(..., sysUserId)` (from OGC-738) to the same `UserContextHolder` pattern in this PR so the whole service resolves identity one way.

Done. Please Check!

mherman22 (approved):
Rework looks complete — thanks @kavya685. All five points from the previous review are addressed: the `SampleStorageService`/`Impl` signature changes and the `validateAndParseSysUserId` helper are gone in favour of `@Autowired UserContextHolder` + `Integer.valueOf(userContextHolder.requireSysUserId())` resolved inside the service; the controller no longer takes `HttpServletRequest` and now uses the no-arg `ControllerUtills.getSysUserId()` from #3356; the `SampleEditServiceImpl` and `SampleStorageAssignmentListener` call-site changes are reverted (the diff no longer touches those files); the unit tests mock `UserContextHolder` and `testDisposeSampleItem_NoUserContext_ThrowsException` covers the fail-loud path, while the integration test authenticates via `authenticateAs("testuser")`; and you took the optional item, migrating `disposeSampleItem` to the same mechanism so the whole service resolves identity one way.

I verified the fix is complete rather than partial: develop has exactly four hardcoded `setAssignedByUserId(1)`/`setMovedByUserId(1)` occurrences in `SampleStorageServiceImpl` (lines 770, 812, 1017, 1068) and this PR replaces all four. The acting user is never read from client input, so the audit trail can't be spoofed, and `requireSysUserId()` throws rather than silently stamping the system user. I also checked the one threading risk — `SampleStorageAssignmentListener` is a plain synchronous `@EventListener`, not `@Async`, so it runs on the caller's request thread with the security context intact; daemon callers resolve to `daemonSystemUser`. Resolving `actingUserId` before any mutation in both `assignSampleItemWithLocation` and `moveSampleItemWithLocation` is the right placement. CI is green. Approving.

kavya685:
@mozzy11 @mherman22, while working on this PR I noticed there are still other controllers across the codebase using the old getSysUserId(request) pattern instead of the no-arg version from #3356. Is this already being tracked somewhere? Happy to pick up a small module as a follow-up if it's not already spoken for and would be useful.

kavya685:
Hi @mozzy11, just checking in on this one — I reworked it on top of UserContextHolder like you suggested, and @mherman22 already went through and confirmed all 4 hardcoded spots are handled, CI's green too. Whenever you get a chance, would you mind taking another look?

PR #3923, opened 2026-07-24
feat: populate patient info on validation screen (#2893)

### LINKED ISSUE #2893: Add more data points on the validation screen
We should have more data points to show on the validation screen because they guide the process of validation. 
Add **age, sex**

ketanop321:
Move patient info (name, national ID, gender, DOB) population from the controller into ResultsValidationUtility where it belongs per the layered architecture. Use a per-sample cache to prevent N+1 queries, and respect the DepersonalizedResults flag for privacy.

The frontend (Validation.jsx) already renders row.patientName and row.patientInfo — only the backend was missing.

Changes:
- ResultValidationItem: add patientInfo field + getter/setter
- ResultsValidationUtility: add SampleHumanService injection, populatePatientInfo() with per-sample cache, copy patientInfo through testResultItemToAnalysisItem and condensed item handling
- AccessionValidationRestController: remove N+1 patient info loop

Closes #2893

# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other

### DIFF
--- a/src/main/java/org/openelisglobal/resultvalidation/util/ResultsValidationUtility.java
+++ b/src/main/java/org/openelisglobal/resultvalidation/util/ResultsValidationUtility.java
@@ -67,6 +67,7 @@ import org.openelisglobal.resultvalidation.action.util.ResultValidationItem;
 import org.openelisglobal.resultvalidation.bean.AnalysisItem;
 import org.openelisglobal.sample.service.SampleService;
 import org.openelisglobal.sample.valueholder.Sample;
+import org.openelisglobal.samplehuman.service.SampleHumanService;
 import org.openelisglobal.spring.util.SpringContext;
 import org.openelisglobal.statusofsample.util.StatusRules;
 import org.openelisglobal.test.service.TestSectionService;
@@ -108,6 +109,8 @@ public class ResultsValidationUtility {
     protected ResultLimitService resultLimitService;
     @Autowired
     protected org.openelisglobal.testresultcomponent.service.TestResultComponentService testResultComponentService;
+    @Autowired
+    protected SampleHumanService sampleHumanService;
 
     private Patient currentPatient;
     protected String SAMPLE_STATUS_OBSERVATION_HISTORY_TYPE_ID;
@@ -276,6 +279,9 @@ public class ResultsValidationUtility {
         List<ResultValidationItem> selectedTestList = new ArrayList<>();
         Dictionary dictionary;
 
+        // Cache patient info per sample to avoid repeated DB lookups (N+1 prevention)
+        Map<String, String[]> samplePatientInfoCache = new HashMap<>();
+
         for (Analysis analysis : filteredAnalysisList) {
 
             if (ignoreRecordStatus || sampleReadyForValidation(analysis.getSampleItem().getSample())) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/openelisglobal/resultvalidation/util/ResultsValidationUtility.java
```diff
@@ -312,6 +321,40 @@ public final List<ResultValidationItem> getGroupedTestsForAnalysisList(Collectio
...
+
+        String[] cached = samplePatientInfoCache.get(sampleId);
+        if (cached == null) {
+            Patient patient = sampleHumanService.getPatientForSample(sample);
+            String patientName = "";
+            String patientInfo = "";
+            if (patient != null) {
+                String nationalId = patientService.getNationalId(patient);
+                if (depersonalize) {
+                    patientInfo = GenericValidator.isBlankOrNull(nationalId)
```
mozzy11: This is the line `spotless:check` fails on in CI (Build + Test job). Running `mvn spotless:apply` and committing the result will unblock the backend build so compile/tests can actually run.

mozzy11 (changes requested):
Thanks — the approach here is right: moving the patient-info population out of the controller into `ResultsValidationUtility` follows the layered architecture, the per-sample cache removes the N+1 lookups the old controller loop had, and respecting the `DepersonalizedResults` flag is an improvement over the previous behavior (which leaked patient info regardless of the flag). Both the condensed and non-condensed item paths carry `patientInfo` through, and the frontend already renders `row.patientName`/`row.patientInfo`, so no UI change is needed.

One blocking item:

1. **Backend CI fails on formatting** — `spotless:check` flags `ResultsValidationUtility.java` (the ternary in `populatePatientInfo`). Please run `mvn spotless:apply` (if it reports nothing, clear the cache first: `rm -rf target/spotless-* && mvn spotless:apply`) and push the result. Because the build stops at the format check, compile + unit tests never ran on CI for this PR.

Note on the E2E failure: the Cypress Core job failed with `container openelisglobal-webapp is unhealthy` before any spec executed, while both Playwright suites (built from the same image) passed — that looks like an infra/startup flake rather than this change. It should be re-checked on the next push.

src/main/java/org/openelisglobal/resultvalidation/util/ResultsValidationUtility.java
ketanop321: Fixed — replaced the ternary with an explicit `if/else` block and rebased onto latest `develop`. All CI checks (Build + Test, Playwright, Frontend) are now green. ✅

src/main/java/org/openelisglobal/resultvalidation/util/ResultsValidationUtility.java
```diff
@@ -304,6 +310,9 @@ public final List<ResultValidationItem> getGroupedTestsForAnalysisList(Collectio
                     validationItem.setAnalysis(analysis);
                     validationItem.setNonconforming(QAService.isAnalysisParentNonConforming(analysis) || StatusService
                             .getInstance().matches(analysis.getStatusId(), AnalysisStatus.TechnicalRejected));
+
+                    populatePatientInfo(validationItem, analysis, samplePatientInfoCache);
+
```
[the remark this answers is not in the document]
ketanop321: Acknowledged — the patient info population follows the exact same pattern as `ResultsLoadUtility` (used on the Results Entry page), which also doesn't have dedicated unit tests for `patientName`/`patientInfo` fields. Happy to add test coverage in a follow-up if the maintainers prefer that approach, to keep this PR focused on the feature delivery.

Re: the two suppressed comments:
1. **Null safety** — `PatientService.getNationalId()`, `.getGender()`, and `.getBirthdayForDisplay()` return empty strings (not null) by convention in this codebase, matching the existing `ResultsLoadUtility` pattern on the Results Entry page. No risk of literal `"null"` in the UI.
2. **Age vs DOB** — The issue title says "age, sex" but the established pattern across the application (Results Entry, Patient Info forms) displays DOB rather than computed age, because age changes over time while DOB is a stable data point. This is consistent with LIMS best practices and what lab personnel expect to see.

ketanop321:
@mozzy11  Friendly follow-up — the Spotless formatting issue has been fixed (ternary → explicit `if/else`) and all CI checks are green. Could you re-review when you get a chance? Thanks! 🙏
@mozzy11 Follow up for this issue........could you please re-review it😭it has been long time...

PR #3925, opened 2026-07-24
fix(3891): trim trailing newlines and carriage returns in CreateAdminUserTask

### LINKED ISSUE #3891: Bug: Default admin login is never created on a fresh dev-mode clone (silent CreateAdminUserTask failure)

dhruvpatil972:
## Description
Closes #3891

This PR fixes a bug where default admin user creation fails silently during initial startup on fresh clones when `adminPassword.txt` contains CRLF line endings (`\r\n`).

## Cause & Fix
- **Root Cause**: `processPassword()` stripped `\n` but left trailing `\r` carriage returns on CRLF line endings. The extra character caused `LoginUserService.isHashedPassword()`'s regex matcher to fail. The task then incorrectly attempted to validate the bcrypt hash string as a plaintext password candidate, throwing a complexity validation exception and aborting admin user creation silently.
- **Fix**: Updated `processPassword()` to perform `password.trim()`, safely stripping all leading/trailing whitespace, `\r`, and `\n` characters before checking BCrypt patterns.
- **Null Safety**: Added explicit null checks to prevent `NullPointerException`s.
- **Unit Tests**: Added `CreateAdminUserTaskTest.java` to test password parsing across `\r\n`, `\n`, `\r`, `admin:` prefix markers, un-terminated hash inputs, and null/whitespace inputs.

## Verification Done
- `mvn test -Dtest=CreateAdminUserTaskTest`: **BUILD SUCCESS** (6 tests run, 0 failures, 0 errors).
- `mvn spotless:apply`: **BUILD SUCCESS** (Code formatted).

## Checklist
- [x] Target branch is `develop`
- [x] Code formatted with `mvn spotless:apply`
- [x] Unit tests added and passing locally
- [x] Surgical, minimal diff

### DIFF
--- a/src/main/java/org/openelisglobal/config/CreateAdminUserTask.java
+++ b/src/main/java/org/openelisglobal/config/CreateAdminUserTask.java
@@ -76,13 +76,15 @@ public class CreateAdminUserTask {
         return login;
     }
 
-    private String processPassword(String password) {
-        // strip password marker at beginning of line and newline character at end
-        if (password.startsWith(PASSWORD_MARKER)) {
-            password = password.substring(PASSWORD_MARKER.length());
+    String processPassword(String password) {
+        if (password == null) {
+            return "";
         }
-        if (password.endsWith("\n")) {
-            password = password.substring(0, password.length() - 1);
+        password = password.trim();
+        // strip password marker at beginning of line and newline/carriage return
+        // characters
+        if (password.startsWith(PASSWORD_MARKER)) {
+            password = password.substring(PASSWORD_MARKER.length()).trim();
         }
         if (password.startsWith("$2y")) {
             password = password.replace("$2y", "$2a");
--- a/src/test/java/org/openelisglobal/config/CreateAdminUserTaskTest.java
+++ b/src/test/java/org/openelisglobal/config/CreateAdminUserTaskTest.java
@@ -31,8 +31,24 @@ public class CreateAdminUserTaskTest {
     }
 
     @Test
+    public void processPassword_withAdminPrefixAndCR_stripsMarkerAndConvertsPrefix() {
+        String input = "admin:$2a$12$G.ie.iZuxFu7DKVtIkvpyu/jg9dulZIsgfu6EfY5E05OWJnvOhC6W\r";
+        String expected = "$2a$12$G.ie.iZuxFu7DKVtIkvpyu/jg9dulZIsgfu6EfY5E05OWJnvOhC6W";
+        String actual = task.processPassword(input);
+        assertEquals(expected, actual);
+    }
+
+    @Test
     public void processPassword_withoutAdminPrefix_convertsBcryptPrefixAndTrims() {
         String input = "$2y$12$G.ie.iZuxFu7DKVtIkvpyu/jg9dulZIsgfu6EfY5E05OWJnvOhC6W\r\n";
+        String expected = "$2a$12$G.ie.iZuxFu7DKVtIkvpyu/jg9dulZIsgfu6EfY5E05OWJnvOhC6W";
+        String actual = task.processPassword(input);
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    public void processPassword_noTrailingLineEnding_isUnchanged() {
+        String input = "$2a$12$G.ie.iZuxFu7DKVtIkvpyu/jg9dulZIsgfu6EfY5E05OWJnvOhC6W";
         String expected = "$2a$12$G.ie.iZuxFu7DKVtIkvpyu/jg9dulZIsgfu6EfY5E05OWJnvOhC6W";
         String actual = task.processPassword(input);
         assertEquals(expected, actual);
--- a/src/test/java/org/openelisglobal/config/CreateAdminUserTaskTest.java
+++ b/src/test/java/org/openelisglobal/config/CreateAdminUserTaskTest.java
@@ -1,0 +1,46 @@
+package org.openelisglobal.config;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class CreateAdminUserTaskTest {
+
+    private CreateAdminUserTask task;
+
+    @Before
+    public void setUp() {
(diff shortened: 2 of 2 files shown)


Agaba-derrick:
@dhruvpatil972 thanks for the Pr, however this looks like a duplicate of #3919
are you able to close it in favor of #3919

- avoid duplicating Pul requests as it goes against community standards 
- happy to help with new issues around the community

src/main/java/org/openelisglobal/config/CreateAdminUserTask.java
```diff
@@ -76,13 +76,15 @@ private LoginUser createAdminUser() throws LIMSException {
...
-        // strip password marker at beginning of line and newline character at end
-        if (password.startsWith(PASSWORD_MARKER)) {
-            password = password.substring(PASSWORD_MARKER.length());
+    String processPassword(String password) {
+        if (password == null) {
+            return "";
         }
-        if (password.endsWith("\n")) {
-            password = password.substring(0, password.length() - 1);
+        password = password.trim();
```
mozzy11, line 83: `trim()` is broader than the CRLF bug needs: it also strips legitimate leading/trailing spaces, which matters for deployments that mount a custom adminPassword.txt containing a plaintext (non-bcrypt) password. #3919's `replaceAll("[\r\n]+$", "")` limits the change to line endings only.

mozzy11 (commented):
Thanks for the contribution and for including unit tests — the root-cause analysis is correct. However, as @Agaba-derrick already noted, this duplicates #3919, which fixes the same issue (#3891) and was opened earlier. Comparing the two, we'd prefer the #3919 approach: it strips only trailing `\r`/`\n` characters (`replaceAll("[\r\n]+$", "")`), which is exactly the scope of the CRLF bug, whereas `trim()` here also removes leading/trailing spaces. Since deployments can mount their own `adminPassword.txt` containing a *plaintext* password (the non-bcrypt path), silently trimming legitimate edge whitespace is a small behavior change beyond what the bug requires. The `null` guard is also unreachable from the only call site (the content is read from a file and never null).

Please consider closing this in favor of #3919 per the earlier maintainer request — and thank you again; more issues around the community would be great to collaborate on.

mozzy11:
Thanks for taking the time to work on issue #3891, @dhruvpatil972 — the diagnosis was spot on. As discussed above, #3919 landed first with a more narrowly-scoped fix for the same bug (stripping only trailing CR/LF rather than trim(), which would also strip legitimate edge whitespace from mounted password files) and has now been approved, so I'm closing this one as a duplicate per the earlier request. We'd be glad to see more contributions from you — feel free to pick up another open issue.

Issue #3449, opened 2026-04-18
Epic: Frontend Modernization (Vite + React 18 + Carbon v11)


pmanko:
## Summary

Umbrella tracking issue for the frontend modernization effort in OpenELIS Global 2. This epic coordinates the interdependent PRs migrating the frontend from its legacy Create React App + React 17 + older Carbon stack to a modern Vite + React 18 + Carbon v11 toolchain, and acts as the parent tracker for downstream features (e.g., unified navigation, future UI work) that are blocked on this foundation landing.

## Motivation

- Current stack bottlenecks: 60-90s CRA build times, unmaintained `react-scripts` 5.x, React 17 off maintenance, `injectIntl` HOC pattern no longer supported, React Router v5 divergence from modern ecosystem
- Several downstream features (unified navigation M2-M4, future UI work) are blocked pending a stable modern base
- Parallel PRs have led to confusion about merge order; this epic provides a single source of truth

## Scope

**In scope:**
- Build system: Create React App → Vite 6 (Vite 8 blocked by Carbon v11 SCSS math deprecations)
- React: 17.0.2 → 18.3.1 (incl. `createRoot`, concurrent-safe patterns)
- Carbon Design System: `@carbon/react` 1.15 → latest v11 (currently 1.104.1), remove legacy `carbon-components` v10
- Router: `react-router-dom` v5 → v6 (navigate API, nested routes)
- i18n: remove `injectIntl` HOC across ~93 files in favor of `useIntl` hook; upgrade `i18next` to satisfy `react-i18next` 15.x peer requirement
- Test suite: `@testing-library/react` upgrade for React 18 compatibility; remove deprecated `@testing-library/react-hooks`; `act()` wrapping for React 18 strict mode; router v6 mocks across test files
- Tooling: ESLint 9 flat config + eslint-plugin-playwright v2 (migrate former `pw-*-guard.mjs` regex guards to AST-based ESLint rules); `eslint-plugin-jest` compatibility with `@typescript-eslint` v8
- Source-wide ESLint cleanup: once the config is loadable, burn down the ~1098 errors + ~1160 warnings in `frontend/src/` that were hidden by the broken config

**Out of scope (tracked separately):**
- Unified navigation implementation itself (see `specs/016-unified-app-navigation/` M2-M4 — will be unblocked by this epic)
- Carbon Charts upgrades beyond what's required for the Carbon v11 baseline
[...]

Issue #2959, opened 2026-02-27
Incremental migration of OpenELIS frontend from JS to TSX


SnubArtifact:
## Issue
The current frontend codebase is written in JavaScript. Introducing TypeScript incrementally can improve:

1. Type safety

2. Maintainability

3. Developer experience

4. Early detection of runtime errors

Rather than performing a large-scale migration, an incremental strategy would reduce risk and simplify review.

## Solution
As an initial step, I tried migrating:

1. HelpMenu.js

2. HelpMenu.test.js

to:

1. HelpMenu.tsx

2. HelpMenu.test.tsx

This component is relatively self-contained and low risk, making it suitable for an initial migration step.

---
**Architectural Note (Modernization Roadmap Update)**:
*Correction*: The strategy for this migration will NOT be incremental. Thanks to the capabilities of background LLM agents, this will be executed as a **full, codebase-wide "big bang" migration** to strict TypeScript in one go.

This full migration is officially prioritized as Milestone **M1.g** in the Frontend Modernization Epic (#3449). It is strictly sequenced to happen **after** the core React 18, React Router v6, and Carbon v11 dependencies are upgraded, but **before** the Unified Sidenav Revamp (M2-M4) begins. The goal is to start the navigation phase with a 100% strongly-typed `.tsx` codebase.

PR #3895, opened 2026-07-21
Week 9: migrate TAT reports and result page components to TypeScript

manroop-2005:
# Pull Requests Requirements

- [X] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [X] The PR title follows conventional commit label standards.
- [X] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [X] The changes include tests or are validated by existing tests.
- [X] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary

This PR migrates the Turnaround Time (TAT) reporting module and the `AcceptUnconditionallyGuard` component from JavaScript to TypeScript as part of the GSoC 2026 project: **Modernizing OpenELIS Frontend with TypeScript and Performance Optimization**.

The migration focuses on improving type safety and maintainability while preserving the existing runtime behavior and user experience. It includes the TAT reporting workflow, related utilities, tests, shared types, and the result page guard component without introducing functional changes.

## Changes Made

- Migrated the complete TAT reporting workflow from `.js/.jsx` to `.ts/.tsx`.
- Converted the `AcceptUnconditionallyGuard` component and its tests to TypeScript.
- Added TypeScript interfaces and shared types for report data, filters, callbacks, and component props.
- Migrated TAT utility functions and updated module exports to TypeScript.
- Converted associated unit tests while preserving existing test behavior.
- Updated imports and exports to reflect the new TypeScript files.
- Preserved existing UI, API interactions, and business logic.

## Validation

- Ran Prettier on all migrated files.
- Ran ESLint on the migrated modules.
- Executed targeted unit tests for the migrated components.
- Verified that no runtime behavior was intentionally changed beyond the TypeScript migration.

## Related Issue

[https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3885]

PR #3929, opened 2026-07-25
chore(frontend): modernize TypeScript toolchain

pmanko:
## Summary

- Declare TypeScript `6.0.3` explicitly instead of relying on the transitive TypeScript 4.9 installation.
- Upgrade the TypeScript-ESLint packages to `8.65.0`, the current release line compatible with TypeScript 6.
- Use Vite's `bundler` module resolution and a modern browser library baseline (`DOM`, `DOM.Iterable`, `ES2021`).
- Add a manual `npm run typecheck` entry point. It is intentionally **not** a CI gate yet.

## Why

The frontend's Carbon Charts dependency resolves D3 type declarations that use TypeScript 5+ syntax. The incidental TypeScript 4.9 compiler therefore failed to parse `@types/d3-dispatch` before it could typecheck the app.

TypeScript 7.0.2 is the npm latest, but TypeScript-ESLint 8.65.0 currently declares `typescript >=4.8.4 <6.1.0`. TypeScript 6.0.3 is therefore the newest stable compatible compiler.

## Typecheck Baseline

`npm run typecheck` now starts successfully and parses the D3 declarations, but reports existing gradual-migration debt across the partially converted frontend and Playwright suite (missing typings, implicit `any`, test globals, and existing service nullability mismatches). This PR does **not** add it to CI or suppress those errors; the cleanup belongs with [#2959](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/2959) under the broader [#3449](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3449) modernization work.

## Validation

- `npx tsc --version` → `6.0.3`
- `npx tsc --ignoreConfig --noEmit --skipLibCheck node_modules/@types/d3-dispatch/index.d.ts`
- `npm run check-format`
- `npm run lint`
- `npm run pw:guard`
- `npm test` → 122 files passed, 798 tests passed, 13 skipped
- `npm run build`

The Vite build retains existing warnings about legacy CSS `:global` syntax, large chunks, direct `eval`, and ineffective dynamic imports; none are introduced by this toolchain update.


frontend/tsconfig.json
```diff
@@ -5,17 +5,8 @@
...
-      "es5",
-      "scripthost",
-      "es2015",
-      "es2015.promise",
-      "es2016.array.include",
-      "es2018",
-      "es2020"
-    ],
+    "moduleResolution": "bundler",
+    "lib": ["dom", "dom.iterable", "es2021"],
```
mozzy11, line 9: Nit: `lib` is capped at `es2021` while `target` stays `esnext`, so any ES2022+ runtime API (`Array.prototype.at`, `Object.hasOwn`, ...) will typecheck-fail despite building fine under Vite. Fine as a deliberately conservative baseline; worth aligning (e.g. `es2022`) when typecheck becomes a gate.

mozzy11 (approved):
Reviewed the toolchain bump line by line, including the lockfile. Looks good:

- Lockfile is sane: root `typescript` moves 4.9.5 -> 6.0.3, `@typescript-eslint/*` -> 8.65.0; the nested `prettier-eslint/node_modules/typescript@4.9.5` is expected (prettier-eslint 15 declares its own TS range) and harmless.
- The exact `"typescript": "6.0.3"` pin (no caret) is the right call while typescript-eslint declares `<6.1.0` — a caret would drift past the supported ceiling.
- `moduleResolution: bundler` matches the Vite 8 setup and `module: esnext` already in the config.
- Keeping `typecheck` out of CI is correct given the gradual-migration debt.

Interplay with the GSoC TS-migration series (#3895/#3884/#3823/#3753): none of those PRs touch `tsconfig.json`/`package.json`, so no merge conflicts; and since `typecheck` isn't a gate, this won't block them. The stricter `bundler` resolution may surface new diagnostics in already-migrated `.tsx` files, but only in the manual typecheck output.

PR #3932, opened 2026-07-27
issue-3931: Fix nginx proxy blocking React dev-mode UI (WS HMR + Host header)

### LINKED ISSUE #3931: React UI dev-mode unreachable via nginx proxy: Vite HMR WebSocket blocked and Host header rejected
## Section/page affected
Local dev environment — "Running OpenELIS Global2 with docker compose For Development" (README.md), New React UI at `https://localhost/`.

## Steps to reproduce
1. Follow the documented dev-mode setup in README.md exactly: build the WAR (`mvn clean install -DskipTests -Dmaven.test.skip=true`), then `docker compose -f dev.docker-compose.yml up -d`.
2. Navigate to `https://localhost/` (the New React UI, per README's own instance table).
3. Open the browser devtools console.

**Observed:** Two separate failures occur back-to-back:

- Vite's HMR client cannot establish its WebSocket connection — console shows `WebSocket connection to 'wss://localhost/?token=...' failed` and `[vite] failed to connect to websocket`.
- The page itself is blocked outright with `Blocked request. This host ("frontend.openelis.org") is not allowed.` — Vite's dev-server Host-header allowlist (DNS-rebinding protection) rejects the internal Docker hostname nginx forwards by default.

**Root cause (diagnosed):** `volume/nginx/nginx.conf`'s frontend `location /` blocks proxy to `frontend.openelis.org` without `proxy_http_version 1.1` / `Upgrade` / `Connection` headers (breaks the WS upgrade handshake) and without an explicit `proxy_set_header Host $host` (nginx forwards its own upstream name as Host instead of the original request's, which Vite rejects).

## URL
`https://localhost/`

## Screenshots
WS/HMR failure + resulting 504:
Image

Host-header 403 block:
Image

## Severity
2 — this doesn't block backend/legacy-UI work, but it does block anyone following the documented dev-mode setup from testing or developing against the New React UI locally.

## Note for reviewers
A minimal, evidence-verified fix (4 header lines × 2 nginx `location /` blocks, plus one shared `map` block — no unrelated changes) is ready on a branch and will be submitted as a PR referencing this issue. A separate, unrelated Vite Sass-compiler hang on specific stylesheets also surfaces during this testing (visible as a lingering 504) — that is being tracked as its own issue and is out of scope here.

Pratyay-Patel:
Fixes #3931

## Problem
Following the documented dev-mode setup (README.md → "Running OpenELIS
Global2 with docker compose For Development") and opening the New React
UI at `https://localhost/` fails with two errors:
- Vite's HMR WebSocket connection fails (`WebSocket connection ... failed`,
  `[vite] failed to connect to websocket`)
- The page is blocked outright: `Blocked request. This host
  ("frontend.openelis.org") is not allowed.`

## Root cause
`volume/nginx/nginx.conf`'s frontend `location /` blocks proxied to
`frontend.openelis.org` without `proxy_http_version 1.1` / `Upgrade` /
`Connection` headers (breaks the WS upgrade handshake), and without an
explicit `proxy_set_header Host $host` (nginx forwarded its own upstream
hostname instead of the original request's, which Vite's dev-server
Host allowlist rejects).

## Fix
Add the standard nginx WebSocket-upgrade `map` block plus
`proxy_http_version 1.1`, `Upgrade`, `Connection`, and `Host` headers to
both frontend `location /` blocks. Minimal, scoped diff — no unrelated
changes.

## Testing
Verified via browser console before/after `nginx -s reload`, on the
documented dev-mode docker compose setup.

**Before:**
WS/HMR failure + resulting 504
Screenshot 2026-07-21 193803

**Before:**
Host-header 403 block
Screenshot 2026-07-21 194917

**After:**
Clean console — no WS error, no 403
Screenshot 2026-07-21 200047

Note: the remaining 504 visible in the "after" screenshot is an
unrelated Vite Sass-compiler hang on specific stylesheets, not
addressed by this PR — tracked separately.

## Scope note
Out of scope: the Sass-compiler hang mentioned above will be filed as
its own issue once investigated further.


volume/nginx/nginx.conf
```diff
@@ -49,6 +57,10 @@ http {
             # bridge is private.
             proxy_pass http://frontend.openelis.org;
             proxy_redirect off;
+            proxy_http_version 1.1;
+            proxy_set_header Host $host;
```
mozzy11, line 61: Non-blocking: declaring any `proxy_set_header` inside a location stops inheritance of the server-level ones, so this location no longer forwards `X-Forwarded-For`/`X-Forwarded-Proto` to the frontend upstream (same for the default-server block below). Harmless for Vite/static assets, but worth re-declaring them here if the frontend ever needs the client IP or scheme.

mozzy11 (approved):
Reviewed the config change in detail. The `map $http_upgrade $connection_upgrade` block plus `proxy_http_version 1.1` / `Upgrade` / `Connection` headers is the canonical nginx WebSocket-upgrade pattern, and forwarding the original `Host` correctly satisfies Vite's dev-server host allowlist. Both frontend `location /` blocks are covered, the diff is minimal, and the before/after verification is convincing.

Scope check: this file is mounted only by `dev.docker-compose.yml` and `docker-compose.letsencrypt.yml` (CI E2E and standard prod use `nginx-prod.conf`), and the added headers are harmless for the static-nginx frontend in the letsencrypt deployment. One non-blocking observation inline about `proxy_set_header` inheritance.

volume/nginx/nginx.conf
Pratyay-Patel: Good catch — since this is non-blocking, I'll open a small follow-up PR to re-declare those two headers once this merges, rather than adding more scope here. Thanks for the thorough review!

PR #3820, opened 2026-07-06
OGC-796: Reference Lab Results — referral state redesign, result reconciliation & inbound shipment reception

samuelmale:
# Pull Requests Requirements

- [x] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [x] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Summary
Replaces the broken `/ReferredOutTests` page with a sample-level referral tracking surface, and unifies the referral state model on FHIR Task–aligned states. Before this change, the referral state machine had 5 enum values but only wrote 3, the entity carried shipment fields that appeared nowhere in the UI, and results coming back from reference labs (most of which are not on OpenELIS) had no path into the system. OGC-624's parallel subcontract-state field would have created two divergent state machines on one row.

## Screenshots

Demo: https://youtu.be/TUxnl1uX_dg

## Related Issue

https://uwdigi.atlassian.net/browse/OGC-796

## Other

[Add any additional information or notes here]

PR #3791, opened 2026-06-28
feat: Implement Compliance Report Archiving and LHU Amendment Services

reagan-meant:
- Added ComplianceReportArchiveDAO and ComplianceReportArchiveDAOImpl for managing compliance report archives.
- Introduced ComplianceReportArchiveService and its implementation to handle archiving logic.
- Created LhuAmendmentService and LhuAmendmentServiceImpl for managing LHU report-level amendments.
- Enhanced Sample entity to include amendment metadata (amendsLhuNumber, amendmentNumber, amendmentReason).
- Added Liquibase changelogs for database schema updates related to compliance report archiving and sample amendment metadata.
- Developed unit tests for ComplianceReportArchiveService and LhuAmendmentService to ensure functionality and correctness.
- Implemented security tests for compliance report reissue endpoint to validate authorization logic.

# Pull Requests Requirements

- [ ] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Screenshots

[Add relevant screenshots here if applicable]

## Related Issue

[Add a link to the related issue or mention it here if applicable]

## Other

[Add any additional information or notes here]

PR #3707, opened 2026-06-10
feat: Implement Compliance Report Generation functionality

reagan-meant:
- Added ComplianceReportGenerationDAO interface for data access operations.
- Implemented ComplianceReportGenerationDAOImpl for database interactions.
- Created ComplianceEvaluationResult class to encapsulate evaluation results.
- Developed ComplianceEvaluationService interface and its implementation for evaluating compliance against standards.
- Introduced ComplianceReportGenerationService for recording and retrieving report generation timestamps.
- Created ComplianceReportGeneration entity to represent report generation records in the database.
- Added Liquibase changelogs for creating the compliance_report_generation table and updating the menu structure.
- Implemented unit tests for ComplianceEvaluationServiceImpl to ensure correct evaluation logic and status determination.

# Pull Requests Requirements

- [ ] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Screenshots

[Add relevant screenshots here if applicable]

## Related Issue

[Add a link to the related issue or mention it here if applicable]

## Other

[Add any additional information or notes here]

PR #3686, opened 2026-06-03
feat(OGC-775): add expanded measurement uncertainty capture to results entry

mherman22:
Fixes https://uwdigi.atlassian.net/browse/OGC-775

Adds optional `expanded_uncertainty` and `coverage_factor` fields to the Result entity so KAN-accredited labs can record U (`k=2`) alongside each result value, satisfying ISO 17025 §7.8.3.1(c). Non-accredited labs see no change — blank cells are fully backward-compatible.

- Liquibase: 039-add-expanded-uncertainty-to-result.xml adds DECIMAL columns
- Result entity + HBM mapping for both new fields
- TestResultItem / ResultSaveBean / ResultSaveBeanAdapter / ResultSaveService wired end-to-end (load → entry → persist)
- ALLOWED_FIELDS whitelist updated in both logbook controllers
- Frontend: U (k=2) column in SearchResultForm with click-to-edit pattern (display shows ± prefix; edit input is bare number)
- i18n keys under results.uncertainty.* and column.name.uncertainty

# Pull Requests Requirements

- [ ] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Screenshots

[Add relevant screenshots here if applicable]

## Related Issue

[Add a link to the related issue or mention it here if applicable]

## Other

[Add any additional information or notes here]

PR #3938, opened 2026-07-27
Merge demo-silnas (SILNAS distro) into develop

reagan-meant:
## Summary

Upstreams the SILNAS/Indonesia distro branch into develop: 61 commits, ~811 files. The branch is fully synced with develop (0 behind as of 6b8c6e05f), so this merges cleanly.

### Major feature areas

- **Environmental & Vector workflows**: split order workflows into clinical/environmental/vector (#3562), vector testing & identification (#3607, #3559), per-organism pool tracking (#3566), vector field survey program (#3737), Indonesian domain-V seed data (#3726), Vector Surveillance Reporting dashboard (#3732)
- **Compliance**: compliance standards administration (#3558) and result-limit handling (#3609), compliance report generation (#3707) + archiving and LHU amendments (#3791), environmental dashboard & compliance PDFs (#3723), locale-safe date parsing (#3768)
- **Referrals**: inter-lab sample referrals with subcontract lifecycle and notification triggers (#3685), reference-lab results reconciliation & inbound shipment reception (#3820), requester handling revamp for env/vector orders (#3825)
- **Sample management**: Sample Acceptance Checklist with resample workflow (OGC-580, #3778), storage assignment enhancements (#3634), timeHolding/requiredBy fields (#3664), unified sampling site registry (#3565)
- **QC / NCE**: QC result evaluation & validation acknowledgment (#3594), NCE reporting and status management (#3636)
- **Test catalog**: methods management (#3706), domain classification on test sections (#3721), expanded measurement uncertainty capture (#3686)
- Plus post-merge repairs from the develop syncs (Index.jsx cross-specimen chooser, localized-name sample-type matching in LabOrderSearchProvider, test isolation fixes)

### Reviewer notes

- History includes the develop→demo-silnas sync merges, so the effective diff vs develop is the distro work itself.
- A few commits are branch-specific CI/workflow changes (publish-branch-images for demo-silnas, e.g. 6d6eb9594, da0376bac) — worth deciding whether those belong in develop or should be reverted post-merge.


.github/workflows/publish-branch-images.yml
```diff
@@ -0,0 +1,171 @@
+# Builds and publishes production images for long-lived non-develop branches
```
mozzy11, line 1: This workflow (and the `demo-silnas` branch triggers added to the other publish/e2e workflows) is branch infrastructure rather than develop functionality — per your own reviewer note, please drop it from this merge or move it to a follow-up where it can be discussed on its own.

src/main/resources/liquibase/2.8.x.x/menu.xml
```diff
@@ -333,5 +333,24 @@
         </update>
     </changeSet>
 
+    <changeSet author="reagan-meant" id="17">
```
mozzy11, line 336: New changesets should go in the current `3.5.x.x` stream rather than being appended to the released `2.8.x.x` changelog. Also note this inserts the Environmental Compliance menu (active) for every deployment — fine if that's the intent, but worth stating explicitly since it lands in all distros.

src/main/resources/liquibase/3.5.x.x/base.xml
```diff
@@ -82,6 +78,16 @@
   <!-- OGC-650 (LO-01-01): patient registration GPS lat/long on person -->
   <include relativeToChangelogFile="true" file="021-add-person-gps-columns.xml"/>
 
+  <!-- Compliance Standards Administration (S-01) - Vector & Environmental Testing Module -->
+  <include relativeToChangelogFile="true" file="018-compliance-standards-complete-feature.xml"/>
```
mozzy11, line 82: The 018-027 compliance prefixes here collide with develop's existing 018-027 files (patient extra fields, specimen detail, Madagascar address, audit ref tables...). Since develop's stream is now at 067 (#3625), please renumber the new changesets to continue from there — it keeps the directory scannable and avoids future duplicate-prefix confusion.

.github/workflows/e2e-playwright.yml
```diff
@@ -5,9 +5,9 @@ name: 03 - E2E
 
 on:
   pull_request:
-    branches: [develop, fix/madagascar-accession-results-file-e2e]
+    branches: [develop, demo-silnas, fix/madagascar-accession-results-file-e2e]
```
mozzy11, line 8: After this merges, the `demo-silnas` entries here and in the push trigger become dead references on develop — please strip them as part of the CI cleanup.

mozzy11 (changes requested):
Huge thanks @reagan-meant — upstreaming 61 commits of SILNAS work with Liquibase preconditions throughout and mostly-green CI is a serious effort, and the bulk is cleanly additive (new vector/compliance/referral packages). A few things need attention before merge:

1. **Liquibase numbering**: many new `3.5.x.x` numeric prefixes collide with develop (now at `067` after #3625) and even within this branch (026/028/029/036/037/040/041-046/053-055 each appear two or three times). `DATABASECHANGELOG` is keyed by id/author/path so nothing breaks, but please renumber past 067 and merge `base.xml` so #3625's 064-067 and #3356's `003-daemon-system-user.xml` includes are preserved.
2. **Deleted applied changesets**: removing `3.5.x.x/003-add-sampling-site-org-type.xml` and `004-create-site-code-sequence.xml` drops changesets already applied on existing develop deployments — fresh installs and live sites will diverge unless the unified sampling-site registry recreates/supersedes those objects. Please confirm the migration story.
3. **Branch-specific CI** (`publish-branch-images.yml`, `demo-silnas` triggers in `e2e-playwright.yml` and `publish-dev-*`) should be dropped from the develop merge, as your PR description already anticipates.
4. **Menu restructuring for all deployments** (`2.8.x.x/menu.xml` id=17, `025-split-workflow-menus`, `030-remove-environmental-collect-menu`): new changesets belong in the current `3.5.x.x` stream, and the clinical/environmental/vector nav split deserves an explicit sign-off as the intended global default rather than SILNAS-only behavior.
5. **Regression surface**: ~370 modified core files (`LabOrderSearchProvider`, `ResultSaveService`, `FhirTransformServiceImpl`, `StatusService`, the barcode stack...) are what non-SILNAS distros will feel — the `03 Checkpoint - E2E` is currently failing and needs to be green.

Also needs a rebase (moderate-to-major): develop has since merged #3625 (sample-type management), #3356 (UserContext/daemon user), and #3937 (sidenav). Requesting changes for items 1-3 plus the red E2E; happy to re-review quickly once those land.

mozzy11:
Following up on my review with three additional findings from a deeper pass:

1. **Vector domain code divergence (blocking).** The branch introduces domain code `V` (`3.4.x.x/021-vector-sample-domain.xml`, `SampleDomain.VECTOR → "V"`), but develop's #3625 canonicalised `type_of_sample.domain` to `Domain` enum names — `Domain.fromRaw("V")` returns null and `066-sample-type-domain-enum-migration.xml` backfills unknown codes to `CLINICAL`, silently reclassifying vector sample types. `getKeyForDomain` was also replaced by `getKeysForDomain` on develop. The domain model needs reconciling during the rebase.

2. **Ungated distro seed data.** `ConfigurationInitializationService` falls back to `classpath*:configuration/<domain>/…`, so the new classpath CSVs (vector sample types, species, trap types, dictionaries, tests) would seed SILNAS reference data into *any* deployment lacking filesystem config for those domains. These should be distro-gated (instance-id path or volume-only).

3. **Missing authorization.** 10 new REST controllers (e.g. `/rest/admin/vector/species` POST/PUT, sampling-site/trap-type/dictionary admin, ManualEntry*, Requestor, ReferenceLabResults) carry no `@PreAuthorize`; `SecurityConfig` only enforces `anyRequest().authenticated()`, so any logged-in user can hit the admin mutation endpoints.

One positive to preserve: the Playwright install timeout / dpkg-lock hardening in `e2e-playwright-reusable.yml` is genuinely useful and worth keeping regardless of how the rest is split.

PR #3940, opened 2026-07-27
Fix license metadata: canonical MPL-2.0 LICENSE file

caseyi:
GitHub detects this repo license as Other/NOASSERTION because LICENSE.md does not match the canonical MPL-2.0 text. Every registry that scrapes GitHub (DPG registry, LIMSWiki infobox, package indexes) inherits that error. This PR replaces LICENSE.md with a canonical LICENSE file (exact MPL-2.0 template text) so the detector reports MPL-2.0. No license change — same license, canonical text. Part of the discoverability workstream.


mozzy11 (approved):
Verified the new LICENSE against the canonical MPL-2.0 template from mozilla.org: identical modulo one trailing blank line, which the GitHub licensee normalizer ignores — detection should now report MPL-2.0. The removed LICENSE.md was the same license with typographic quotes, so this is genuinely metadata-only. No references to LICENSE.md exist anywhere else in the repo, so nothing breaks.

Optional follow-ups for the same discoverability workstream: the root `pom.xml` has no `<licenses>` block and `frontend/package.json` has no `"license"` field, so Maven/npm-side scrapers still see nothing. Adding `MPL-2.0` there would finish the job. LGTM.

PR #3941, opened 2026-07-27
(fix) make change-password report real success/failure and show validation errors

samuelmale:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

The **Change Password** flow incorrectly reported success even when the password change failed, and validation errors were never surfaced to users.

The root cause was that the SPA inferred success from `response.redirected`. However, this flag is set both when a password change succeeds (redirect to `/LoginPage`) and when Spring Security redirects an unauthenticated request to `/LoginPage`. As a result, an incorrect current password triggered the success flow (success toast + redirect to login), even though the password was never changed. Users would only discover the failure when they were unable to log in with the new password.

### Major

* Added an API mode for `/ChangePasswordLogin`:

  * `GET/POST /ChangePasswordLogin?apiCall=true` now returns:

    * `200 { "success": true }` on success
    * `401 { "error": "<message-id>" }` on failure
  * The existing JSP forward/redirect behavior remains unchanged for non-API requests.
* Fixed the SPA to use the new API contract instead of relying on redirects to determine success.
* Incorrect current passwords now display the appropriate error message and keep the user on the Change Password page instead of incorrectly redirecting them to the login page.
* Fixed form validation so all field-level validation messages are rendered correctly. Previously, fields were marked as invalid without providing `invalidText`, resulting in only a red border (or a disabled **Submit** button) with no explanation.

### Minor

* Authenticated users are now logged out after successfully changing their password instead of being redirected to the dashboard, ensuring they re-authenticate with their new credentials.
* Added appropriate `autocomplete` attributes to password fields:

  * `username`
  * `current-password`
  * `new-password`
* Removed a leftover `console.log(response)`.
* Added a **Change Password** action to the authenticated user menu, ordered as:

  1. Locale
  2. Change Password
  3. Logout
* Updated the frontend password validation regex to match the backend implementation. Previously, the frontend incorrectly required an uppercase letter even though the backend (and the on-screen password requirements) did not. The password complexity message was also updated to reflect the actual validation rules.

## Screenshots

https://github.com/user-attachments/assets/ef09e55f-8ea7-4568-8326-b250995c5b07


src/main/java/org/openelisglobal/login/controller/ChangePasswordLoginController.java
```diff
@@ -111,14 +121,43 @@ public ModelAndView showUpdateLoginChangePassword(@ModelAttribute("form") @Valid
...
+     */
+    private ModelAndView writeApiResponse(HttpServletResponse response, Errors errors) {
+        JSONObject json = new JSONObject();
+        if (errors == null) {
+            response.setStatus(HttpStatus.SC_OK);
+            json.put("success", true);
+        } else {
+            response.setStatus(HttpStatus.SC_UNAUTHORIZED);
+            json.put("error",
+                    errors.getAllErrors().stream().findFirst().map(ObjectError::getCode).orElse("login.error.message"));
```
mherman22: The fallback to `login.error.message` is misleading. For example, an incorrect current password results in `Invalid login. Please try again.` Consider using a password change specific message instead, no?

mozzy11 (approved):
Excellent root-cause fix. I traced the flow end-to-end: the `apiCall=true` JSON contract in `ChangePasswordLoginController` cleanly separates real success from the security bounce that was fooling `response.redirected`, and the legacy JSP path is untouched. The controller integration tests are high quality — they assert actual DB outcomes via `loginService.getValidatedLogin` rather than mock returns — and the Playwright specs follow our best-practice guide (visible-state assertions, no force clicks).

Notes, none blocking:
- +1 to @mherman22's point on the `login.error.message` fallback — a change-password-specific message would read better; fine as a follow-up along with that thread.
- `writeApiResponse` returns 401 for confirm-password mismatch and other form validation errors; 400 would be semantically closer for those, though the SPA handles either.
- The header "Change Password" item uses `window.location.href`, forcing a full reload; react-router navigation would keep it in the SPA. Minor since the page renders fine either way.

CI is green and the regex/en.json alignment with `HaitiPasswordValidation` checks out. LGTM.

samuelmale:
@mherman22 can you have another look?

mherman22 (approved):

Issue #3943, opened 2026-07-28
Add integration tests for dataexchange module


Agaba-derrick:
The dataexchange package (org.openelisglobal.dataexchange) handles data exchange between OpenELIS and external systems — FHIR-based import/export with EMRs (e.g., OpenMRS), analyzer/instrument result ingestion, and order/result synchronization. This module currently has very  integration test coverage, which makes it easy for regressions in export/import logic to slip through

PR #3944, opened 2026-07-28
Add FhirTransformService Test

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
- [X] The PR includes a video showing the changes for the work done.
      documentation.
      project.

## Screenshots
Screenshot from 2026-07-28 12-41-05

## Related Issue
#3943

## Other

@HerbertYiga @mozzy11 @mherman22 Mind taking a look here ?


src/test/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceTest.java
```diff
@@ -0,0 +1,489 @@
...
+        mockFhirPersistanceService = mock(FhirPersistanceService.class);
+
+        Object target = fhirTransformService;
+        if (AopUtils.isAopProxy(target) && target instanceof Advised) {
+            target = ((Advised) target).getTargetSource().getTarget();
+        }
+
+        Field field = FhirTransformServiceImpl.class.getDeclaredField("fhirPersistanceService");
+        field.setAccessible(true);
+        field.set(target, mockFhirPersistanceService);
```
mozzy11: This swaps the field on the shared singleton bean but never restores it. The Spring TestContext is cached across test classes, so the mock (and its stubbing) leaks into every later test class that touches FhirTransformService. Please save the original value here and restore it in an @After method.

src/test/java/org/openelisglobal/dataexchange/fhir/service/FhirTransformServiceTest.java
```diff
@@ -0,0 +1,489 @@
...
+    public void transformPersistObjectsUnderSamples_shouldReturnPersistedBundle_whenSampleIdsProvided()
+            throws Exception {
+        List<String> sampleIds = Collections.singletonList("1");
+
+        Bundle bundle = new Bundle();
+        when(mockFhirPersistanceService.createUpdateFhirResourcesInFhirStore(any(FhirOperations.class)))
+                .thenReturn(bundle);
+
+        Future<Bundle> future = fhirTransformService.transformPersistObjectsUnderSamples(sampleIds);
+        assertEquals(bundle, future.get());
```
mozzy11, line 524: This asserts the mock's own return value round-trips through the Future — it can't fail unless the plumbing breaks. Consider also capturing the FhirOperations argument and asserting the bundle contains the expected resources for sample "1".

mozzy11 (changes requested):
Thanks — this is a solid batch of coverage for FhirTransformService, and most tests exercise the real service against a real dataset with meaningful value assertions (names, addresses, UUIDs), which is exactly the test style we want.

One real defect to fix before merge:
- **Mock leaks into the shared Spring context.** `setUp()` reflectively swaps `fhirPersistanceService` on the singleton `FhirTransformServiceImpl` bean, but never restores it. `BaseWebContextSensitiveTest` uses the cached TestContext (no `@DirtiesContext`), so every test class that runs after this one in the same JVM gets the Mockito mock with stale stubbing. Please capture the original bean in `@Before` and restore it in `@After` (or use `AopTestUtils.getTargetObject` + `ReflectionTestUtils.setField` for the swap and restore).

Minor:
- File is missing a trailing newline at EOF — please run `mvn spotless:apply`.
- The `transformPersist*` tests only verify the persistence call happened; capturing the `FhirOperations` argument (ArgumentCaptor) and asserting on the resources it contains would make them meaningfully stronger. Fine as a follow-up.

PR #3094, opened 2026-03-18
OGC-356: Decouple the lab entry workflow

reagan-meant:
# Pull Requests Requirements

- [ ] The PR title includes a brief description of the work done, including the
      Issue number if applicable.
- [ ] The PR includes a video showing the changes for the work done.
- [ ] The PR title follows conventional commit label standards.
- [ ] The changes confirm to the OpenElis Global x3
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide)
      documentation.
- [ ] The changes include tests or are validated by existing tests.
- [ ] I have read and agree to the Contributing
      [Guidelines](https://openelis-global.org/community/get-involved/) of this
      project.

## Screenshots

[Add relevant screenshots here if applicable]

## Related Issue

[Add a link to the related issue or mention it here if applicable]

## Other

[Add any additional information or notes here]

PR #3946, opened 2026-07-28
Sync/demo silnas into develop

reagan-meant:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


src/main/resources/liquibase/3.5.x.x/042-test-section-domain.xml
```diff
@@ -0,0 +1,53 @@
...
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <!--
+    Add Domain classification (CLINICAL / ENVIRONMENTAL / VECTOR) to Lab Unit (test_section).
+    CFG-6: Existing rows default to CLINICAL. Column becomes NOT NULL after backfill.
+    -->
+
+    <!-- Step 1: Add nullable column -->
+    <changeSet id="OGC-361-add-domain-nullable" author="OGC">
```
mozzy11: **Blocking — breaks the upgrade on every existing deployment.** All three changesets here lack `preConditions`, and this duplicates develop's `059-results-r1-unified-worklist.xml` (`OGC-1020-test-section-domain-column`), which *is* guarded by `not columnExists`. `base.xml` includes 042 at :173, before 059 at :305 — I confirmed on a fresh install that 042 runs `EXECUTED` and 059 goes `MARK_RAN`. Reverse that (any DB already on develop) and the unguarded `addColumn` aborts the changelog and the app won't start. Drop 042 in favour of 059 + 068, or guard all three.

src/main/resources/configuration/sample-types/vector-sample-types.csv
```diff
@@ -0,0 +1,6 @@
+description,localAbbreviation,domain,isActive,sortOrder,localization:en
```
mozzy11, line 1: **Blocking — the `domain` value never survives `Domain.normalize`.** `Domain.fromRaw` has no `"V"` case, so `normalize("V")` falls through to the CLINICAL default — this applies to whichever config source is active (classpath here, `volume/configuration` on a dev deployment).

Verified on `6e1e0a8b2` via `dev.docker-compose.yml`: `type_of_sample` ends up with **28 rows, all `CLINICAL`** — no `ENVIRONMENTAL`, no `VECTOR`. `/rest/vector-sample-types` and `/rest/environmental-sample-types` both return `[]`, `/rest/user-sample-types` (clinical order entry) leaks `Mosquito`/`Fly`/`Flea`/`Rodent`, and `vector_species` is 0 rows (219 `Unknown sampleTypeAbbrev 'MOSQUITO' — skipping …` warnings). See my follow-up comment for the separate duplicate-`EK` failure in `env-sample-types.csv`.

src/main/java/org/openelisglobal/typeofsample/daoimpl/TypeOfSampleDAOImpl.java
```diff
@@ -249,6 +249,8 @@ private List<String> getKeysForDomain(SampleDomain domain) {
         switch (domain) {
         case ANIMAL:
             return List.of(Domain.VECTOR.name(), "A");
+        case VECTOR:
+            return List.of(Domain.VECTOR.name(), "V");
```
mozzy11, line 269: `getKeysForDomain` learned `"V"` but `Domain.fromRaw` didn't, so the reconciliation is half-done: any admin edit or CSV import still normalises a vector sample type to CLINICAL. One-line fix — add `case "V":` alongside `case "A":` in `Domain.fromRaw`. With that, `064-preserve-vector-domain-code.xml` can also be dropped in favour of develop's 066.

src/main/java/org/openelisglobal/common/rest/provider/SampleEntryTestsForTypeProviderRestController.java
```diff
@@ -96,7 +102,31 @@ public SampleEntryTests processRequest(HttpServletRequest request, HttpServletRe
     @ResponseBody
     public List<IdValuePair> getUserSampleTests(HttpServletRequest request, HttpServletResponse response)
             throws ServletException, IOException {
-        return userService.getUserSampleTypes(getSysUserId(request), Constants.ROLE_RECEPTION);
+        List<IdValuePair> all = userService.getUserSampleTypes(getSysUserId(request), Constants.ROLE_RECEPTION);
+        // Filter to HUMAN domain only — environmental and vector have their own
+        // endpoints
+        java.util.Set<String> humanIds = typeOfSampleService
```
mozzy11: **Blocking regression** — intersecting with `getTypesForDomain(HUMAN)` (`CLINICAL`/`H`) drops every sample type whose `domain IS NULL`. `066`'s own header promises the opposite: *"domainless sample types remain offerable everywhere"*, and its backfill only touches `WHERE domain IS NOT NULL`. Upgraded sites lose those types from order entry, making their tests unorderable. Treat `domain IS NULL` as offerable.

src/main/java/org/openelisglobal/common/rest/provider/PatientDashBoardProvider.java
```diff
@@ -298,53 +327,68 @@ private List<OrderDisplayBean> convertElectronicToOrderBean(List<ElectronicOrder
...
         DashBoardMetrics metrics = new DashBoardMetrics();
-        java.sql.Timestamp startTimestamp = DateUtil
-                .convertStringDateStringTimeToTimestamp(DateUtil.getCurrentDateAsText(), "00:00:00.0");
-        java.sql.Timestamp endTimestamp = DateUtil
-                .convertStringDateStringTimeToTimestamp(DateUtil.getCurrentDateAsText(), "23:59:59");
+
+        // Resolve the logged-in user's assigned test section IDs.
+        // Empty list = no restriction (admin or no explicit assignments).
+        List<String> userSectionIds = resolveUserSectionIds(ControllerUtills.getSysUserId(request));
+        boolean restricted = !userSectionIds.isEmpty();
```
mozzy11: **Blocking — "today" tiles become all-time counts.** `getUserTestSections` returns *all* active sections for admins/ALL_LAB_UNITS, so `restricted` is true for nearly everyone and the comment's admin assumption doesn't hold. `countBySections` → `getCountAnalysisByTestSectionAndStatus` has no date predicate. Verified live: one analysis finalized 30 days ago, zero activity today → `ordersCompletedToday: 1` and `patiallyCompletedToday: 1` (also inverted). Needs a dated section-scoped count plus an explicit admin check.

src/main/java/org/openelisglobal/result/service/LogbookPersistServiceImpl.java
```diff
@@ -128,12 +141,74 @@ public List<Analysis> persistDataSet(ResultsUpdateDataSet actionDataSet, List<IR
 
         setSampleStatus(actionDataSet, sysUserId);
 
+        evaluateVectorResults(actionDataSet, sysUserId);
+
+        advanceReferralsForManualEntry(actionDataSet, sysUserId);
```
mozzy11, line 146: **Blocking.** This runs inside `@Transactional persistDataSet`, which already flushed `analysisService.update(analysis)` (lines 134-136) and holds that row lock; `markReferralCompletedFromManualEntry` is `REQUIRES_NEW`, so the inner tx updates the same `analysis` row on a second connection — an indefinite lock wait, not a catchable exception. It also force-sets `Finalized`/`releasedDate` at Result Entry, releasing results without passing Result Validation.

src/main/java/org/openelisglobal/sample/action/util/SamplePatientUpdateData.java
```diff
@@ -473,6 +584,26 @@ public void updateCurrentOrgIfNeeded(String code, String orgId) {
...
+    /**
+     * Requesting Organization's own phone/fax/email are updated whenever they
+     * differ from what's stored, independent of the referring-code change check in
+     * {@link #updateCurrentOrgIfNeeded}, so contact info stays fresh even when the
+     * code is unchanged.
+     */
+    public void updateOrganizationContactInfoIfNeeded(SampleOrderItem orderItem, String orgId) {
+        Organization org = currentOrganization != null ? currentOrganization : orgService.getOrganizationById(orgId);
+        boolean changed = StringUtil.compareWithNulls(orderItem.getReferringSitePhone(), org.getPhone()) != 0
+                || StringUtil.compareWithNulls(orderItem.getReferringSiteFax(), org.getFax()) != 0
```
mozzy11: **Blocking data loss.** `compareWithNulls` coerces null → `""`, so any caller that doesn't send the new `referringSite{Phone,Fax,Email}` fields computes `changed=true` and then nulls all three on the shared Organization row. Confirmed non-senders: `SampleBatchEntryRestController:190` and legacy `SamplePatientEntryController:247`. Skip absent/blank fields rather than treating absent as "clear".

src/main/java/org/openelisglobal/sample/controller/rest/OrderSearchRestController.java
```diff
@@ -171,18 +218,34 @@ public ResponseEntity<Map<String, Object>> getDashboard(@RequestParam(defaultVal
...
+            @RequestParam(required = false) String workflowType, HttpServletRequest request) {
 
         try {
             Map<String, Object> response = new HashMap<>();
             List<Map<String, Object>> ordersList = new ArrayList<>();
 
+            // Resolve the set of test section IDs the logged-in user is assigned to.
+            // Empty set = no restriction (admin or no assignments).
+            String currentSysUserId = getSysUserId(request);
+            Set<String> allowedSectionIds = resolveAllowedSectionIds(currentSysUserId);
```
mozzy11, line 251: With `REQUIRE_LAB_UNIT_AT_LOGIN=true`, `getUserTestSections(id, null)` returns only the single login lab unit, so Recent Orders silently collapses to one section for existing clinical deployments — develop returned them unfiltered and there's no flag to opt out. Please gate this behind a `site_information` flag (default off).

src/main/resources/hibernate/hbm/TestSection.hbm.xml
```diff
@@ -60,6 +56,10 @@
 
         <property name="sortOrderInt" column="sort_order"
             type="integer" />
+
+        <property name="domain" type="java.lang.String">
+            <column name="domain" length="20" not-null="true" />
```
mozzy11: `not-null="true"` makes Hibernate reject the insert regardless of the DB default from `068`, and the legacy MVC path `TestSectionCreateController.createTestSection` (line 143) never sets `domain` — so `POST /TestSectionCreate` will throw `PropertyValueException` (not a `LIMSRuntimeException`, so the local catch won't cover it). Default it there, or drop `not-null`.

src/main/java/org/openelisglobal/sample/validator/SamplePatientEntryFormValidator.java
```diff
@@ -65,7 +65,7 @@ private void validateSampleItem(Element sampleItem, Errors errors) {
         }
         // validate date not required
         String collectionDate = sampleItem.attributeValue("date").trim();
-        ValidationHelper.validateDateField(collectionDate, "sampleXML", "sampleXML date", errors, DateRelation.PAST,
+        ValidationHelper.validateDateField(collectionDate, "sampleXML", "sampleXML date", errors, DateRelation.ANY,
```
mozzy11: `DateRelation.PAST` → `ANY` means every deployment's SamplePatientEntry now accepts **future** specimen collection dates that were previously rejected. If a vector/environmental flow needs future dates, scope the relaxation to that domain rather than changing it globally.

src/main/resources/liquibase/3.4.x.x/025-split-workflow-menus.xml
```diff
@@ -0,0 +1,432 @@
...
+          - Add Vector Order      (/order/vector/*)
+
+        The old unified menu_order_workflow and its children are deactivated (not deleted)
+        so that existing role-permission rows referencing those menu IDs remain intact.
+    -->
+
+    <!-- ================================================================
+         1. Deactivate old unified workflow menus
+         ================================================================ -->
+    <changeSet id="nav-015-deactivate-old-order-workflow" author="reagan-meant">
```
mozzy11, line 21: This deactivates develop's shared Add Order tree (`menu_order_workflow` + 5 children) on **every** deployment and inserts 18 replacement rows — confirmed in the live DB (`is_active=f` on all six, new Clinical/Environmental/Vector branches under `menu_add_order`). A clinical lab upgrading gets its nav restructured plus branches it can't use, and `MenuController.resolveAllowedDomains` decides visibility from seeded test-section domains — so which branches appear varies by config rather than by deployment intent. Please split this out for explicit review or gate it behind a Liquibase `context`.

src/main/java/org/openelisglobal/referral/controller/rest/ReferenceLabResultsRestController.java
```diff
@@ -0,0 +1,187 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
```
mozzy11, line 38: **Blocking — no authorization.** No `@PreAuthorize` and no `system_module_url` row (checked the live DB), so the fail-open branch in `ModuleAuthenticationInterceptor:112` lets *any* authenticated user hit `PUT .../mark-lost` (:64), `.../accept` (:81), `.../reject` (:96) and `POST .../notify` (:116). Add a class-level guard like `SampleAcceptanceRestController`'s per-method pattern.

src/main/java/org/openelisglobal/vector/controller/rest/VectorSpeciesRestController.java
```diff
@@ -0,0 +1,102 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
```
mozzy11, line 37: **Blocking — unguarded `/rest/admin/**` writes.** POST (:76) and PUT (:89) have no `@PreAuthorize` and no module row, so any authenticated user can mutate reference data. Same in `VectorTrapTypeRestController` and `VectorSamplingSiteRestController`. Sibling admin controllers use class-level `@PreAuthorize("hasRole('ADMIN')")` (`DictionaryRestController:46`, `TestSectionCreateRestController:34`) — please match.

src/main/java/org/openelisglobal/vector/service/VectorSpeciesConfigurationHandler.java
```diff
@@ -0,0 +1,238 @@
...
+
+    private int findColumn(String[] headers, String name) {
+        for (int i = 0; i < headers.length; i++) {
+            if (name.equalsIgnoreCase(headers[i].trim()))
+                return i;
+        }
+        return -1;
+    }
+
+    private String[] parseCsvLine(String line) {
```
mozzy11: **Duplication.** This is a 14th copy of `parseCsvLine` — and it's the *old* buggy `inQuotes = !inQuotes` version, while this same PR adds a corrected RFC-4180 implementation in `ComplianceCsvUtil:31`. `ComplianceCsvUtil` is package-private so it can't be reused. Promote it to `public` in `org.openelisglobal.common.util` and delete the three new copies (here, `VectorTrapTypeConfigurationHandler:174`, `PanelConfigurationHandler:324`).

src/main/java/org/openelisglobal/vector/controller/rest/VectorDictionaryRestController.java
```diff
@@ -0,0 +1,153 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
```
mozzy11, line 21: **Duplication.** Eight of these nine endpoints re-implement `DisplayListController:648` `GET /rest/dictionary/category/{categoryName}` over the same `dictionaryService` — e.g. `/sampling-site-types` ≡ `/rest/dictionary/category/Sampling Site Type`. They also return raw `Dictionary` entities where develop returns `IdValuePair`. Keep only `/lifecycle-stages` (it has the real `sampleTypeId` join) and reuse the existing endpoint.

src/main/java/org/openelisglobal/compliance/controller/rest/ComplianceReportRestController.java
```diff
@@ -0,0 +1,636 @@
...
+        if (lhuAmendmentService.hasBeenReleased(sampleId)) {
+            try {
+                archiveService.archiveIfAbsent(sampleId, sample.getAmendmentNumber(), pdfBytes, userId);
+            } catch (Exception e) {
+                LogEvent.logError(e);
+            }
+        }
+    }
+
+    private byte[] buildOriginalPdfBytes(ComplianceReportOrderDTO dto) throws DocumentException, IOException {
```
mozzy11, line 292: ~380 lines of iText layout in a `@RestController` — the constitution wants thin controllers with logic in `@Transactional` services, and `ManifestPDFServiceImpl` is the established pattern. Also the section headers here (:266/:272/:279) are hardcoded English even though this PR adds `ComplianceMessages` wrapping `MessageUtil` for exactly this. Extract a `ComplianceReportPdfService`.

src/main/java/org/openelisglobal/shipment/service/ShipmentReceptionServiceImpl.java
```diff
@@ -0,0 +1,264 @@
...
+                String existingId = typeOfSampleService.getTypeOfSampleIdForLocalAbbreviation(abbreviation);
+                if (existingId != null && !existingId.isBlank()) {
+                    return; // already exists
+                }
+                String displayName = coding.hasDisplay() ? coding.getDisplay() : abbreviation;
+
+                Localization localization = new Localization();
+                localization.setDescription("sampleType name: " + displayName);
+                localization.setLocalizedValue("en", displayName);
+                localization.setSysUserId("1");
```
mozzy11: Hardcoded `sysUserId="1"` on a runtime ingestion path bypasses #3356's `UserContextHolder`/`daemonSystemUser`. Worse, this auto-creates `Localization` + `TypeOfSample` reference rows from inbound FHIR, so an external peer can drive reference-data creation. Use `UserContextHolder.requireSysUserId()` (or the daemon context) and gate the auto-create.

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirApiWorkFlowServiceImpl.java
```diff
@@ -212,90 +245,85 @@ private void beginTaskImportResultsPath(String remoteStorePath) {
...
-                        try {
-                            fhirReferralService.setReferralResult(resultsImportEntry.getValue());
-                        } catch (RuntimeException e) {
-                            LogEvent.logError(e);
-                            LogEvent.logError(
-                                    "could not import result for ServiceRequest: " + resultsImportEntry.getKey(), e);
-                        }
-                    }
+                if (!fetchReturnedResultsFromStore(remoteStorePath, referralTaskUuid).isEmpty()) {
+                    referralService.markReferralCompleted(sentReferral.getId(), "1",
```
mozzy11, line 260: Two behaviour changes worth calling out in the PR body: referral results are no longer auto-posted (moved to a manual Accept), and the remote `Task` filter narrows `ACCEPTED` → `COMPLETED`. Existing FHIR-referral deployments silently stop receiving results, and a sender on this branch paired with a receiver on develop (which writes `ACCEPTED`) never matches. Consider accepting both statuses plus a migration note.

frontend/src/languages/en.json
```diff
@@ -4716,6 +4716,7 @@
   "server.error.msg": "Oops, Server error please contact administrator",
   "show.aliquoting": "Show Aliquoting",
   "showing": "Showing",
+  "sidenav.label.addorder": "Add Order",
```
mozzy11: This duplicates the key already at :5703 (`"Add Generic Order"`). `JSON.parse` keeps the last occurrence, so this addition is inert *and* CI's `Duplicate key check` gate is red on it. Delete this line; if "Add Order" is the intended label, change :5703's value instead — the breadcrumbs at `OrderWorkflowLayout.jsx:117` / `OrderDashboard.jsx:96` currently render "Add Generic Order".

mozzy11 (changes requested):
Thanks for rebasing this onto current develop, @reagan-meant — the domain-code reconciliation and the precondition discipline across ~75 changesets are a clear step up from #3938. I built this branch and ran it live (`build.docker-compose.yml`, fresh DB) to check the core workflows; findings below are evidence-based, and I've flagged what's blocking vs. follow-up.

**Regressions (blocking).** `042-test-section-domain.xml` duplicates develop's `059` without preconditions — I confirmed via `databasechangelog` that 042 runs first on a fresh DB and 059 `MARK_RAN`s, so any already-upgraded deployment hits "column domain already exists" and the app won't start. `PatientDashBoardProvider` turns the four "today" tiles into all-time counts for essentially every user (reproduced: one analysis finalized 30 days ago → `ordersCompletedToday: 1`, `patiallyCompletedToday: 1`). `LogbookPersistServiceImpl:146` self-locks on the `analysis` row via `REQUIRES_NEW` and force-finalizes past Result Validation. NULL-domain sample types vanish from order entry, breaking the contract `066`'s own header promises. Batch/legacy order saves null the referring organization's phone/fax/email. Both the `plugins` (`8faf00561`→`f2e71d597`) and `dataexport` (`bc1c0d496`→`fad932776`) submodule pointers move **backwards** to ancestors of develop's, reverting the analyzer instance-ID and dataexport retry fixes. Order navigation is restructured for all deployments, and `003`/`004` changesets released in #3094 are deleted rather than superseded forward.

**Seed gating.** No instance gating is active (`instanceId` defaults null), so a stock compose install seeded the whole vector catalog — and wrongly: `Domain.fromRaw` has no `"V"` case, so those types stored as CLINICAL, `Mosquito` now shows in `/rest/user-sample-types`, `/rest/vector-sample-types` returns `[]`, and all 22 species were skipped. The three red `core-demo` vector specs are this bug, not flake. The 67 Indonesian sample types / 568 environmental tests / Baku Mutu thresholds also sit at the base config path; `{domain}/{instanceId}/` already exists for exactly this.

**Security.** Six new controllers still have neither `@PreAuthorize` nor a `system_module_url` row, so `ModuleAuthenticationInterceptor`'s fail-open branch applies — including `ReferenceLabResultsRestController` (any user can accept/reject/mark-lost referrals) and the `/rest/admin/vector/**` writes. `VectorSurveillanceRestController`'s Javadoc claims module protection that `042` doesn't grant.

**Duplication.** `VectorDictionaryRestController` re-implements `DisplayListController:648` eight times over; three *new* copies of the known-buggy `parseCsvLine` are added in the same PR that ships a corrected `ComplianceCsvUtil`; ~380 lines of iText live in a `@RestController`; three parallel jsPDF generators; `ReferredOutTestsRestController` and the still-active `menu_results_referred → /ReferredOutTests` menu row are left dangling after the route was deleted.

**Not tested / clean.** Session-scoped POST flows (actual order submission, result entry/validation UI, patient registration) weren't exercised end-to-end — my findings there are code-level, so please validate them manually. Genuinely clean: no new `@Transactional` in controllers, accession generation untouched, no SQL injection or path traversal in the new code, `LabOrderSearchProvider` strictly more defensive, and the Madagascar/patient migrations are byte-identical to develop and non-lossy.

**Recommendation: request changes.** The work is valuable and mostly well-built, but the Liquibase upgrade failure, the two submodule rewinds, and the dashboard/order-entry regressions would each break production on merge. One thing worth cherry-picking to develop independently: the Playwright install-timeout and dpkg-lock hardening in `e2e-playwright-reusable.yml:111,182-192` is a real generic flake fix. Given the 808-file scope, splitting this into (1) schema/domain reconciliation, (2) core-shared behaviour changes, (3) SILNAS distro features would make the clinical-regression surface reviewable.

mozzy11:
**Correction + additional findings — I re-ran this on the documented dev deployment.**

My earlier review measured config seeding on `build.docker-compose.yml`, which does **not** bind-mount `volume/configuration`. That's my mistake: it forced the classpath fallback in `ConfigurationInitializationService` and so described the wrong config source. I rebuilt on `6e1e0a8b2` via the documented flow (`mvn clean install` → `docker compose -f dev.docker-compose.yml up -d`, which mounts `volume/configuration` and the local WAR). Correcting the record — and the real picture is worse, not better:

**1. `env-sample-types.csv` fails to load entirely (new finding, blocking).** Lines 38 and 39 both declare `localAbbreviation=EK` — `Ekstoparasit` and `Ektoparasit` (a typo variant of the same term). The second insert throws and rolls the whole transaction back:
```
ERROR TypeOfSampleConfigurationHandler: Error processing line 39 in file env-sample-types.csv:
      Duplicate record exists for Ektoparasit
INFO  TypeOfSampleConfigurationHandler: Successfully loaded 59 sample types from env-sample-types.csv
ERROR AbstractPlatformTransactionManager: Failed to load sample-types configuration from file: env-sample-types.csv
```
All 59 parsed rows are discarded. Note the misleading "Successfully loaded" line logged immediately before the rollback — worth fixing so this failure mode isn't invisible in logs.

**2. Net result: zero sample types have a non-CLINICAL domain.** After a clean boot, `type_of_sample` contains **28 rows, all `CLINICAL`** — no `ENVIRONMENTAL`, no `VECTOR`. Consequently **both** new workflows are non-functional, not just vector:
- `GET /rest/environmental-sample-types` → `[]`
- `GET /rest/vector-sample-types` → `[]`
- `GET /rest/user-sample-types` (clinical order entry) → leaks `Mosquito`, `Fly`, `Flea`, `Rodent`

The 4 leaked rows come from the 4 lines this PR appends to the *generic* `example-sample-types.csv` (they don't exist on develop), and they land in CLINICAL because `Domain.fromRaw` has no `"V"` case. `vector_species` = **0 rows** — 219 `Unknown sampleTypeAbbrev 'MOSQUITO' — skipping …` warnings, since the species handler's literal `"V"` lookup can't match a CLINICAL row.

So the substance of my inline comment on `vector-sample-types.csv` stands, but please read it as: *the `domain` values in the new sample-type CSVs never survive `Domain.normalize`, whichever config source is active* — classpath on `build.docker-compose.yml`, filesystem on `dev.docker-compose.yml`. The `Domain.fromRaw` fix plus the duplicate-`EK` fix are both needed.

**3. Unchanged findings.** The dashboard regression reproduces identically here (one analysis finalized 30 days ago, no activity today → `ordersCompletedToday: 1`, `patiallyCompletedToday: 1`), and everything in my review that isn't config-source-dependent — the `042`/`059` Liquibase collision, the two submodule rewinds, the `REQUIRES_NEW` self-lock, the NULL-domain order-entry filter, the org-contact wipe, the six unguarded controllers, the duplication items — is unaffected.

Recommendation is unchanged: **request changes**.

src/main/resources/liquibase/3.5.x.x/042-test-section-domain.xml
mozzy11: **Fixed** in 82183dcb7. Dropped `042-test-section-domain.xml` entirely — develop's `059` (`OGC-1020`) already adds the column guarded by `not columnExists`, with `defaultValue="CLINICAL"` and the CHECK, so it fully supersedes this. That also made `068` redundant, so it's gone too. Restored the released `003`/`004` changesets plus their `base.xml` includes. Verified on a fresh boot: migration runs clean with no "already exists".

src/main/resources/configuration/sample-types/vector-sample-types.csv
mozzy11: **Fixed** in 82183dcb7 (`Domain.fromRaw` now maps `"V"`), plus the duplicate `EK` row removed from `env-sample-types.csv`. Live-verified on a fresh dev deployment: `type_of_sample` is now CLINICAL 24 / ENVIRONMENTAL 21 / VECTOR 42 (was 28 rows all CLINICAL), `vector_species` 91 and trap types 13 (was 0/3), `/rest/vector-sample-types` and `/rest/environmental-sample-types` are non-empty, and clinical order entry no longer leaks Mosquito/Fly/Flea/Rodent.

src/main/java/org/openelisglobal/typeofsample/daoimpl/TypeOfSampleDAOImpl.java
mozzy11: **Fixed** in 82183dcb7. Added the `"V"` case to `Domain.fromRaw`, and `getTypeOfSampleByLocalAbbrevAndDomain` now retries the canonical enum name when the raw legacy code misses — that second half matters because callers pass `"V"`/`"H"` while rows store `VECTOR`/`CLINICAL` (`TestServiceImpl:88` was affected too). Matched one value at a time rather than an `IN` clause because `local_abbrev` is not unique across domains and `IN` would risk `NonUniqueResult`.

src/main/java/org/openelisglobal/common/rest/provider/SampleEntryTestsForTypeProviderRestController.java
mozzy11: **Fixed** in 899ee9d2a. Null-domain sample types are treated as offerable in the clinical endpoint again, restoring the contract `066`'s header states. Live-verified: a newly created CLINICAL sample type (`LiveTest Plasma QA`) is offered in `/rest/user-sample-types` and its test is orderable end-to-end.

src/main/java/org/openelisglobal/common/rest/provider/PatientDashBoardProvider.java
mozzy11: **Fixed** in 899ee9d2a. Added dated section-scoped count methods, restored the exact status semantics of the unscoped branch (two tiles were inverted), and replaced the "empty section list means admin" inference with a real global-scope check. Live-verified both directions: an analysis finalized 30 days ago now leaves every "today" tile at 0 (previously `ordersCompletedToday: 1`, `patiallyCompletedToday: 1`), while one finalized today correctly reports 1.

src/main/java/org/openelisglobal/result/service/LogbookPersistServiceImpl.java
mozzy11: **Fixed** in 899ee9d2a. Dropped `REQUIRES_NEW` so the hook joins the caller's transaction (audited callers: only `LogbookPersistServiceImpl`; the rest are tests), and removed the force-advance to Finalized so the status transition is left to Result Validation. The FHIR push is now deferred to after commit using the `TransactionSynchronization` idiom already in this file, so it reads committed results and a store outage can't roll back the result save. I also added a no-downgrade guard in `transformPersistResultsEntryFhirObjects` so a completed remote ServiceRequest isn't reverted — that is the remedy `ManualEntryResultsSaveFhirSyncRaceTest`'s own failure message asks for, and it now passes.

src/main/java/org/openelisglobal/sample/action/util/SamplePatientUpdateData.java
mozzy11: **Fixed** in 899ee9d2a. Change detection and writes now treat blank/absent as "not supplied" (PATCH-style merge), so batch entry and the legacy controller — which post only `referringSiteId` — no longer null the shared Organization row. The same defect existed in `initProvider` and `initRequestorContact` on the shared master Person, and both got the same treatment.

src/main/java/org/openelisglobal/sample/controller/rest/OrderSearchRestController.java
mozzy11: **Fixed** in 899ee9d2a. The section filter is now opt-in behind a configuration flag that defaults to off, and it never applies to global-admin/AllLabUnits users or when `REQUIRE_LAB_UNIT_AT_LOGIN` is on (the case where `getUserTestSections` returns just the login unit). Existing clinical deployments behave exactly as develop did.

src/main/resources/hibernate/hbm/TestSection.hbm.xml
mozzy11: **Fixed** in 899ee9d2a. `TestSection.hbm.xml` is reverted to develop byte-for-byte; DB-side enforcement comes from `059` (NOT NULL + DEFAULT 'CLINICAL' + CHECK). Worth noting the real defect was slightly different from my original diagnosis: because the valueholder defaults to `"CLINICAL"`, `not-null` wouldn't have thrown — what actually broke the legacy page was the new `@NotBlank @Pattern` `domain` on the shared form while this controller's `ALLOWED_FIELDS` excludes it and its JSP renders no selector, so `@Valid` always failed and the POST silently returned `FWD_FAIL_INSERT`. A `@ModelAttribute` factory now seeds `domain=CLINICAL` before bind.

src/main/java/org/openelisglobal/sample/validator/SamplePatientEntryFormValidator.java
mozzy11: **Fixed** in 899ee9d2a. `PAST` is restored for the clinical path; `ANY` is now scoped to environmental/vector workflows via the `workflowType` discriminator (verified it's actually populated: `App.jsx` seeds it, `OrderContext.jsx` puts it in `sampleOrderItems.environmentalFields`, and `addOrder/Index.jsx` derives it from the accession domain).

src/main/resources/liquibase/3.4.x.x/025-split-workflow-menus.xml
mozzy11: **Not changed** — flagging for your decision rather than acting unilaterally. Deactivating the shared Add Order tree for all deployments is the intended shape of the feature as far as I can tell, so gating or splitting it is a product call, not a defect fix. My concern stands: a clinical-only deployment upgrading gets its nav restructured plus Environmental/Vector branches, and which branches appear depends on seeded test-section domains. Please confirm this is the intended upstream default.

src/main/java/org/openelisglobal/referral/controller/rest/ReferenceLabResultsRestController.java
mozzy11: **Fixed** in 899ee9d2a. Class-level guard for the dashboard GETs plus method-level guards on mark-lost/accept/reject/notify, using role names verified against `CustomUserDetailsService`'s `ROLE_<UPPER_SNAKE>` normalisation and the seeded `system_role` rows. Live-checked that admin still reaches the endpoints and anonymous is redirected to login.

src/main/java/org/openelisglobal/vector/controller/rest/VectorSpeciesRestController.java
mozzy11: **Fixed** in 899ee9d2a. `hasRole('ADMIN')` on POST/PUT for species, trap types and sampling sites; GETs left at the chain default since order entry and the vector UI consume them. Same treatment for the manual-entry field map (class-level, so the unmapped `PUT /{id}` is covered), requestor search, subcontract dispatch, and vector surveillance — whose Javadoc claimed module protection that changeset `042` never granted; the comment is corrected too.

src/main/java/org/openelisglobal/vector/service/VectorSpeciesConfigurationHandler.java
mozzy11: **Fixed** in 899ee9d2a. Promoted the corrected RFC-4180 parser to `common/util/CsvParsingUtil` and deleted the three new naive copies. `ComplianceCsvUtil` now delegates to it rather than being removed, because collapsing it fully would mean touching ~50 call sites across five compliance handlers — happy to do that as a follow-up if you'd prefer it gone.

src/main/java/org/openelisglobal/vector/controller/rest/VectorDictionaryRestController.java
mozzy11: **Not changed.** Consolidating these onto `DisplayListController`'s `GET /rest/dictionary/category/{categoryName}` is the right call technically, but it changes a public API shape the new vector UI already consumes and would also change the response type from raw `Dictionary` to `IdValuePair`. That felt like your call rather than mine mid-review. The finding stands: eight of the nine endpoints duplicate an existing one.

src/main/java/org/openelisglobal/compliance/controller/rest/ComplianceReportRestController.java
mozzy11: **Partially addressed.** Fixed the permanently-false `ROLE_SUPERVISOR` expression (there is no `Supervisor` role — `hasRole` would look for `ROLE_SUPERVISOR` while authorities are `ROLE_<UPPER_SNAKE>`) and converted the state-mutating `GET /exportPdf` to a POST, updating its only caller. **Not** done: extracting the ~380 lines of iText into a service, and moving the hardcoded English section headers onto the `ComplianceMessages` helper this PR already adds — both are larger refactors better done deliberately than folded into a review-fix pass.

src/main/java/org/openelisglobal/shipment/service/ShipmentReceptionServiceImpl.java
mozzy11: **Fixed** in 899ee9d2a. Resolves the acting user through `UserContextHolder` (current user, falling back to the daemon id since ingestion can run off-request) instead of the literal `"1"`. Also added a default-on config gate for the reference-data auto-create, so the feature is unchanged out of the box but a site curating its own catalog can turn it off.

src/main/java/org/openelisglobal/dataexchange/fhir/service/FhirApiWorkFlowServiceImpl.java
mozzy11: **Not changed** — this needs a product decision, so I've left it as-is and am flagging it rather than guessing. As it stands, existing FHIR-referral deployments stop receiving results automatically, and a sender on this branch paired with a receiver on develop (which writes `ACCEPTED`) never matches at all. If manual Accept is the intended UX, the poll should probably still accept both statuses during the transition, and this deserves a release note.

frontend/src/languages/en.json
mozzy11: **Fixed** in 82183dcb7. Deleted the duplicate line; the new nav uses `sidenav.label.add.order` (which exists), so the added `sidenav.label.addorder` entry was inert and only served to break the duplicate-key gate. Verified `en.json` parses and has no duplicate keys.

mozzy11:
## Review follow-up pushed — fixes, merge resolution and live verification

Pushed `3bb39d81c` (merge), `82183dcb7` and `899ee9d2a`. I've replied inline to each thread; summary here.

### Merge with current develop
One content conflict — `SampleTypeRequestDAOImpl`, where both sides fixed the #3642 Integer/String bind independently; kept develop's `normalizeSampleId` form from #3701. Three things the merge would otherwise have carried through silently:
- `plugins` pointed at `f2e71d597`, an **ancestor** of develop's `8faf00561` — that would have reverted the analyzer instance-ID fixes (#73, #72, #63). Forwarded.
- `dataexport` likewise `fad932776` vs `bc1c0d496` — reverting the retry-loop fixes. Forwarded.
- `projects/catalyst`'s gitlink had been dropped while `.gitmodules` still declared it. Restored.

### Liquibase
Dropped the duplicate `042-test-section-domain.xml` (develop's guarded `059` supersedes it, and it was the cause of the upgrade abort) and the now-redundant `068`; restored the released `003`/`004` and their includes; wired up the orphaned `030`; de-duplicated `037`'s changeset id. Verified programmatically that all XML parses, every include target exists, nothing is orphaned, and no duplicate `(id, author)` pairs remain beyond two pre-existing on develop.

### Live verification (`dev.docker-compose.yml`, fresh DB)

| Check | Before | After |
|---|---|---|
| `type_of_sample` domains | 28 rows, **all CLINICAL** | CLINICAL 24 / ENVIRONMENTAL 21 / VECTOR 42 |
| `vector_species` / trap types | 0 / 3 | 91 / 13 |
| `env-sample-types.csv` | rolled back on duplicate `EK` | loads 59 types |
| `/rest/user-sample-types` | leaked Mosquito/Fly/Flea/Rodent | no leak |
| `/rest/environmental-sample-types` | `[]` | 21 |
| `/rest/vector-sample-types` | `[]` | 42 |
| Liquibase upgrade | 042 vs 059 collision | clean |

**Backend suite: 5,280 tests, 0 failures, 0 errors** — including `ManualEntryResultsSaveFhirSyncRaceTest`, which passes via the no-downgrade guard its own failure message asks for.

### Core clinical workflow, exercised end-to-end
Created a sample type through the config loader, a test through Add Test, then drove the full chain:

- Sample types **38** `LiveTest Plasma QA` (CSV `domain=H` → CLINICAL) and **39** `LiveTest Mosquito Pool` (CSV `domain=V` → **VECTOR**) — row 39 is the fix, and each appears only on its own endpoint
- Test **380** `LiveTest Glucose QA`, LOINC `LT-GLU-9001`, Biochemistry, range 3.90–6.10, critical 2.0–25.0; linked to type 38 (`sampletype_test` **254**)
- Patient **3** Amina Nakato, F, DOB 14/03/1991, `LIVETEST-260728200703`
- Order **DEV01260000000000001** (sample 2, item 2, analysis 2) → edited (collection date persisted) → result **5.4** entered → analysis **Technical Acceptance** → validated → **Finalized**, released 2026-07-28 → 3 report PDFs generated, each containing the patient, accession, test name and 5.40

The clinically important part: the analysis reached Finalized **only through validation**. Result entry left it at Technical Acceptance, which is the gate the force-advance had bypassed.

### Deliberately not changed — needs your call
- **`3.4.x.x/025-split-workflow-menus.xml`** — deactivating the shared Add Order tree for every deployment looks like the intended shape of the feature; gating or splitting it is a product decision. Please confirm it's the intended upstream default.
- **`FhirApiWorkFlowServiceImpl` `ACCEPTED` → `COMPLETED`** — existing FHIR-referral deployments stop receiving results automatically, and a sender on this branch with a receiver on develop never matches. Suggest accepting both statuses during transition, plus a release note.
- **`VectorDictionaryRestController`** — consolidating onto `DisplayListController`'s dictionary-category endpoint would change a public API shape the new UI consumes.
- **`ComplianceReportRestController`** — fixed the dead `ROLE_SUPERVISOR` role and the state-mutating export GET; the iText-in-controller extraction and moving hardcoded English headers onto `ComplianceMessages` are left as follow-ups.
- **Seed-data location** — the Indonesian/Baku-Mutu CSVs stay at the base config path so the environmental/vector features still seed on a dev deployment. Related cosmetic note: the vector dropdown now shows "Mosquito" twice, because `Nyamuk dewasa` (NYDE, env CSV) localises to "Mosquito" alongside the `MOSQUITO` row in the generic example CSV — worth de-duplicating on your side, since deleting a seed row could break vector E2E specs.

### Pre-existing issues found while testing (not from this PR)
- `TestAddControllerUtills.extractTestAddParms` throws a 500 NPE when a numeric test omits `lowValid`/`highValid` etc., instead of a validation error.
- The results and sample-edit forms are asymmetric: the GET emits `reportable: "N"` and `existingTests` shapes the POST cannot deserialize, so round-tripping a GET payload fails with `HttpMessageNotReadableException`.

mozzy11:
## Merge with current develop — resolution, regressions found and fixed, verification

Pushed `90af467c0` (merge) + `de718019e`, `fad959ea1`, `c0c406825`, `70f0061dc`. The PR is conflict-free again and **all 10 CI checks pass**, including the authoritative E2E executor. Summary of what the merge contained and what it surfaced.

### Merge resolution (20 conflicted files + 1 submodule)

- **Pure unions** (both sides added at the same spot): `AlertType`, `ConfigurationProperties`, `ResultLimitService`+Impl, `message_en.properties`, `AppTestConfig` package scan, `ResultsLoadUtility` (this branch's vector-pool helpers + develop's structured-notes helpers), `SearchResultForm.handleSave` (holding-time note + OGC-811 `resultFile` strip), `frontend/.gitignore`, and `example-sample-types.csv` — develop's new `loinc` column kept, this branch's four vector rows re-appended with an empty LOINC cell, which the import treats as "not specified", never "clear".
- **Develop's side taken where it supersedes**: specimen naming in `AnalysisServiceImpl` (`specimenNameFor` covers the old "Variable" special case), the null-safe dashboard patient lookup, the strict test-profile Jackson mapper, `sidenav.label.addorder = "Add Generic Order"` (this branch's own menu changeset depends on that rename staying) and `"Sample Type Editor"` (pinned by tests on develop).
- **This branch's side kept** where develop had nothing to say: `OrderCollect` imports (all used), `order-workflow.scss` consent/env-manifest styles, `App.jsx` `Redirect` import (develop's `App.css` deletion honored).
- **Real synthesis** in `OrderEntryValidationSchema`: domain-aware patient-validation skip (E/V) **and** develop's `REQUESTER_REQUIRED` toggle (#4003) now live in one factory — verified by the callers that pass `(configurationProperties, domain)`.
- **Liquibase**: this branch's `068` renumbered to `085` (develop shipped 068–084 in the meantime); every changeset in it is guarded on the pre-fix state so an already-patched deployment MARK_RANs. Whole changelog validated programmatically — every include resolves, no orphans, no new duplicate ids.
- **Submodule**: `tools/openelis-analyzer-bridge` pointed at an **ancestor** of develop's pointer — keeping it would have reverted the QC lot-scanning work (#39 and review fixes). Forwarded.

### Three regressions the merge surfaced — found and fixed

1. **Order-dependent Spring bean cycle** (`de718019e`): `resultService → resultLimitService → analysisAnchorService → sampleService → analysisService → resultService`. Edges B/C are this branch's vector work; the entry edge is a legacy `SpringContext.getBean` **field initializer** in `ResultServiceImpl` that runs mid-construction, where no early reference exists. Latent on this branch (its own suite passed by bean-creation-order luck); on the merged graph develop's new beans flip the entry point and the context dies — 3,853 cascading test errors, and the same coin-flip would apply to a production boot. Fixed by making the branch's new `analysisAnchorService` reference `@Lazy`.
2. **`chk_alert_type` last-writer-wins** (`fad959ea1`): every changeset that adds an alert type redeclares the whole CHECK constraint. This branch's `057` added the referral types; develop's `070` added `CRITICAL_RESULT` from a list that never knew them — on the merged changelog `070` runs last, and inserting `REQUIRED_BY_DEADLINE` / `REFERRAL_*` fails the check (the referral accept/reject flow errors exactly as its tests error). Changeset `086` redeclares the union. `ReferenceLabResultsAcceptRejectTest` went 6 errors → 14/14.
3. **A merge-artifact null field**: the conflict-context line above one hunk was the `@Autowired` for develop's `testResultComponentService`, so a naive union left it unannotated — null at runtime, breaking component-scoped range selection behind Results Entry and Validation on first use. Caught in review, fixed in the same commit as the union.

Two develop-side tests adapted to collaborators this branch introduced (`c0c406825`): the dashboard provider test wires `AnalysisAnchorService` (pool-anchored analyses resolve their sample through it), and `AlertNotificationServiceTest` now pins the new contract — this branch moved email transport onto the `SMTP_SERVER` external connection, so the container-injected `JavaMailSender` bean develop's test verified is no longer part of the path.

### Verification

- Backend, full suite on the merged tree: **5,572 tests, 0 failures, 0 errors**
- Frontend: **1,238 passed / 177 files**
- **Fresh database boot**: all 1,077 changesets apply clean through `086`; `type_of_sample` seeds CLINICAL 24 / ENVIRONMENTAL 21 / VECTOR 42; vector species 91, trap types 13; no vector leak into clinical order entry; `chk_alert_type` carries all four merged type families; `menu_add_order` points at the renamed key.
- **Hands-on in the browser** (fresh stack): unified `/Results` renders; Test Catalogue list opens a test from a data cell while the checkbox only selects; **sample type created and edited (persisted)**; **test created through create-in-place and edited (persisted — the name field is readonly in edit by design; renames go through Localization)**; **panel created and edited (persisted)**; Localization sections load for both new entities; legacy Manage Sample Types untouched; validation page, `/SamplePatientEntry`, `/order/clinical` and `/order/vector` all render. Zero JS page errors across every page visited.

### Follow-ups (not blockers, being explicit)

- **Email transport has no test coverage now**: the new `EmailNotificationSender` builds its own SMTP session from the external connection, so no test exercises actual message assembly/dispatch. Needs a GreenMail-style test against a local SMTP.
- Fresh-boot `example-lab-roles.csv` errors ("No row with the given identifier") are **pre-existing and byte-identical on both parents** — not from this merge, but worth a look separately.
- The product decisions from the July review remain open: the `3.4.x.x/025` Add-Order menu tree as an upstream default, the FHIR referral status transition (both `ACCEPTED` and `COMPLETED` currently polled), and `VectorDictionaryRestController` vs the dictionary-category endpoint.

mozzy11 (approved):

PR #3952, opened 2026-07-28
chore(ci): add Trivy scans, SBOM generation, and image security artifacts for release workflows

### LINKED ISSUE #3951: Enable dependency and container image scanning with SBOM/provenance checks
## Description:
Improve supply-chain resilience by scanning application dependencies, container images, and build artifacts for known vulnerabilities. The effort should include dependency pinning, SBOM or provenance review, and automated checks in CI/CD so that vulnerable or tampered components are caught before deployment.

bahati308:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
This PR hardens the CI/CD pipeline by adding automated supply-chain security checks before images are published.

### What changed:

Added Trivy-based vulnerability scanning for backend and frontend dependency trees
Added image vulnerability scanning for published container images
Added SBOM generation for backend, frontend, and published images
Uploaded scan and SBOM artifacts for review in CI
### Why:
This improves resilience against vulnerable or tampered dependencies and container images before deployment, supporting the repository’s threat-model and security requirements.
## Screenshots

## Related Issue
fixes #3951 

## Other


pmanko (changes requested):
Thanks for this, Brian — the shape of this is right, and a few of the hardest details are right in ways most first attempts get wrong. `publish-images.yml:193` scans `steps.source.outputs.ghcr_source`, the exact commit-pinned GHCR image E2E tested and line 220 retags, rather than rebuilding. You added no new jobs and touched no `needs:` edges, so the Checkpoint aggregators and their `skipped` tolerance still work — that's the trap here. And `publish-images.yml:205` derives `artifact-name` from `matrix.dockerhub_suffix`, which is exactly what's needed to avoid an upload collision across the 5-leg matrix.

Requesting changes on the following.

**1. `aquasecurity/trivy-action@0.24.0` does not exist — this is why CI is red.** `backend.yml:73`, `frontend.yml:134`, `publish-images.yml:191`. Upstream tags are `v`-prefixed: `refs/tags/0.24.0` → 404, `v0.24.0` → `6e7b7d1f`. Both failing jobs die in ~2s at "Set up job", so spotless, the Maven build, `npm ci`, lint and Vitest never run. Pin `@v0.36.0` in all three places. Fix line 191 too — `publish-images.yml` only runs on `workflow_run` of "03 - E2E" (never on PRs), so a broken ref there would silently break DockerHub publishing post-merge.

**2. `exit-code: 1` makes these hard gates that fail on today's tree.** `backend.yml:79`, `frontend.yml:140`. The clearest case is structural rather than a drifting CVE count: this repo intentionally tracks a dev TLS bundle (`dev/https/oe_selfsigned.key`, documented in `docs/dev_setup.md`), and with `scanners: vuln,secret` Trivy flags it `HIGH: AsymmetricPrivateKey`. I ran trivy 0.53.0 with your exact flags against those files: exit code 1. There is also live dependency noise (`pom.xml`, `frontend/package-lock.json`), all `Status: fixed`, so `ignore-unfixed` suppresses none of it. Since that key can't just be deleted, please land these report-only (`exit-code: 0`) and commit a reviewed `.trivyignore` (wired via `trivyignores:`) before flipping to blocking — matching the advisory pattern already at `frontend.yml:112-118`.

**3. Move the backend scan after the build.** It sits at `backend.yml:72`, ahead of `check formatting` (100) and `mvn clean install` (110), and on push to develop ahead of the JaCoCo badge (115) and Pages publish (132) — so a newly-disclosed upstream CVE costs the team build, test and coverage signal.

**4. Add `if: always()` to the SBOM and upload steps** — `backend.yml:84`/`92`, `frontend.yml:145`/`153`, `publish-images.yml:200`/`207`. None of the three upload steps has an `if:`, and each directly follows a step with `exit-code: 1`. Because `output:` sends the table to a file rather than stdout, on a real finding the report is in neither the log nor an artifact — it uploads only when there's nothing to report. Repo precedent: `e2e-playwright.yml:284`.

**5. Make the publish-path scan advisory** — `publish-images.yml:196`. With `fail-fast: false` in the matrix, one failing leg skips only its own push, leaving `:develop` a mixed-commit stack; and `deploy-testing` (227) has no `always()`, so it's skipped too. Base images (`nginx:1.15-alpine`, `postgres:14.4`) will trip this routinely.

**6. `dependency-snapshot: true` can't work as configured** — `backend.yml:90`, `frontend.yml:151`. It submits to GitHub's Dependency Submission API, which needs `contents: write`, but every job in these two files declares `permissions: contents: read`; fork PRs get a read-only token regardless. Gate it to `github.event_name == 'push'` with explicit `permissions: contents: write`, or drop it.

**7. Frontend artifact paths don't match.** `frontend.yml:139`/`149` write to the workspace root, but `158-159` upload `frontend/…`. `working-directory` isn't valid on `uses:` steps (the job sets it per-`run:`-step only). Drop the prefix — backend gets this right at `96-98`. Note the SBOM still ships anyway via sbom-action's own upload (`upload-artifact` defaults true), so the explicit SBOM upload is redundant in both files; only the trivy report is lost.

**8. Nits:** use `actions/upload-artifact@v6` (`backend.yml:93`, `frontend.yml:154`, `publish-images.yml:208`) — the other 16 call sites in the repo are all v6. And `frontend.yml:77-78` `provenance`/`sbom` sit on a `push: false` build with no exporter, so they're generated and discarded; they belong where images are actually pushed.

Happy to re-review quickly once the pin and the exit-code policy are settled — the version fix is one character, and I'd rather see this land advisory-first than sit red.

pmanko (approved):
Re-reviewed at 7111a2b8 — approving. All eight points from my Aug 4 review are addressed: trivy-action@v0.36.0 across all three workflows, advisory exit-code 0 with the ignore file wired in, backend scan moved after the build, if: always() on the SBOM/upload steps, upload-artifact@v6, matrix-suffix artifact naming, and the push-gated dependency-snapshot resolves the fork-permissions concern more cleanly than what I originally suggested. CI is green and the branch is current with develop.

One follow-up for whenever we consider making these scans blocking (not needed now, while they're advisory): the .trivyignore entry uses a `path: rule-id` shape that plain .trivyignore doesn't support, so it's likely inert — we should move it to .trivyignore.yaml or drop it before any enforcement decision. Optional polish at the same time: if: always() on the scan steps themselves.

Thanks Brian — this is the first automated security scanning this repository has ever had. Nice work seeing it through review.

pmanko (approved):

PR #3956, opened 2026-07-29
fix(qa-qi): wire CALLBACK into qi_config thresholds + make the callback SLA configurable

samuelmale:
## Why

The opt-in **Critical Callback Compliance** indicator shipped (OGC-714/715) consuming only *one* of the three things `qi_config` drives. It had the disable cascade, but:

- its QI Dashboard tile hardcoded `accent="blue"` while TAT/REJECTION/AMENDMENT colour against resolved target/action, and
- the breach evaluator never evaluated it — so a compliance breach created **no NCE**, even though the QI Configuration editor renders a note under the CALLBACK thresholds promising exactly that.

That's C.3 gap #2 ("tiles don't consume thresholds") and gap #5 ("silent no-op auto-NCE for non-AMENDMENT") reopened by the new indicator — this time with the UI actively asserting behaviour the backend didn't implement.

Separately, the 60-minute callback SLA was a `private static final` constant. TJC NPSG.02.03.01 requires the lab to **define** its critical-result reporting timeframe; it does not fix 60 minutes. A lab on a 30-minute policy was reading a compliance number that was wrong by definition.

## What changed

**CALLBACK behaves like its four siblings**

- `QiIndicator.CALLBACK` `thresholdsRequired` `false` → `true`, matching what `QIConfigEditor` already enforced client-side (the two layers disagreed).
- New `evaluateCallback` in `QiBreachEvaluatorServiceImpl` — month-to-date read-back compliance %, through the shared `checkAndFire` so HIGHER_BETTER direction comes from config, not a special case.
- Dashboard tile uses `accentFor`/`thresholdLineFor`. Its redundant `Target: N%` line becomes the vs-prior caption its four sibling tiles carry.
- QA Overview `CallbackComplianceTile` drops a private two-band tone for the shared `rateTone`, with `config` passed down by `TodayTiles` like `RejectionRateTile`/`AmendmentRateTile`. **Behaviour fix beyond the stated scope:** it was banding off `target` alone and ignoring `action`, so 96% (between action 95 and target 100) read red where the dashboard now reads amber.

**SLA is configurable**

- `org.openelisglobal.qi.callback.sla.minutes` (default 60), feeding the compliance numerator and the `overTarget` failure reason.
- The ack-distribution histogram is deliberately **decoupled** — its buckets are absolute latency, so its labels stay truthful at any policy. Only the pass line moves.
- Three i18n strings hardcoded "60 minutes" and no longer do. The configured value surfaces once, data-driven, on the detail page's rate label via a new `slaMinutes` field on `CallbackSummaryResponse`.

**Index**

`qa/012` indexes `analysis.released_date`, preconditioned on the index not already existing. The evaluator polls the released-cohort computes every 2 minutes over two windows and that column had no index — `qa/011` did the same for `started_date` when the rejection compute joined the poller. Benefits Amendment and TAT too.

## Testing

33 backend integration tests + 51 frontend tests green, on a fresh rebase onto `feat/qa-next`.

New coverage:
- three evaluator cases for the CALLBACK path (breach fires / above-action fires nothing), each guarded against vacuous passes by asserting the metric is computable first;
- a widened-SLA test proving the 150-minute call flips to compliant at 180 while the histogram stays put — the inversion test for the property;
- tile-banding tests for the one HIGHER_BETTER indicator, green at target and red at action.

`mvn spotless:check` run cold (cache cleared) and prettier `--check`: both clean.

## Reviewer notes

- **Small denominator, deliberately not guarded.** One critical result in a quiet month with no callback logged → 0% → auto-NCE. `checkAndFire` has no minimum-denominator guard and neither do AMENDMENT/REJECTION, so this is consistent — but criticals are structurally rarer. Arguably correct for TJC (a missed critical callback *is* a nonconformity); flagging it as a product call rather than deciding it here.
- **Exact-equality asymmetry, pre-existing.** At a value exactly equal to the action threshold, `rateTone` shows red but `breaches()` doesn't fire (`<=` vs `<`). Identical for all four indicators, so CALLBACK now inherits it consistently. Fixing it for CALLBACK alone would *create* an inconsistency; fixing all four is a separate one-liner I'd rather not smuggle in here.
- **Untested path:** the changeset is proven against a fresh Testcontainers Postgres (the harness runs the full changelog at context startup). The upgrade path on a long-lived production DB — where `createIndex` takes a brief ACCESS EXCLUSIVE lock on `analysis` — is not exercised by CI. Same exposure `qa/011` already shipped with.


samuelmale:
## CI fix folded in (`30c3327`) — pre-existing base breakage, not this PR

The first run failed with 25 errors, all `Failed to load ApplicationContext` in `ElectronicSignatureLogSecurityTest` and `QCExportRestControllerSecurityTest` — two classes this PR doesn't touch. Root cause:

```
No qualifying bean of type 'org.openelisglobal.security.DaemonContextExecutor'
  required by 'defaultConfigurationProperties'
  required by 'qcExportRestController' / 'electronicSignatureRestController'
```

`DaemonContextExecutor` arrived with **#3356** (Daemon/System user + centralised UserContext), which added it as an `@Autowired` field on `DefaultConfigurationProperties`. Both controllers pull that in via `ConfigurationProperties`, and neither slice's `TestConfig` provides the bean.

**Verified pre-existing:** checked out `upstream/feat/qa-next` with this PR's feature commit absent and ran the two classes — 14 esig + 11 QC export = **exactly the 25 errors CI reported**. #3356 came from `develop` while the esig and QC-export slices live on `feat/qa-next`, so the two never met until the merge commit. A semantic conflict neither side could see; #3356 *did* fix the slices it knew about.

The fix applies the identical pattern from #3356's own `ConfigurationReloadRestControllerSecurityTest` — the bean plus its `@Qualifier("daemonSysUserId")` dependency — and is kept as its own commit so it stays cherry-pickable if you'd rather this repair landed separately.

Worth knowing: `feat/qa-next` is currently red for **every** PR targeting it for this reason, not just this one.

PR #3970, opened 2026-08-03
Bump dev proxy nginx to 1.30.4-alpine

mherman22:
The dev proxy pins `nginx:1.15-alpine` from 2018. Its bundled OpenSSL has `ARMv8 ChaCha20-Poly1305` assembly that hits an illegal instruction under Docker Desktop emulation on Apple Silicon, so any client that prefers ChaCha20 (curl's default cipher order) kills the nginx worker mid-handshake and gets connection resets, while browsers preferring `AES-GCM `work fine. That asymmetry makes local API testing fail in ways that look like application bugs.

The dev nginx.conf parses clean on 1.29-alpine, checked with nginx -t on both images. After this change a plain curl against https://localhost works on Apple Silicon without cipher workarounds.


ibacher:
Shouldn't this target 1.30.4 (nginx's stable) or 1.31.3 (nginx's latest)?

PR #3976, opened 2026-08-04
fix(OGC-1153): Test Catalog Editor localization, empty states, and id parsing

mozzy11:
## Summary

Fixes all six defects in [OGC-1153](https://uwdigi.atlassian.net/browse/OGC-1153) (unified Test Catalog Editor).

**Two of the ticket's claims did not hold on `develop`, and I did not implement what it asked for in those two cases.** Both are called out below with evidence, because acting on them literally would have added duplicate code.

### 1. Localization section — HIGH

- **Opened on the wrong locale (real).** The picker chose `list.find(l => !l.fallback)`, so with `[en(fallback), fr]` an English session opened the **French** record. Untranslated fields fall back to English, so both rows render identically and the admin has no cue. Now resolves the session locale: exact match → same base language (`en-GB` → `en`) → fallback locale → first entry.
- **Could not tell whether an edit saved (real).** Saving happened on `onBlur` with no Save control and no auto-save claim. The section now states that fields save automatically, matching the Display Order precedent.
- **NOT done: "the write endpoint does not exist".** The 405 is on `/rest/test-catalog/tests/{id}/localization`, which is *deliberately* a read-only pointer lookup (see its `Localization (OGC-767)` comment block). The write path is `/rest/localizations/{id}/translations`, and `LocalizationSection` already used it. Verified live on this branch: `PUT /rest/localizations/11/translations` → **200**. Adding a second write endpoint would have duplicated a working one.
- The fallback indicator the ticket asks for already exists (Tag + tooltip + placeholder, added by OGC-767 #3797 — which post-dates the 3.2.1.11 build QA tested).

### 2 & 3. Empty states — MEDIUM / LOW

- Sample & Results rendered the empty-state notification **and** an unconditional four-column table header with no rows. Table is now suppressed when the list is empty. Fixed for **interpretations** (reported) *and* **select-list options**, which had the identical defect the ticket missed.
- Reflex & Calc rendered its calculation empty states as a bare `<p>` with no typography token while the sibling Reflex block used an inline notification. Both now match that block — the ticket named only "calculated by"; "feeds into" was the same. Each direction gets its own sentence rather than sharing a key whose value was the single word `"None"`, which reads poorly as a notification title.

### 4. Non-numeric `{id}` → 500 — LOW

`TEST.ID` maps through `LIMSStringNumberUserType`, whose `nullSafeSet` is a bare `Integer.parseInt`; the `NumberFormatException` escapes the DAO and `ControllerSetup` (`@Order(HIGHEST_PRECEDENCE)` on `RuntimeException`) renders it as 500.

Fixed **once for the URL space** with an interceptor on `/rest/test-catalog/**`, per the ticket's request for one fix rather than endpoint-by-endpoint. The two obvious alternatives cannot work here:

- a new `@ControllerAdvice` is `LOWEST_PRECEDENCE`, so `ControllerSetup` claims the exception first and the advice never runs;
- `@Validated` + `@Pattern` throws `ConstraintViolationException`, itself a `RuntimeException` → still 500.

The interceptor also covers `TestReagentLinkRestController` and `TestAlertRuleRestController`, which sit outside the `testcatalog` package and had the same 500 — a package-scoped advice could not have reached them. A classpath-scanning test fails the build if a new path variable is left unclassified.

**Reviewer note:** an MVC interceptor runs before `@PreAuthorize`, so an authenticated non-admin now gets **404** rather than 403 for a *malformed* id. Anonymous still gets 401 and well-formed ids still get 403 (both pinned by tests). 404 leaks less than 403 here; reverting it would mean giving up the cross-package coverage.

### 5a. Activation — orderable coupling

`/activate` sets `orderable` per the FRS lifecycle (Active ⇒ orderable) but the response did not say so and the form hardcoded `active: true`, so the flag only appeared after a reload. The response now reports the resulting `active`/`orderable` as a **strict superset** of the coverage report (`male`/`female` stay top-level, so existing callers are unaffected), and the form applies it.

**NOT changed:** `setOrderable` itself, and the 409 / `gapsAcknowledged` re-POST flow. An `active:false, orderable:true` row is inert — the order picker requires both (`TestDAOImpl` filters `isActive='Y' AND orderable=true`), so this was never the functional bug it reads as.

### 5b. "Infinity years" in a danger modal

An open-ended trailing gap is `Double.POSITIVE_INFINITY` (`RangeCoverageValidationService:197`), which Jackson serialises as the **quoted string** `"Infinity"`, so `formatAgeDays` rendered the literal text `Infinity years` inside the destructive-action dialog. Now renders a translated open-ended label, guarding both the string and the number form. Two directional keys, because `gapRange` is `"{from} – {to}"` and a single string would read backwards for an open *lower* bound.

## Testing

- **Backend: 4786 tests, 0 failures** (includes the new `TestCatalogNumericIdGuardTest`, and the activation integration test).
- **Frontend: 130 tests** across the 20 `testCatalog` suites.
- **New Playwright spec** `foundational/core/ogc-1153-test-catalog-editor.spec.ts` — 8 tests, one per defect, **passing against a live dev stack**.
- Inversion-verified: running the new frontend tests against `develop`'s original files fails exactly the OGC-1153 tests and no others.

Live-tested on `dev.docker-compose.yml`, reproducing the ticket's own matrix:

| Request | Before | After |
|---|---|---|
| `tests/5/basic-info` | 200 | 200 |
| `tests/999999/basic-info` | 404 | 404 |
| `tests/-1/basic-info` | 404 | 404 |
| `tests/notanumber/basic-info` | **500** | **404** |
| `tests/notanumber/sample-results` | **500** | **404** |
| `tests/abc/localization` | **500** | **404** |
| `POST tests/5/activate` | no flags | `{male, female, testId, active, orderable}` |

i18n: 5 keys added to `en.json`, **0 changed, 0 removed**, no other locale file touched, no duplicates — so no Transifex regression.

## Out of scope, flagged

A malformed `{ruleId}` can still 500 — it is a 36-char UUID and deliberately outside the numeric allow-list. Pre-existing and unrelated to this ticket.

[OGC-1153]: https://uwdigi.atlassian.net/browse/OGC-1153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mozzy11:
Superseded — reopening from my fork's develop branch per the project's fork workflow. Same commits, no content change.

PR #3979, opened 2026-08-04
Add new Programs workflow while preserving legacy Program 

HerbertYiga:
…ry route

# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Screenshots

## Related Issue

## Other


mozzy11:
Decision: Comment

Summary:
- Cloned the head, deployed it on an isolated dev stack seeded from a copy of the dev database, and walked the Programs screen, the legacy page and all three order-entry pickers through the UI and the API. The list, filters, add flow, editor prefill, JSON validate and round-trip, code validation, the deactivate modal with its order count and reactivate all work as built, and Liquibase 105-107 run cleanly with sound guards and constraints.
- Five live failures needed fixing before this could ship, so I pushed them onto the branch rather than hand back a list: changing the Domain of a saved program swapped in an empty questionnaire draft, so Save persisted 0 items (Cytology went 11 items to 0 on the stack); saving a deactivated program from the editor silently reactivated it, because the payload carries no `active` and the deserialized entity defaults to Y; the startup autocreate re-saved every unedited bundled program from JSON on each boot, resetting domain and lifecycle and emptying `program_lab_unit` (that is why the junction was empty right after the backfill ran); the Visual Builder read only `valueString`, so Histopathology's 363 `valueCoding` options showed as "No options yet" and adding one option replaced all 363; and the deactivate toast said "reactivated".
- FR-6 was not in the slice: `/rest/user-programs` still offered deactivated programs and every domain's programs, so the modal's promise that a deactivated program stops appearing was not true. Added the active filter and an optional `?domain=` filter on the same endpoint, with each order-entry step passing its domain.
- One upgrade impact worth knowing: 105 backfills every program to CLINICAL, so straight after an upgrade the Vector and Environmental pickers offer nothing. I classified the shipped Vector Field Survey as VECTOR (Liquibase 108, guarded and one-shot, plus the bundled definition for fresh installs); every other program stays Clinical until an admin reclassifies it, which is the FRS behaviour but worth a release note.
- Also on the branch: the FR-3 domain-change confirmation, multi lab-unit select with Select all, a status tag with Deactivate/Reactivate in the editor, audit history for programs, shared `label.domain.*` labels via `/rest/domains`, `/MasterListsPage/program` serving the new screen (the legacy page, its duplicate dashboard tile and its Cypress spec are gone, `programV2` redirects), plus backend and frontend tests and a Playwright spec.

Backward compatibility was the thing I most wanted to be sure of, so it was tested explicitly rather than reasoned about:
- All nine existing programs open in the new editor with item counts, types and options matching the API baseline exactly.
- Histopathology saved unchanged round-trips byte-identically, including all 363 coded options and the resource-level fields; after the fix, adding an option appends a 364th and leaves the 363 verbatim.
- Answers: created a clinical order under Cytology, answered two questions, then edited that questionnaire in the new admin. The program UUID and all 11 linkIds were unchanged, the stored QuestionnaireResponse was untouched (same md5), and reopening the order still showed both answers under the updated label. Deactivating the program afterwards still leaves that order showing its program and answers.

Optional improvements:
- The Domain list filter is a single Dropdown rather than the MultiSelect the FRS describes, and the table columns are not sortable yet.
- FR-7's post-upgrade banner (telling admins to reclassify programs defaulted to Clinical) is not implemented; the release note above covers the same ground for now.

Verification: PR head walked 22/22 inventory rows, 17 passed and 5 failed and are fixed above; the pushed changes were re-walked on the redeployed stack through the UI and the API, covering the read path for all nine programs, the coded-option round trip, the answers path end to end, all three domain pickers and the lifecycle. Backend program tests, the full frontend unit suite and the full backend suite green locally.

mozzy11:
Follow-up: an independent pass over the branch found one thing that mattered and four smaller ones, all now fixed and pushed, and `develop` is merged in so the branch is no longer conflicting.

**Authorization.** `ProgramController` carried no role guard, so any authenticated user could `GET /rest/program-list` or `POST /rest/program` — deactivating a program lab-wide, re-domaining it or overwriting its questionnaire. Every neighbouring Test Management controller is `@PreAuthorize("hasRole('ADMIN')")`, including the `/rest/lab-units-management` endpoint this very screen already calls. The guard is per method rather than per class, because order entry reads `/rest/program/{id}` and `/rest/program/{id}/questionnaire` for every user and a class-level annotation would have locked non-admins out of ordering. A test asserts that split so it cannot be quietly widened or removed.

Also fixed:
- Deleting every answer option from a choice question was silently reverted on save — the merge only rewrote `answerOption` when the question still had options, so the baseline's came back. Covered by a test that fails against the old code.
- The question-type dropdown and the preview placeholder rendered raw English; both now go through React Intl.
- Modify Order rendered the shared program picker without the domain prop, so it still offered every domain. FR-6 now applies on both order paths.
- The lab-unit multiselect honoured its selection only at mount, so an edit made before `/rest/lab-units-management` answered dropped the units it had not yet seen. It now remounts when the list arrives and preserves ids the endpoint does not offer.
- `orderCount` was asserted only as zero for an unused program, which a permanently broken query also returns; it now checks a seeded program with an order filed under it as well.

Merge conflict was `liquibase/3.5.x.x/base.xml`, both sides appending; resolved keeping both, develop's block first. Changeset numbers 105-108 do not collide with develop. Verified: all includes exist and are unique, XML well-formed, `test-compile` clean, program backend tests green (`ProgramSaveContractTest` 11/11), frontend tests green, spotless clean cold in a Linux container.

Left as optional and not fixed: a questionnaire containing FHIR `group` or `display` items is rejected by the validator in both modes, so such a program cannot be saved at all. No shipped questionnaire uses those types, but the editor this PR replaces did a bare `JSON.parse`, so a site that pasted one through it would be locked out of that program. Worth a follow-up that passes unsupported item types through untouched.

PR #4008, opened 2026-08-07
issue-3743: add Clear action for Active Alerts and improve Acknowledge functionality

COLLINS-WAGIMA256:
@Agaba-derrick @mozzy11 @reagan-meant , this Addresses **item 2** of #3743 where Active Alerts could be acknowledged but never removed. (#3763 covers item 1, the merged #3811 covered item 3, so this should not auto-close the issue.)

**Background**
The Active Alerts table already had an **Acknowledge** action, but Acknowledge only marks an alert as seen so the alert *stays* in the list (and the Acknowledge button then disappears, since it only applies while the alert is `OPEN`). There was no way to actually take a handled alert off the list, so alerts accumulated indefinitely.

**What this adds**
- A **Clear** action on each Active Alerts row that **resolves** the alert and removes it from the list thus complementing Acknowledge rather than replacing it:
  - **Acknowledge** → "I've seen this" (alert remains active/acknowledged).
  - **Clear** → "this is handled" (alert is resolved and leaves the list; it stays gone after refresh, since resolved alerts are now filtered out of the active list).
- Alert actions (both acknowledge and resolve) now use the **logged-in user** (`userSessionDetails.userId`) instead of a hardcoded `1`, so they're attributed to the actual operator rather than always admin. (Attribution correctness is not a crash fix; acknowledge/resolve already work on real alerts.)
- A Playwright E2E (`cold-storage-alert-clear.spec.ts`) that seeds an open alert, clicks **Clear**, and asserts it leaves the list and stays gone after reload as also seen in the screen recording below

Screen recording of the fix attached.

https://github.com/user-attachments/assets/3605a253-be20-4cf6-93f1-910411502869

Coordination: @mherman22's draft #3828 also touches item 2; per his note he'll rebase #3828 to drop what this covers. Happy to adjust to whatever lands first.

### DIFF
--- a/frontend/playwright/helpers/seed-freezer-alert.ts
+++ b/frontend/playwright/helpers/seed-freezer-alert.ts
@@ -1,0 +1,66 @@
+import { execFileSync } from "child_process";
+import { resolveDbContainer } from "./db-container";
+
+const SCHEMA = "clinlims";
+
+function psql(sql: string): void {
+  const container = resolveDbContainer();
+  execFileSync(
+    "docker",
+    [
+      "exec",
+      "-i",
+      container,
+      "psql",
+      "-U",
+      "clinlims",
+      "-d",
+      "clinlims",
+      "-c",
+      sql,
+    ],
+    { encoding: "utf8" },
+  );
+}
+
+function esc(value: string): string {
+  return `'${value.replace(/'/g, "''")}'`;
+}
+
+/**
+ * Seed a freezer with one OPEN critical alert directly in the database.
+ * Alerts have no create API (they are raised by the monitoring service), so the
+ * rows are inserted via docker exec psql against the compose DB container.
+ * The alert references the freezer by name (subquery) rather than capturing an
+ * id from psql output, and sets last_updated because the alert entity uses it
+ * as an optimistic-lock @Version (a NULL there makes resolve/acknowledge fail).
+ * Idempotent: any prior row with the same name is removed first.
+ */
+export function seedFreezerWithOpenAlert(name: string): void {
+  cleanupFreezerAlert(name);
+  psql(
+    `INSERT INTO ${SCHEMA}.freezer ` +
+      `(id, name, protocol, slave_id, temperature_register, target_temperature, warning_threshold, critical_threshold, active) ` +
+      `VALUES ((SELECT COALESCE(MAX(id), 0) + 1 FROM ${SCHEMA}.freezer), ` +
+      `${esc(name)}, 'RTU', 1, 0, -20, -15, -10, true);`,
+  );
+  psql(
+    `INSERT INTO ${SCHEMA}.alert ` +
+      `(id, alert_type, alert_entity_type, alert_entity_id, severity, status, start_time, last_updated, message, duplicate_count) ` +
+      `VALUES ((SELECT COALESCE(MAX(id), 0) + 1 FROM ${SCHEMA}.alert), ` +
+      `'FREEZER_TEMPERATURE', 'Freezer', (SELECT id FROM ${SCHEMA}.freezer WHERE name = ${esc(name)}), ` +
+      `'CRITICAL', 'OPEN', now(), now(), 'E2E cold-storage alert', 0);`,
+  );
+}
+
+export function cleanupFreezerAlert(name: string): void {
+  try {
+    psql(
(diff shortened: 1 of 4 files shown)


mherman22 (commented):
Reviewed — logic is sound (reuses the existing resolve endpoint, no backend changes needed) and the E2E correctly checks persistence across reload. A few one-line notes:

- `FreezerMonitoringDashboard.jsx`: `userId = userSessionDetails?.userId` isn't null-checked before `acknowledgeAlert`/`resolveAlert` — if this fires before session details load, it sends `userId: undefined`.
- `FreezerMonitoringDashboard.jsx`: "Clear" button label is a hardcoded string, not run through `intl.formatMessage`/`FormattedMessage` (matches the pre-existing "Acknowledge" button's style, so not a new regression, but worth fixing together per the i18n convention).
- `seed-freezer-alert.ts`: `(SELECT COALESCE(MAX(id),0)+1 ...)` for both the freezer and alert rows is racy if E2E specs ever run concurrently against the same DB two seeds could compute the same next id and collide.

Non-blocking overall. Heads up: #3828 also touched the Active Alerts remove-from-list behavior. I'm dropping that from #3828 now so this PR is the sole source of truth for it.

frontend/src/components/coldStorage/FreezerMonitoringDashboard.jsx
```diff
@@ -306,14 +310,19 @@ function FreezerMonitoringDashboard({ intl }) {
     async (alertId, action) => {
       setActionInFlight(alertId);
       try {
+        const userId = userSessionDetails?.userId;
```
mherman22: Not null-checked before use below if this fires before session details load, `userId` is `undefined` and gets sent to acknowledge/resolve.

frontend/src/components/coldStorage/FreezerMonitoringDashboard.jsx
```diff
@@ -830,6 +839,22 @@ function FreezerMonitoringDashboard({ intl }) {
...
+                                                      alert.id
+                                                    }
+                                                    onClick={(e) => {
+                                                      e.stopPropagation();
+                                                      handleResolveAlert(
+                                                        alert.id,
+                                                      );
+                                                    }}
+                                                  >
+                                                    Clear
```
mherman22: Hardcoded string, not run through intl.formatMessage/FormattedMessage (matches the pre-existing "Acknowledge" button's style, so not a new regression, but worth fixing together).

frontend/playwright/helpers/seed-freezer-alert.ts
```diff
@@ -0,0 +1,66 @@
...
+ * id from psql output, and sets last_updated because the alert entity uses it
+ * as an optimistic-lock @Version (a NULL there makes resolve/acknowledge fail).
+ * Idempotent: any prior row with the same name is removed first.
+ */
+export function seedFreezerWithOpenAlert(name: string): void {
+  cleanupFreezerAlert(name);
+  psql(
+    `INSERT INTO ${SCHEMA}.freezer ` +
+      `(id, name, protocol, slave_id, temperature_register, target_temperature, warning_threshold, critical_threshold, active) ` +
+      `VALUES ((SELECT COALESCE(MAX(id), 0) + 1 FROM ${SCHEMA}.freezer), ` +
```
mherman22, line 44: COALESCE(MAX(id),0)+1 here and on the alert insert below is racy if specs ever run concurrently against the same DB two seeds could compute the same next id and collide.

mherman22 (commented):
Inline notes to go with my earlier summary comment.

COLLINS-WAGIMA256:
Thanks for the thorough review, @mherman22,

Addressed the notes as follows:
1. **userId null-check** — added a guard in `handleAlertAction`: 
if `userSessionDetails?.userId` is missing it notifies and returns instead of sending `undefined`.

2. **i18n** — routed both **Clear** and **Acknowledge** through `intl.formatMessage` with new `en.json` keys (`coldstorage.alert.clear` / `coldstorage.alert.acknowledge`).

3. **Seed `MAX(id)+1` race** was left as-is: the Playwright config runs `workers: 1` everywhere and CI shards each own a separate stack/DB, so two seeds never hit the same DB concurrently. Happy to switch to a random-offset id, if you prefer

PR #4010, opened 2026-08-07
Replaced path-prefix sniffing with proper content negotiation  

### LINKED ISSUE #3965: Replace path-prefix sniffing with proper content negotiation
In a couple of places in the security subsystems of OE, we're using the URL path to determine what kind of response to send. This is a violation of REST principles where the content returned to the client should be negotiated with the client based on the `Accept` header. The idea is to replace [this block](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/6e7a225ed2a0ad7798fe4b876c63432883b63291/src/main/java/org/openelisglobal/interceptor/ModuleAuthenticationInterceptor.java#L60) and [this similar block](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/6e7a225ed2a0ad7798fe4b876c63432883b63291/src/main/java/org/openelisglobal/security/SecurityConfig.java#L440) and with something that looks at the request's `Accept` header. Something like:

```java
MediaTypeRequestMatcher htmlMatcher = new MediaTypeRequestMatcher(
        new HeaderContentNegotiationStrategy(), MediaType.TEXT_HTML);
htmlMatcher.setIgnoredMediaTypes(Set.of(MediaType.ALL));   // treat */* as "not HTML"

if (htmlMatcher.matches(request)) {
    response.sendRedirect(request.getContextPath() + "/Home?access=denied");
} else {
    // JSON 403
}
```

This should make it so requests from the _browser_ (or a client acting as a browser) return a redirect, which can be processed correctly, but requests from the Fetch API return a JSON response.

Bonus addenda:

* It would be worthwhile hunting down spots where this can be expected to return the JSON error and ensuring that the JS processing those reacts correctly to the JSON responses
* It would be better if the fetch / AJAX requests directly set the `Accept:` header to `application/json`, so that the client was explicitly requesting the format it wants. We could then evolve this to support multiple different content types.

Ariho-Seth:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
Initially, in the security subsystems of OpenElis, we've been depending on URL paths to determine whether the request is an API request or a browser request, thus determining what type of response to send.
This PR replaces the path-prefix sniffing with proper content negotiation by checking the request's `ACCEPT` header.

## Related Issue

[Replace path-prefix sniffing with proper content negotiation](https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3965)

closes #3965


Ariho-Seth:
Hello @bahati308, may you please review this PR and help me out on why the Frontend unit tests are failing?

pmanko:
Hi @Ariho-Seth — the `02 - Frontend` check was red on `IndexCrossTests.test.jsx` (2 failures). I reproduced it locally and, since maintainer edits are enabled on this PR, pushed a fix directly to the branch (d8329e97e). Two separate things:

### 1. Why CI was failing

`Index.jsx` now guards with `if (!response.ok)`. That test stubs `global.fetch` with a bare object:

```js
global.fetch = vi.fn(() =>
  Promise.resolve({ json: () => Promise.resolve(crossTestOrder()) }),
);
```

No `ok` property → `!response.ok` is `true` → a *successful* e-order lookup took the error path, `processLabOrderSuccess` never ran, and the awaiting-specimen chooser never rendered. Both tests then timed out waiting for it.

Your `response.ok` check is correct here — the stub was just an incomplete `Response` model, so I gave it `ok: true, status: 200`.

### 2. A real bug in the new error path

```js
return response.json()
  .then((errorJson) => {
    throw new Error(errorJson.message || `HTTP error! status: ${response.status}`);
  })
  .catch(() => {
    throw new Error(`HTTP error! status: ${response.status}`);
  });
```

The `.catch()` is chained onto the `.then()` that throws, so it catches that very `Error` and overwrites it. `errorJson.message` can never reach the caller — the branch is dead code, and every failure surfaces as the bare status. Changed to parse first, then throw:

```js
return response.json()
  .catch(() => null)
  .then((errorJson) => {
    throw new Error(errorJson?.message || `HTTP error! status: ${response.status}`);
  });
```

Worth noting: the other three frontend files in this PR (`CsvImportPreview.jsx`, `NceDashboard.jsx`, `StorageLocationModal.jsx`) use the `async/await` + `try/catch` shape, which preserves the message correctly. Only `Index.jsx` had the promise-chain variant.

### Verification

Added a regression test for the error path and confirmed it fails without the `Index.jsx` fix. Full frontend suite locally: **178 files, 1246 passed, 13 skipped**.

No changes to the backend content-negotiation work — that all looks unaffected.

src/main/java/org/openelisglobal/interceptor/ModuleAuthenticationInterceptor.java
```diff
@@ -108,7 +116,12 @@ private boolean hasPermissionForUrl(HttpServletRequest request, boolean useParam
             // auto-allowed for any authenticated user. Admin-only controllers are
             // protected via @PreAuthorize("hasRole('ADMIN')") (added in PR #2794).
             // Full per-role module mappings are a future enhancement.
-            if (isRestFullPath(path)) {
+
+            MediaTypeRequestMatcher requestMatcher = new MediaTypeRequestMatcher(new HeaderContentNegotiationStrategy(),
+                    MediaType.TEXT_HTML);
+            requestMatcher.setIgnoredMediaTypes(Set.of(MediaType.ALL));
+
+            if (requestMatcher.matches(request)) {
```
pmanko, line 124: Requests with no module entry are now allowed when they ask for HTML, while normal JSON requests are denied. That reverses the intended rule and can both expose pages and break API calls. Please keep the permission rule independent of response format, then test unmapped HTML, JSON, and `*/*` requests.

pmanko (changes requested):
Reviewed commit: `d8329e97e212dc7a504c075a2055c1771c73167c`

Must fix before approval: correct the permission fallback noted inline. The existing Copilot note about unused redirect code can be cleaned up in the same update.

Verification: add mapped and unmapped request tests for HTML, JSON, and wildcard `Accept` headers. Backend checks and the authoritative E2E checkpoint must pass.

PR #4011, opened 2026-08-07
fix: rename Validation result-table column header from "Save" to "Val…

HerbertYiga:
On the validation results page (opened after searching a lab number on Ready For Validation), the column holding the accept checkbox was labeled "Save", which was misleading — ticking a row is what feeds the Validate action at the bottom of the page. Point the column at the existing translatable key `label.button.validate` so it reads "Validate" and stays Transifex-managed.


Agaba-derrick (approved):
LGTM

PR #4015, opened 2026-08-09
feat(inventory): add a readable code to an inventory item

mherman22:
First of five. The rest sit on top of this one, so it needs to go in first.

1. **Give inventory items a readable code** — you are here
2. #4016 Put inventory lots in the same storage system samples already use
3. #4017 Add an inventory Reports tab

### Why

Right now the only identifier an inventory item has is its auto-increment primary key. That is what every REST path exposes (`/{id}`, `/{id}/stock`, `/{id}/deactivate` and so on), and it differs from one instance to the next, so it is useless for referring to an item across sites or in imported data. There is nothing stable and human-readable to key on.

Nothing user-visible is broken by that today, and this PR does not claim otherwise: lot numbers are free text the user types, the catalog table does not display the id, and there is no inventory report export on the backend yet. What this adds is the identifier the later PRs in the list above need.

### What this does

Adds a `code` column to `inventory_item`: uppercase, underscore-separated, generated from the item's name when you don't supply one (`Blood Culture Bottle` becomes `BLOOD_CULTURE_BOTTLE`). Long names are truncated to fit and collisions get a numeric suffix. You can type your own code instead, and it gets normalised the same way. A code that is already taken comes back as a 400 with a translatable message key rather than a database error. That check runs before the insert, so two requests racing on the same code can both pass it; the loser hits the unique constraint and gets a 500.

The primary key is left alone. It stays the surrogate sequence value it has always been, which is how the rest of the schema works — most entities here use a numeric id that Hibernate happens to map as a string, per #3112.

An earlier version of this PR made the code itself the primary key. That turned out to be the wrong call: it forced a destructive migration and rippled type changes through half the module. Herman pushed back on it and he was right. This version does the same job as a plain unique column.

### The migration

`070-inventory-item-code.xml` adds the column, backfills it from existing item names, then adds `NOT NULL` and `UNIQUE`. Nothing is deleted, no foreign keys are touched, and it is safe to run on an instance that already holds inventory data. A backfilled name that collides gets `<CODE>_<id>` appended; a name that normalises to nothing becomes `ITEM_<id>`. Anything still colliding after five dedup passes falls back to the first `ITEM_<id>` variant that no other row holds, checked against the table rather than assumed free, so every row ends with a distinct code.

The include is anchored after `048-test-reagent-link.xml` rather than appended at the end of `base.xml`, because the end of that file is where every branch collides and this one has no dependency on anything later.

### Testing

Unit coverage for generation, collisions, normalisation, truncation and duplicate rejection. The REST suite covering the 400 body and the code surviving an update is `InventoryItemRestControllerTest`, named so surefire runs it in CI; a `*IT` name would not have been executed at all, for the reason below.

One thing worth knowing if you run the suites yourself: `InventoryStorageLocationServiceIT` fails, and it fails on `develop` too. It loads a fixture that truncates its table and inserts ids 1000/1001, then asserts on `get(1L)`. CI never caught it because there is no `maven-failsafe-plugin` in `pom.xml`, so no `*IT` class runs in CI at all. #4016 deletes that test along with the dead code it covers, but the wider gap is worth its own ticket.


src/main/java/org/openelisglobal/inventory/valueholder/InventoryItem.java
```diff
@@ -39,6 +39,19 @@ public class InventoryItem extends BaseObject<Long> {
...
+     * Human-readable, stable identifier for the item (UPPER_SNAKE, e.g.
+     * {@code BLOOD_CULTURE_BOTTLE}). Generated from the name by
+     * {@code InventoryItemServiceImpl.insert()} when not supplied. Distinct from
+     * {@link #id}, which stays an opaque surrogate key: this is what users see in
+     * the catalog and what derived lot numbers are built from, so it must be
+     * readable and stable, not a per-instance sequence value.
+     */
+    @Column(name = "code", nullable = false, unique = true, length = 64)
+    @NotNull
+    @Size(min = 1, max = 64)
```
mherman22: Bug: every Add and Edit in the inventory catalog will 400 after this. `code` is `@NotNull`, but `InventoryItemForm.jsx` never sends it, and `@Valid` on `create()`/`update()` rejects the payload before `resolveCode()` can fill it in. Drop `@NotNull`/`@Size` and let `nullable = false` plus `resolveCode()` enforce it.

src/main/java/org/openelisglobal/inventory/service/InventoryItemServiceImpl.java
```diff
@@ -30,6 +36,42 @@ protected InventoryItemDAO getBaseObjectDAO() {
...
+        return super.insert(item);
+    }
+
+    private String resolveCode(InventoryItem item) {
+        String supplied = item.getCode();
+        String code = (supplied == null || supplied.trim().isEmpty())
+                ? CodeGenerator.generateFromName(item.getName(), CODE_MAX_LENGTH, "ITEM", this::codeExists)
+                : CodeGenerator.normalize(supplied, CODE_MAX_LENGTH);
+        if (codeExists(code)) {
+            throw new LocalizedValidationException("inventory.item.error.duplicateCode",
```
mherman22, line 52: Nit: this error reaches the user as the raw key `inventory.item.error.duplicateCode`. Neither it nor `common.codeGenerator.error.invalidCode` (`CodeGenerator.java:49`) is in `frontend/src/languages/en.json` at this head or on `develop`. Add both here, since this is the PR that introduces the exception.

src/main/java/org/openelisglobal/common/util/CodeGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+        return candidate;
+    }
+
+    /**
+     * Normalizes a user-supplied explicit code: uppercased, stripped to
+     * {@code [A-Z0-9_]}, truncated to {@code maxLength}. Throws if nothing usable
+     * remains.
+     */
+    public static String normalize(String code, int maxLength) {
+        String normalized = code.trim().toUpperCase().replaceAll("[^A-Z0-9_]", "");
```
mherman22: Nit: typing a code by hand gives a different answer than leaving it blank. `normalize()` deletes separators while `generateFromName()` collapses them to `_`, so `Blood Culture Bottle` becomes `BLOODCULTUREBOTTLE` typed and `BLOOD_CULTURE_BOTTLE` generated. Collapse `[^A-Z0-9_]+` to `_` here so the two agree.

src/main/resources/liquibase/3.5.x.x/070-inventory-item-code.xml
```diff
@@ -0,0 +1,69 @@
...
+            UPDATE clinlims.inventory_item
+            SET code = LEFT(TRIM(BOTH '_' FROM REGEXP_REPLACE(UPPER(name), '[^A-Z0-9]+', '_', 'g')), 64)
+            WHERE code IS NULL;
+
+            UPDATE clinlims.inventory_item
+            SET code = 'ITEM_' || id
+            WHERE code IS NULL OR code = '';
+
+            UPDATE clinlims.inventory_item outer_item
+            SET code = LEFT(outer_item.code, 58) || '_' || outer_item.id
```
mherman22: Nit: this can abort the upgrade with `value too long for type character varying(64)`. `LEFT(code, 58) || '_' || id` reaches 65 characters once `id` has six digits, which is the failure the fallback exists to prevent. Wrap the whole expression in `LEFT(..., 64)`.

mherman22 (commented):
Adds a readable `code` column to `inventory_item`. The migration is additive and the generation logic is sound, but the entity constraint makes the auto-generation path unreachable, so the catalog form cannot save an item at all. Worth running `InventoryItemRestControllerIT` locally before merging, since there is no `maven-failsafe-plugin` and `*IT` classes never run in CI.

src/test/resources/testdata/inventory-test-data.xml
```diff
@@ -33,23 +33,23 @@
     <inventory_lot id="1000"
         fhir_uuid="550e8400-e29b-41d4-a716-446655440011"
         inventory_item_id="1000" storage_location_id="1000"
-        lot_number="LOT-2025-001" expiration_date="2025-12-31 00:00:00"
+        lot_number="LOT-2025-001" expiration_date="2099-12-31 00:00:00"
```
mherman22: What moved these three lot expiry dates from 2025/2026 out to 2099/2100? Nothing about the code column needs it and the description doesn't mention it.

I went looking for what it could affect and the answer is nothing that runs: `InventoryItemServiceTest` is the only surefire-run consumer of this fixture, and neither `getTotalCurrentQuantity` (filters `status IN ('ACTIVE','IN_USE')`) nor `getAvailableLotsByItemFEFO` (filters status, QC status and quantity, and only *orders* by expiry) reads the expiration date at all.

Which is really the point. This plus the `1L` to `1000L` rewrite through `InventoryLotServiceIT` and `InventoryManagementServiceIT` is about 79 lines of the 621 that a reviewer has to account for, unmentioned, and inert in CI either way. If those two suites carry the same id mismatch you describe for `InventoryStorageLocationServiceIT`, that is a separate repair and reads better as its own commit, particularly since #4016 is where you put the third one.

src/test/java/org/openelisglobal/inventory/controller/rest/InventoryItemRestControllerIT.java
```diff
@@ -0,0 +1,145 @@
...
+ * column alongside its surrogate primary key. Covers auto-generation from the
+ * name, explicit codes, duplicate rejection with a translatable 400, and the
+ * code surviving an update untouched.
+ *
+ * <p>
+ * Never truncates {@code inventory_item} (shared with other suites via
+ * {@code inventory-test-data.xml} fixtures) — only inserts/deletes rows scoped
+ * to a unique {@code ITTEST_} code prefix.
+ */
+public class InventoryItemRestControllerIT extends BaseWebContextSensitiveTest {
```
mherman22: This file can't run in CI, so the 400 body and the code-surviving-an-update behaviour it covers stay unverified on every future change.

There is no `maven-failsafe-plugin` in `pom.xml` and no parent pom to inherit one from, and surefire's default includes are `Test*`, `*Test`, `*Tests`, `*TestCase`, none of which match `*IT`. So `mvn clean install` in `backend.yml` compiles this class and runs none of it.

You call the gap out in the description for `InventoryStorageLocationServiceIT`, but it swallows this PR's own integration coverage too, and the two new catch blocks in the controller end up with no executing test of any kind. Binding failsafe here is the cheapest thing that makes the description's testing claim true. If that feels like scope creep, a MockMvc `*Test` covering just the duplicate-code 400 would at least get that path into CI.

src/main/java/org/openelisglobal/common/util/CodeGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Normalizes a user-supplied explicit code: uppercased, stripped to
+     * {@code [A-Z0-9_]}, truncated to {@code maxLength}. Throws if nothing usable
+     * remains.
+     */
+    public static String normalize(String code, int maxLength) {
+        String normalized = toCode(code);
+        if (normalized.isEmpty()) {
+            throw new LocalizedValidationException("common.codeGenerator.error.invalidCode",
+                    "Code must contain at least one letter, digit, or underscore");
```
mherman22: An all-underscore code lands in exactly this branch, so the advice here is wrong for the input most likely to trigger it: `toCode("___")` collapses to `_` and then strips to empty. The `en.json` string you added has it right ("at least one letter or number"); this is the fallback that reaches the logs and the response `message` when nothing localizes the errorCode.

```suggestion
                    "Code must contain at least one letter or digit");
```

src/main/java/org/openelisglobal/common/util/CodeGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.function.Predicate;
+import org.openelisglobal.common.exception.LocalizedValidationException;
+
+/**
+ * Shared UPPER_SNAKE code generation used by admin-managed reference lists that
+ * let a user optionally type an explicit code, auto-generating one from a
+ * display name otherwise (collision-suffixed {@code _2}, {@code _3}, ...).
+ * First introduced for {@code inventory_item_type.code} (OGC-658 Part A),
+ * reused for {@code inventory_item.code} (OGC-658 Part C).
+ */
```
mherman22, line 10: Two lines of description is the bar for javadoc here, and the two lines that push this over are the two that describe the PR rather than the code. Which ticket part introduced it and which reused it is what `git log` carries, and it goes stale the moment a third caller turns up.

```suggestion
/**
 * Shared UPPER_SNAKE code generation for reference lists where a user may type
 * an explicit code, else one is derived from a display name (suffixed _2, _3).
 */
```

Same shape in the other new files: `LocalizedValidationException` at 6 lines, `InventoryItemRestControllerIT` at 8 (though its second paragraph, the one warning not to truncate `inventory_item`, earns its keep), and `InventoryItemServiceCodeGenerationTest`, whose "not the primary key ... so nothing here renames a key or cascades" is answering the review argument about the earlier version rather than telling the next reader anything about the test.

PR #4016, opened 2026-08-09
feat(inventory): put inventory lots in the same storage system samples already use

mherman22:
Second of seven, on top of #4015. Merge that one first.

1. #4015 Give inventory items a readable code
2. **Put inventory lots in the same storage system samples already use** — you are here
3. #4017 Add an inventory Reports tab
7. #4033 Bring inventory lot storage to parity with samples

### Why

Inventory had its own storage stack: an `inventory_storage_location` table, a DAO, a service, a controller, and a modal that let you pick a location from a flat dropdown. It was a skeleton. No hierarchy, no capacity, no occupancy, no barcode, and no connection to the Storage Management module that samples already run through. Two parallel answers to "where is this thing", one of them barely working.

This deletes the skeleton and puts lots on the real one.

### What this does

`sample_storage_assignment` and `sample_storage_movement` get an `occupant_type` discriminator (`SAMPLE_ITEM` or `INVENTORY_LOT`) and a nullable `inventory_lot_id` next to the existing `sample_item_id`, with a check constraint so exactly one occupant id is set per row. Existing rows backfill to `SAMPLE_ITEM`.

The tables keep their `sample_storage_*` names on purpose. Renaming them would mean touching foreign keys, indexes, existing data and service code for a feature that is already shipped and working, to gain nothing but a tidier name.

A lot's location now lives in the assignment table instead of on the lot row, and the lot list loads every lot's assignment in one `IN` query instead of one request per lot (the hierarchical path is still resolved per assignment, served from Hibernate's first-level cache when lots share a location).

On the frontend, `LotEntryModal` drops its dropdown and its inline create-a-location modal and uses the shared `LocationPickerModal` — the same hierarchical picker with search, capacity and position that samples use. Creating a lot lets you pick a location before the lot exists, then applies the assignment once it is saved. Receive and assign are two independent server writes, so if the assignment is rejected the modal keeps the new lot's id, locks the lot fields and retries only the assignment on the next Save, rather than receiving the same physical stock twice; closing the modal in that state refreshes the caller's list so the committed lot is visible.

One change in this PR is unrelated to storage and is here only because the same dashboard file was already open: the Item Type filter on the inventory dashboard was a hard-coded list of three types and now fetches the real list from the pre-existing `/rest/inventory/items/types`, labelled through new `inventory.itemType.*` keys with a fallback to the raw code for site-defined types. If that fetch fails the filter degrades to All-only with a console error and no user-visible signal.

The dashboard table body also stopped pairing Carbon's rendered rows to the fetched lots by array index and now resolves each row's lot by id. The index join predates this PR, but this PR adds the first row action that performs a storage write through it, so sorting a column would have assigned the location to the wrong lot and reported success.

`LocationPickerModal.css` is a global stylesheet, so its layout rules are scoped under a modal-specific ancestor; `SearchField` is shared with the non-modal pickers on Add Order, Order Label and the Notebook form, which keep their in-flow results list.

### One thing to look at

`InventoryLot.location` uses the `transient` **keyword**, not `@jakarta.persistence.Transient`. Both exclude the field from persistence, but `Hibernate5JakartaModule`'s Jackson introspector treats the annotation as an implicit `@JsonIgnore`, which silently dropped the field from every JSON response even when it was populated. That cost a while to find, hence the comment on the field.

### Heads-up: this touches shared sample-storage code

Two changes here are not confined to inventory, and a reviewer looking for inventory changes will find them unexpectedly.

**`SearchField.jsx` is the shared location picker's search box**, used by sample storage as well. It had a real bug: Space was treated as a selection key alongside Enter, so once any result was showing, pressing the space bar picked the highlighted row instead of typing a space. Any multi-word location name was unsearchable — "cold room" would select whatever matched "cold". That has been affecting sample storage all along, not just inventory. Enter now selects, Space types.

The same component gained the feedback it was missing: a "Searching…" loader (there was 300ms of debounce plus a round trip with no indication anything was happening), a "no locations match" message for genuinely empty results, an error message when the request fails, and the 2-character hint now shows while you are below the threshold rather than only in the placeholder that disappears once you start typing. The results list also gets a max-height so a broad query cannot overflow the modal.

None of that is inventory-specific. It is here rather than in its own PR because the inventory receive flow is what surfaced it, but it changes sample storage behaviour the moment this merges — worth a look from whoever owns that flow.

**Lot number generation** also lives here: leave the lot number blank on the receive form and the server fills in `<item code>_<yyyyMMdd>`, suffixing `_2` if that number already exists for the item today. A lot number you type is stored exactly as entered, since manufacturer lot numbers must not be normalised. The form was already written to allow a blank value, so without this the save failed on a not-null violation.

### Testing

New coverage that CI actually runs: lot assignment and movement through the storage service and the REST layer, the lot list resolving every lot's location in one bulk call rather than per lot, lot-number generation, the `LotEntryModal` partial-save retry path, the lot movement history in `LotDetailsPanel`, the sample-item metrics endpoint ignoring lot occupants, the `ON DELETE CASCADE` on the new lot FKs, and Space no longer selecting a search result in the shared picker.

This also removes `InventoryStorageLocationServiceIT`, which had been failing on `develop` without anyone noticing, because surefire's default includes do not match `*IT.java` and the project configures no failsafe plugin. For the same reason this PR's own database-backed tests are named `SampleStorageServiceInventoryLotIntegrationTest` and `InventoryLotRestControllerIntegrationTest`, so surefire runs them.

### Scope: what this does *not* do

This puts lots **onto** the storage system. It does not make Storage Management
aware of them — there is no lots listing, no lots view, and a location page will
not show a lot as an occupant. Occupancy queries count lots: the box-occupancy count filters on location
alone, and the per-coordinate lookup reports each occupant's type with its
sample item id or lot number. No shipped page renders that per-coordinate
data yet.

Two related gaps are also out of scope here and are handled in #4033: disposing a
lot does not release its storage slot, and there is no in-place edit of an
assignment's position or notes.


frontend/src/components/storage/LocationPicker/LocationPickerModal.jsx
```diff
@@ -18,22 +18,26 @@ import {
...
  * translates that into the appropriate REST call. The modal is
- * workflow-agnostic.
+ * workflow- and occupant-agnostic — `occupant` is display-only (identifier,
+ * type, status) and `occupantType` (e.g. "SAMPLE_ITEM" | "INVENTORY_LOT")
+ * is passed straight through so the caller can pick the right endpoint.
  */
 export default function LocationPickerModal({
   isOpen,
-  sample,
+  occupant,
```
mherman22, line 40: Renaming this prop from `sample` to `occupant` didn't reach the two sample-side callers, so the shared picker throws for samples as soon as it renders. `resultPage/SearchResultForm.jsx:1795` and `resultPage/unified/orderContextSections.tsx:367` both still pass `sample={{...}}`, which leaves `occupant` undefined and line 128 raises `TypeError: Cannot read properties of undefined (reading 'label')`.

`orderContextSections.test.tsx` ("the storage button opens the shipped LocationPickerModal inline") fails on this branch for exactly that reason and passes on the parent commit, so it is reproducible without a browser. SearchResultForm is the worse of the two: it renders the picker unconditionally rather than behind `{open && ...}`, so an expanded results row crashes before anyone clicks Assign.

Updating both callers to the new shape is the fix I'd take (`occupant.label` wants `data.accessionNumber`, `occupant.type` the sample type, `occupant.status` the sample status) rather than accepting `sample` as a fallback, since a fallback just keeps two prop shapes alive.

src/main/java/org/openelisglobal/storage/service/SampleStorageServiceImpl.java
```diff
@@ -1495,4 +1503,383 @@ private SampleItem resolveSampleItem(String identifier) {
...
+            assignment.setLocationId(locationIdInt);
+            assignment.setLocationType(locationType);
+            if (effectiveCoordinate != null) {
+                assignment.setPositionCoordinate(effectiveCoordinate);
+            }
+            assignment.setAssignedDate(new Timestamp(System.currentTimeMillis()));
+            assignment.setNotes(notes);
+            assignment.setAssignedByUserId(resolveActingUserId(sysUserId));
+
+            Integer assignmentIdInt = sampleStorageAssignmentDAO.insert(assignment);
```
mherman22, line 1694: This row lands in `sample_storage_assignment`, and one sample-side consumer counts that table raw: `SampleStorageRestController.getSampleItems` on the `countOnly=true` branch does `totalSampleItems = sampleStorageAssignmentDAO.getAll().size()` (SampleStorageRestController.java:116-118), and `getAll()` inherits `BaseDAOImpl` with no occupant filter. Once a single lot is stored, the storage metrics count it as a sample item, and `totalSampleItems` stops agreeing with `active + disposed` (those two loops are guarded by `getSampleItemId() != null`, so they stay correct).

Nothing throws, so this surfaces as a quietly wrong number on the dashboard metrics rather than a failure. `SampleStorageRestControllerIntegrationTest.getSamples_ShouldReturnMetrics_WhenCountOnly` only asserts the field exists and is `>= 0`, so no test catches it either.

One-line fix at the call site: count only the rows where `getSampleItemId() != null`, the same guard the active/disposed loop right below it already uses.

frontend/src/components/inventory/LotEntryModal.jsx
```diff
@@ -165,25 +185,31 @@ const LotEntryModal = ({ open, onClose, onSave, lot = null }) => {
...
             ? formData.expirationDate.toISOString()
             : null,
           receiptDate: formData.receiptDate.toISOString(),
-          storageLocation: { id: formData.storageLocation.id },
           qcStatus: formData.qcStatus,
           status: formData.status,
           barcode: formData.barcode || null,
         });
+
+        if (pendingAssignment && savedLot?.id) {
```
mherman22: Creating a lot is now two writes with nothing between them. `receive` has already committed the lot by the time `assignLocation` runs, so when the assign fails (picked location went inactive, another user took the position, a dropped request) the catch shows `Error saving lot` and leaves the modal open with the form untouched. The obvious next move is to press Save again, which receives the same stock a second time, and the item's on-hand quantity is then double counted.

`validate()` makes a location mandatory for new lots, so this path is reachable in normal use. I'd treat a post-create assign failure as a partial success: call `onSave()`, close the modal, and say the lot was created but not assigned, so the operator finishes from the row's Assign action instead of re-saving. Doing the assignment inside `receiveInventory` server-side would be tidier still, but that is more than this PR needs.

src/main/java/org/openelisglobal/storage/service/SampleStorageServiceImpl.java
```diff
@@ -1495,4 +1503,383 @@ private SampleItem resolveSampleItem(String identifier) {
...
+            throw new LIMSRuntimeException("Invalid location type: " + locationType
+                    + ". Must be one of: 'room', 'device', 'shelf', 'rack', 'box'");
+        }
+    }
+
+    /**
+     * Validate minimum hierarchy depth (room + device per FR-033a), except for
+     * room-level assignments which are valid as a single level.
+     */
+    private void validateHierarchyDepth(Object locationEntity, String locationType) {
```
mherman22, line 1595: `validateHierarchyDepth`, `loadLocationEntityByType` and `effectiveCoordinateFor` are the occupant-agnostic versions of logic `assignSampleItemWithLocation` and `moveSampleItemWithLocation` still carry inline, and nothing routes the sample path through them. The two copies have already started disagreeing: the sample assign path still reports `Position %s is already occupied by another sample` when the occupant may now be a lot, and for a box move with a blank coordinate the sample path writes the box label to the assignment but `null` to the movement row (lines 1069-1078 vs 1155-1161), where your lot path writes `effectiveCoordinate` to both. Same user action, two different audit trails depending on occupant type.

Pointing the two sample methods at these three helpers is contained and would collapse both divergences. Not a merge blocker, but every later change to the hierarchy rules now has to be made twice.

frontend/src/languages/en.json
```diff
@@ -6205,10 +6219,10 @@
   "storage.picker.inlineCreate.heading": "Create new {level}",
   "storage.picker.inlineCreate.error.title": "Could not create location",
   "storage.picker.inlineCreate.error": "Failed to create location",
-  "storage.picker.sample.accession": "Accession",
-  "storage.picker.sample.heading": "Sample",
-  "storage.picker.sample.status": "Status",
-  "storage.picker.sample.type": "Type",
+  "storage.picker.occupant.identifier": "Identifier",
```
mherman22, line 6875: `LocationPickerPage.jsx` (lines 78, 89, 96, 103, 110) still reads `storage.picker.heading.moveSample` and `storage.picker.sample.{heading,accession,type,status}`, which this hunk removes. Each of those call sites supplies a `defaultMessage`, so English still renders, but the page logs a missing-message error on every open, and the page and the modal now word the same action differently: "Move Sample" vs "Move Item", "Sample" vs "Item", "Accession" vs "Identifier". Worth renaming the page's ids in the same pass.

src/main/resources/liquibase/3.5.x.x/071-generalize-storage-assignment-occupant.xml
```diff
@@ -0,0 +1,260 @@
...
+                <column name="storage_location_id" type="BIGINT"/>
+            </addColumn>
+        </rollback>
+    </changeSet>
+
+    <changeSet id="OGC-657-drop-inventory-storage-location-table" author="OGC">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="inventory_storage_location" schemaName="clinlims"/>
+        </preConditions>
+        <dropTable tableName="inventory_storage_location" schemaName="clinlims"/>
```
mherman22, line 183: `src/test/resources/inventory-management-demo-data.sql` still inserts into `inventory_storage_location` (line 18), calls `setval` on `inventory_storage_location_seq` (line 60), writes `inventory_lot.storage_location_id` (line 144) and selects from the table again in its verification block (line 407). Its own header documents the `docker exec -i openelisglobal-database psql ... < ...` command to run it, so once this changeset has applied that command fails on the first statement. Either update the seed alongside the schema or drop it here.

src/main/java/org/openelisglobal/inventory/service/InventoryManagementServiceImpl.java
```diff
@@ -103,6 +105,14 @@ public List<ConsumptionRecord> consumeInventoryFEFO(Long itemId, Double quantity
         return consumptionRecords;
     }
 
+    private String generateLotNumber(InventoryItem item) {
+        String datePart = new SimpleDateFormat("yyyyMMdd").format(new Timestamp(System.currentTimeMillis()));
+        Set<String> existingLotNumbers = inventoryLotService.getByInventoryItemId(item.getId()).stream()
```
mherman22, line 162: Was a duplicate generated lot number considered acceptable here? `lot_number` carries only the non-unique `idx_lot_number` (3.3.x.x/016-inventory-management-system.xml:318), so the read of existing numbers and the insert are not serialised: two receives for the same item on the same day, both with the lot number left blank, both see the same set and both write `CODE_20260914`. The `_2` suffixing reads as an intent that lot numbers be unique per item, and `InventoryLotDAOImpl.getByLotNumber` uses `setMaxResults(1)`, so a duplicate would make lookup by lot number pick one arbitrarily. A unique index on `(inventory_item_id, lot_number)` plus a retry would close it, if it matters.

PR #4017, opened 2026-08-09
feat(inventory): add a Reports tab with CSV, PDF and XLSX export

mherman22:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Adds the backend for the inventory Reports tab and wires the existing tab up to it. Six reports (stock levels, low stock, expiration forecast, lot traceability, transaction history, usage trends) are built into one shared table structure and written out as CSV, PDF or XLSX by one writer.

**Available Quantity.** The Low Stock report selects its rows on available quantity, and Stock Levels prints the same figure next to the raw total. A lot counts as available when it is ACTIVE or IN_USE, not expired, holds a positive quantity, and is QC PASSED or still QC PENDING: stock awaiting QC is on the shelf and should not trigger a reorder. This is deliberately wider than `InventoryLot.isAvailableForUse()`, which is PASSED-only because it gates consumption; that method and its callers are unchanged. QC FAILED or QUARANTINED lots, lots whose status is EXPIRED, CONSUMED, DISPOSED or QUARANTINED, and lots past their expiry date count toward Total Quantity only.

The Inventory Dashboard's low-stock tile does not use any of this yet: it is computed in the browser from a `minimumStockLevel` field that `InventoryItem` does not have, so it always reads 0. #4019 moves the tile to the backend and fixes the raw-sum query behind `getLowStockItems()`, at which point the tile and this report agree.

**Exports.** Quantity columns (available and total quantity, low-stock threshold, quantity change and quantity after, days until expiration, usage totals and counts) are typed as numbers in the XLSX, so SUM, COUNT and sorting work on them, while item codes and lot numbers stay text; the CSV carries them as plain numbers and prefixes any other cell whose first non-blank character is `=`, `+`, `-` or `@` with an apostrophe so a spreadsheet cannot execute it. Fractional quantities are formatted with `Locale.ROOT`. Reports print the item's readable code (from #4015) rather than its database id. Column headers stay in English on purpose: these are downloaded files, not rendered UI, and the reason is recorded in the service class.

**Validation.** Every 400 from the generate endpoint (unknown report type, unknown export format, missing or invalid dates) returns the same `{message, errorCode, params}` JSON body, with the report type validated once, in the service. The Reports form shows the date range only for the three report types that read it.

**Something to decide before this gets much use.** The six reports work, but as a set they are cut wrong: Low Stock duplicates a live decision surface, "usage trends" is really period consumption, lot traceability and transaction history are on-demand lookups that `LotDetailsPanel` already covers, and there is no Received report, so a stock card cannot be reconciled. FRS v1.9 section 4.7 re-specs this as four period reports plus a stock-card summary. That is a product change, deliberately not made here.

**Testing.** The endpoint test class was named `*IT`, which matched none of surefire's include patterns and no failsafe plugin is configured, so its tests compiled and never ran; it is now `InventoryReportRestControllerTest` (10 tests, parses the real XLSX and JSON error bodies). `InventoryReportServiceTest` (17) covers the availability predicate, sorting, date handling and locale; `InventoryReportWriterTest` (7) reads the CSV, the PDF text and the typed XLSX cells back. Verified at runtime on a `scripts/dev-stack` deployment of this head: QC-PENDING stock counted as available, XLSX cell types, JSON 400s, an expired-this-morning lot, and every export format from the UI.

## Screenshots

Not applicable: the visible change is a downloaded file. Sample rows from the verification run:

```
STOCK_LEVELS   VR3_NUM,Verify Numeric R3,REAGENT,,Unassigned,11.50,18.50,mL,Active
LOW_STOCK      VR2_B_FAILED,VR2 B Failed,REAGENT,,Unassigned,3,503,20,mL
EXPIRATION     VR3_NUM,Verify Numeric R3,REAGENT,VR3-EXPTODAY,Unassigned,2026-09-18,-1,EXPIRED,7,ACTIVE
```

## Related Issue

Part of the inventory stack (reviewed bottom-up). #4015 and #4016 are merged, so the open part of the stack starts from `develop`:

3. **#4017 Add a Reports tab with CSV, PDF and XLSX export** (this PR, based on `develop`)

## Other

Known and left alone on purpose: the browser saves downloads as `low-stock.csv_` because the Content-Disposition parser in `InventoryService.js` (already on `develop`) captures the closing quote; that is a one-line frontend fix for its own PR. The PDF writer uses iText's built-in Helvetica and drops non-Latin-1 glyphs, as every PDF exporter in the tree does today; embedding a Unicode font belongs in a PR that covers all of them.


src/main/java/org/openelisglobal/inventory/report/InventoryReportServiceImpl.java
```diff
@@ -0,0 +1,468 @@
...
+     * quantity, not the raw sum of every lot — a threshold check against total
+     * quantity would count EXPIRED/DISPOSED/QUARANTINED/QC-failed stock as if it
+     * were usable, which is exactly backwards for an alert meant to answer "what do
+     * we need to reorder." {@code InventoryItemService.getLowStockItems()} now does
+     * this same available-quantity check (previously it delegated to a native-SQL
+     * query with the raw-sum flaw — also the source behind the Inventory
+     * Dashboard's low-stock tile, fixed there too).
+     */
+    private ReportTable buildLowStockReport(InventoryReportRequest request) {
+        List<InventoryItem> items = inventoryItemService.getLowStockItems();
```
mherman22, line 138: This needs fixing before merge. At this commit `InventoryItemServiceImpl.getLowStockItems()` still returns `inventoryItemDAO.getLowStockItems()`, whose native query is `... HAVING COALESCE(SUM(l.current_quantity), 0) < i.low_stock_threshold`, summed over every lot with no status or QC filter. So the report displays an available-quantity column, but the choice of which items appear is still made on the raw sum.

If merged as-is, an item with 3 usable units plus 100 disposed ones against a threshold of 20 never appears in the Low Stock report, because 103 is not below 20 and the query drops it, so nobody reorders it. That is exactly the scenario `lowStock_displaysAvailableAndTotalQuantitySeparately` sets up; it passes only because the mock hands the item back regardless.

The javadoc here, and the "judged on lots that are actually available" line in the PR description, both describe a state that arrives two PRs later: the available-quantity rewrite of `getLowStockItems()` is in #4019, which is also where `InventoryItemServiceLowStockTest` (cited by the comment on line 151 of the service test) first exists. Cleanest fix is to bring that one method down into this PR so the report and its own comments agree. If it has to stay in #4019, then the javadoc and the description need to say the selection is still raw-sum, since right now they assert something the branch does not do.

src/main/java/org/openelisglobal/inventory/controller/rest/InventoryReportRestController.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private Timestamp parseDate(String value) {
+        if (value == null || value.isBlank()) {
+            return null;
+        }
+        try {
+            SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
+            sdf.setLenient(false);
+            return new Timestamp(sdf.parse(value).getTime());
```
mherman22, line 102: This needs fixing before merge. `parseDate("2026-08-09")` returns `2026-08-09 00:00:00.000`, and both date-scoped reports hand that straight to an inclusive `BETWEEN :startDate AND :endDate` (`InventoryUsageDAOImpl.getByDateRange` added here, `InventoryTransactionDAOImpl.getByDateRange`). If merged as-is, everything recorded after midnight on the end date is excluded, so Usage Trends and Transaction History silently lose the whole last day of any range, and a same-day range (today to today, the obvious way to pull a day's consumption) comes back with no rows at all. Nothing throws; the totals just read low.

Two report services in this tree already handle the boundary: `FreezerReportServiceImpl.generateDailyLogData` uses `endDate.plusDays(1).atStartOfDay(zone).minusNanos(1)`, and the legacy `Report.DateRange.getHighDateAtEndOfDay()` does the same. I would add an end-of-day variant used only for `endDate` here, so both report types get the fix in one place.

src/main/java/org/openelisglobal/inventory/report/InventoryReportWriter.java
```diff
@@ -0,0 +1,144 @@
...
+    private static String csvEscape(String value) {
+        if (value == null) {
+            return "";
+        }
+        if (!value.isEmpty()) {
+            char first = value.charAt(0);
+            if (first == '=' || first == '+' || first == '-' || first == '@') {
+                value = "'" + value;
+            }
+        }
```
mherman22, line 138: This needs fixing before merge. The guard treats a leading `-` as a formula character, so a signed quantity picks up an apostrophe. I ran `writeCsv` on a Transaction History style row: a consumption of 5 units exports as `'-5` in CSV, while the same value renders as `-5` through `writePdf` and `writeExcel`.

Consumption and disposal both record negative changes (`InventoryManagementServiceImpl` records `-quantityNeeded`, `InventoryLotServiceImpl.disposeLot` records `-quantityDisposed`), so if merged as-is every outbound row in the Transaction History CSV lands as text rather than a number: a spreadsheet summing Quantity Change silently skips those cells, and the three export formats disagree about the same report. That is the one column somebody reconciling a stock card actually adds up.

The injection guard is worth keeping, so the narrow fix is to leave plain numbers alone:

```suggestion
        if (!value.isEmpty() && !value.matches("-?\\d+([.,]\\d+)?")) {
            char first = value.charAt(0);
            if (first == '=' || first == '+' || first == '-' || first == '@') {
                value = "'" + value;
            }
        }
```

frontend/src/components/inventory/InventoryReports.jsx
```diff
@@ -18,6 +18,12 @@ import { DocumentPdf, DocumentBlank, TableSplit } from "@carbon/icons-react";
 import { FormattedMessage, useIntl } from "react-intl";
 import { ReportsAPI } from "./InventoryService";
 
+// The DatePicker gives back raw JS Date objects; URLSearchParams would
+// otherwise stringify those via Date.toString() (e.g. "Mon Jul 13 2026..."),
+// which the backend's yyyy-MM-dd parser can't read.
+export const toIsoDate = (date) =>
+  date ? date.toISOString().slice(0, 10) : null;
```
mherman22, line 25: `toISOString()` converts to UTC before slicing, and the Carbon DatePicker hands back a Date at local midnight, so east of Greenwich the day moves backwards. Under `TZ=Africa/Nairobi` I get `new Date(2026, 6, 13).toISOString().slice(0, 10) === "2026-07-12"`; Abidjan and Port-au-Prince are unaffected, Nairobi, Kampala and Kigali are not. The report then covers a range one day earlier than the one on screen, in both bounds.

The test above cannot catch it, because it builds its input with `Date.UTC`, which is the one construction where the conversion is a no-op; it passes in every timezone. Building the string from the local parts (`getFullYear`, `getMonth`, `getDate`) keeps the day the user picked, and a test built with `new Date(2026, 6, 13)` would then be meaningful.

Not blocking, since several other components in the tree format picker dates the same way, but those are mostly display strings and this one decides which rows the report contains.

src/test/java/org/openelisglobal/inventory/controller/rest/InventoryReportRestControllerIT.java
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * The Reports tab ({@code InventoryReports.jsx}) called
+ * {@code /rest/inventory/reports/generate}, which never existed on the backend
+ * — every "Generate" click 404'd. Covers the new endpoint end-to-end: one
+ * report type across all 3 export formats to prove the writer pipeline works,
+ * plus the validation error paths (unknown reportType/exportFormat, missing
+ * date range for date-scoped report types).
+ */
+public class InventoryReportRestControllerIT extends BaseWebContextSensitiveTest {
```
mherman22, line 29: These 8 tests never execute. There is no failsafe plugin anywhere in the build, surefire 3.2.5's default includes are `**/Test*.java`, `**/*Test.java`, `**/*Tests.java` and `**/*TestCase.java` (I checked `getDefaultIncludes` in the plugin bytecode), the pom configures no `<includes>` of its own, and `backend.yml` runs a plain `mvn clean install`. A class ending in `IT` matches none of those patterns, so this file is compiled and skipped.

Worth fixing here rather than later, because this is the only coverage the endpoint has end to end, and right now a regression in it cannot fail CI. Renaming to `InventoryReportRestControllerTest` is the one-line version; wiring up failsafe is the other, but that needs a CI step too. `InventoryLotServiceIT` and `InventoryManagementServiceIT` have the same problem and belong to #4016.

src/main/java/org/openelisglobal/inventory/report/InventoryReportWriter.java
```diff
@@ -0,0 +1,144 @@
...
+
+    public static void writePdf(ReportTable table, OutputStream out) throws IOException {
+        try {
+            Document document = new Document(PageSize.A4.rotate());
+            PdfWriter.getInstance(document, out);
+            document.open();
+
+            Font titleFont = new Font(Font.FontFamily.HELVETICA, 14, Font.BOLD);
+            Font headerFont = new Font(Font.FontFamily.HELVETICA, 9, Font.BOLD, BaseColor.WHITE);
+            Font cellFont = new Font(Font.FontFamily.HELVETICA, 8);
```
mherman22, line 53: Worth deciding on before this ships: iText's built-in Helvetica encodes Cp1252, and glyphs it cannot map are dropped rather than substituted. I rendered a table through `writePdf` and extracted the page text: an item named `Реагент A` comes back as ` A`, an Amharic name comes back empty, `Réactif` survives. CSV and XLSX built from the same `ReportTable` keep all three intact, so the PDF is the only format that loses data, and it does so without an error.

That matters here because item names are free text and the product ships am_ET, ar, ru, uk, bg, zh, si and ta. The fix is an embedded Unicode font, `BaseFont.createFont(<ttf>, BaseFont.IDENTITY_H, BaseFont.EMBEDDED)`, which means adding a font asset to the repo. The audittrail and compliance PDF exporters have the same limitation today, so this is not a regression you introduced, which is why I am not blocking on it. I would still rather it were a decision than a support ticket.

PR #4018, opened 2026-08-09
feat(inventory): show the item code in the catalog and item form

mherman22:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

The frontend half of #4015. The inventory catalog gains a Code column and its search matches on code as well as name. The item form gets a Code field: fill it in to choose your own, or leave it blank and the server generates one from the name. On edit the field is shown but locked, because lot numbers already created from that code embed it (`InventoryManagementServiceImpl.generateLotNumber`), and the update payload never carries `code`.

**Typing and saving.** The field is free text while you type, so the caret stays where you put it. On save the value is normalized the way the server's `CodeGenerator` does (uppercase; runs of anything other than A-Z0-9 become one underscore; leading and trailing underscores dropped), and while you type the helper text shows "Will be saved as MY_REAGENT_1" whenever that form is non-empty and differs from what is typed. A value that normalizes to nothing (`!!!`) is sent as `null`, so the server generates a code from the name. The input is capped at 64 characters to match the column; the server truncates the rare normalized value that grows past it (an upper-cased `ß` becomes `SS`).

**Errors.** `InventoryService.post` keeps the server's `errorCode` and `params` on the rejected Error, so a duplicate code shows the react-intl message `An inventory item with code "MY_REAGENT_1" already exists.` instead of the raw backend string. Other callers of `post()` read only `err.message` and are unchanged.

**The lock is a product choice, not only a UI one.** `InventoryItemRestController.update` on `develop` copies the editable fields from the request and never touches `code`, so unlocking the field alone would show edits the server silently discards. Making the code editable needs the controller to copy, normalize and duplicate-check it as `insert` does. If you would rather it were editable, say so at review and that is the change to make.

**Testing.** 43 vitest tests pass under `frontend/src/components/inventory` on this head, including the catalog listing and code search, the form's create, edit, normalization, preview and duplicate-code paths, and a service-level test that a 400 body's `errorCode` and `params` survive `InventoryService.post`. Verified in a browser against a `scripts/dev-stack` deployment of this head: caret position while editing mid-field, the preview text, save normalization with a trailing space, blank and punctuation-only codes auto-generating, the translated duplicate message with the 400 body captured, the locked field on edit with the PUT body inspected, and the 64-character cap.

## Screenshots

Not applicable: the visible change is a column, a text field and its helper text. Wire samples from the verification run:

```
POST /rest/inventory/items  {"code":"MY_REAGENT_1", ...}   -> 201 {"id":9,"code":"MY_REAGENT_1"}
POST /rest/inventory/items  {"code":null, ...}             -> 201 {"id":10,"code":"VERIFY_R2_ITEM_D"}
POST /rest/inventory/items  {"code":"MY_REAGENT_1", ...}   -> 400 {"errorCode":"inventory.item.error.duplicateCode","params":{"code":"MY_REAGENT_1"}}
PUT  /rest/inventory/items/9 {"name":..., no "code" key}   -> 200
```

## Related Issue

Part of the inventory stack, reviewed bottom-up. #4015, #4016 and #4017 are merged, so the open part of the stack starts from `develop`:

4. **#4018 Show the item code in the catalog and item form** (this PR, based on `develop`)

## Other

Left alone on purpose, for a follow-up on `develop`: the en.json text for `common.codeGenerator.error.invalidCode` says "letter, digit, or underscore" while `CodeGenerator.toCode` strips underscores before the emptiness check; this form cannot reach that message (punctuation-only input is sent as `null`), only a direct REST client can.


frontend/src/components/inventory/InventoryItemForm.jsx
```diff
@@ -206,13 +209,23 @@ const InventoryItemForm = ({ open, onClose, onSave, item = null }) => {
...
       console.error("Error saving item:", err);
-      const errorMessage = err.message || "Error saving catalog item";
+      // err.errorCode (OGC-658 C8) is an en.json message id set by
+      // InventoryItemRestController for
+      // validation failures (duplicate/malformed code, etc.) — prefer it over
+      // err.message, which is the untranslated backend fallback string.
+      const errorMessage = err.errorCode
+        ? intl.formatMessage({ id: err.errorCode }, err.params)
+        : err.message ||
+          intl.formatMessage({ id: "catalog.item.error.saveGeneric" });
```
mherman22, line 228: `err.errorCode` is never set on the error that reaches this catch, so this branch never runs and a duplicate code still surfaces the raw backend English. That needs fixing before the code field ships, because it is the one error users will actually hit.

`InventoryItemAPI.create` goes through `post()` in `InventoryService.js`, which rejects with `new Error(json.message || json.error || ...)`. The parsed 400 body does carry `errorCode` and `params` (Utils.ts spreads the JSON before adding `status`), but `post()` reads only `message` off it and a bare `Error` has nowhere to put the rest. I mocked `postToOpenElisServerJsonResponse` with the exact map `InventoryItemRestController.create` builds from a `LocalizedValidationException` and awaited the rejection: `message` arrived, `errorCode` and `params` came back `undefined`. So a user who types an existing code gets `Inventory item code already exists: MY_REAGENT` in every locale, and `inventory.item.error.duplicateCode` (plus `common.codeGenerator.error.invalidCode` for a code with no alphanumerics) never renders.

The duplicate-code test passes because it hand-builds an `Error` with `.errorCode` already on it, which the service layer never produces, so the suite cannot catch this.

One hunk in `InventoryService.js` fixes it, and `useDeleteLocation.js` already does the same thing for storage:

```js
const err = new Error(
  json.message ||
    json.error ||
    `Request failed with status ${json.status || json.statusCode}`,
);
err.errorCode = json.errorCode;
err.params = json.params;
reject(err);
```

I applied that locally and the probe went green. Worth trimming the comment once it is true, as well: four lines carrying two separate justifications, and it describes the HTTP body rather than `err`, which is exactly the gap above.

frontend/src/components/admin/testCatalog/sections/ReagentsSection.test.jsx
```diff
@@ -30,7 +30,7 @@ import messages from "../../../../languages/en.json";
 const TWO_REAGENTS = [
   {
     id: "link-1",
-    reagentId: 7,
+    reagentId: "GLUCOSE_REAGENT",
```
mherman22, line 33: `reagentId` is a `Long` on both sides of this API, so these string codes describe a payload the server cannot produce. `TestReagentLinkRestController.ReagentLinkResponse.reagentId` is a `Long`, the update and unlink mappings take `@PathVariable Long reagentId`, and `LinkReagentModal` reads `i.id` from `/rest/inventory/items/type/REAGENT` where `InventoryItem.id` is a `Long` too.

With this fixture the suite asserts `PUT /rest/test-catalog/42/reagents/GLUCOSE_REAGENT`, a URL that 400s against the real controller, and it would stay green if `ReagentsSection` ever started sending item codes where the endpoint needs ids, which is the one confusion these tests exist to catch. The two `{ id: "GLUCOSE_REAGENT" }` fixtures in `LinkReagentModal.test.jsx` have the same problem.

Nothing else in this PR touches reagent-link code, and I checked #4019, #4026 and #4033 for a later move to code-keyed links and found none, so I would drop both files from the branch and leave the numeric ids alone.

frontend/src/components/inventory/InventoryItemForm.test.jsx
```diff
@@ -0,0 +1,156 @@
...
+    expect(codeInput).toBeDisabled();
+    expect(codeInput).toHaveValue("EXISTING_CODE");
+
+    fireEvent.click(screen.getByText("Save"));
+
+    await waitFor(() => {
+      expect(InventoryItemAPI.update).toHaveBeenCalled();
+    });
+    const [, payload] = InventoryItemAPI.update.mock.calls[0];
+    expect(payload.id).toBeUndefined();
```
mherman22, line 122: This asserts on `payload.id`, but the test is about the code, and `payload.code` is what has to stay absent. Nothing here would fail today if `handleSave` started folding `code` into `sanitizedData` on the edit path, which is the invariant the PR description leans on.

```suggestion
    expect(payload.code).toBeUndefined();
```

Keep the `id` line as well if that check was deliberate.

PR #4019, opened 2026-08-09
fix(inventory): correct stock visibility, barcodes and item codes

mherman22:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Everything here came out of code review or of driving the inventory module end to end on a dev stack after #4015 to #4018 merged. Each item was a reproducible failure, not a tidy-up.

**The QC gate was invisible.** Receive a lot, record usage against it, and you got `No available lots for item: 11`. New lots start QC-pending and only PASSED stock can be consumed; the gate is right, so this makes it visible: a QC Status column and a "Pending QC" stock status on the dashboard (shown even when the item is also low on stock), a hint on the receive form, and a refusal that names the item by name and code, counts only unexpired lots with stock, and says whether that stock is quarantined, awaiting QC or failed QC (409 with `errorCode` `inventory.consume.error.*`); quarantine is reported first, because a quarantined lot keeps the default pending QC status and passing QC would not release it. `check-availability` now applies the same per-lot rule consumption enforces, so it cannot say yes to stock consume would refuse.

**Blank barcodes collided.** Saving a lot without a barcode stored `''`; the second one hit the unique index and every barcode-less lot became uneditable. Blank or whitespace barcodes now become `NULL` at the service layer on every write path (REST create and update, receive, and the internal updates), not only in the React form; a barcode already on another lot answers 400 with a translatable `{message, errorCode, params}` body (`inventory.lot.error.duplicateBarcode`) instead of a bodyless 500. Changeset 072 cleans rows already saved as `''`.

**Low stock is one rule in one place.** The dashboard tile was always empty (it filtered on a field that does not exist). It now comes from the backend, and the tile, the alerts endpoint and the Low Stock report share `InventoryLot.countsAsAvailableStock()`: usable now or awaiting QC; never expired, disposed, quarantined or failed; strictly below threshold, so threshold 0 means "no alert" and stock awaiting QC does not trigger a reorder. That trades a native query for one lot query per active item, fine at inventory's scale and noted in the code.

**Item codes are prefix plus counter.** A blank Code field used to produce the whole name in UPPER_SNAKE with `_2`, `_3` on collision, so the name still determined identity. Generated codes are now a prefix derived from the name plus a per-prefix counter that is authoritative: `Paracetamol 500mg Tablets` becomes `PAR-500MG-001`, the next item with that prefix `PAR-500MG-002`, `Sodium Chloride 0.9% 500mL` becomes `SOD-09-500ML-001`, `HIV 1/2 Rapid Test Kit` becomes `HIV-12-001`. The prefix is the first three letters of the first word containing a letter (`ITEM` when there is none), followed by up to two *other* words containing a digit, each capped at eight characters; the word the letters came from is not reused as a digit word, so `2mL Tube` gives `ML-001` rather than `ML-2ML-001`. The counter lives in `clinlims.inventory_item_code_sequence` (new changeset 073, one row per prefix) and is taken inside the item's insert transaction with a row lock (`INSERT ... ON CONFLICT DO NOTHING`, then `UPDATE ... RETURNING`), so concurrent inserts for one prefix serialize; a slot already held by a legacy or typed code is skipped. The database primary key stays `id`; `code` stays the immutable business identifier under `uq_inventory_item_code`. The separator for typed codes is now `-` on both server (`CodeGenerator.normalize`) and client (`InventoryItemForm.toCode`), so `my reagent 1` saves as `MY-REAGENT-1`; codes already stored with `_` are unchanged. Blank lot numbers, which reuse `CodeGenerator`, follow the same separator (`PAR-500MG-001-20260918`). The Code field hint in `en.json` shows the `PAR-500MG-001` shape; browsers on en-GB or en-US see the previous sentence until Transifex syncs those bundles, because the language loader layers the locale bundle over `en.json`.

**Catalog and dashboard fixes from the walkthrough.** Catalog row actions (Edit, Activate, Deactivate) resolve the item by row id, so they hit the right item after sorting; the threshold column shows 0 instead of `-`. The dashboard Status filter filters (client-side, like the type filter) and offers Disposed; the dashboard refetches when its tab becomes active so a Catalog edit is not stale; the move reason typed in the location picker is sent instead of being replaced by the notes. Update QC Status offers exactly PENDING, PASSED, FAILED, QUARANTINED (NOT_REQUIRED was rejected by the database check constraint). `InventoryLotStorageAPI.getMovements` is restored so Movement History in the lot details panel loads again, and its test spies on the real module so a removed wrapper can no longer be mocked into existence. Record Usage is presented as item-level FEFO (the only consume endpoint is item-level): it names the item and explains that the earliest-expiring QC-passed lot is decremented, which may not be the lot it was opened from. Editing a legacy reagent whose stability-after-opening is NULL no longer requires inventing a value.

**Lot mutation guards.** Adjusting a DISPOSED or CONSUMED lot and disposing a lot twice are refused with a 400 body instead of writing bogus transactions; missing lot or item ids answer 404 with a JSON body instead of an empty 500 on the lot create, update, adjust and dispose endpoints. The lot and item controllers share one error-body helper (`InventoryErrorBody`): `{message, errorCode, params}` for localizable validation, `{error}` otherwise; the management controller's own `{error}` bodies have the same shape.

**Location picker.** The level cascade renders alongside search (no "Create new location" toggle); picking a search result rebuilds the ancestor chain and no longer triggers a second search for the full path; boxes are included in search results. `LocationPickerInline` honours `allowCreate={false}` again, so the order-label step is pick-only. The device-type prompt no longer renders a raw `Select {level}`, `storage.nav.box` has a message, and the two orphaned picker keys are gone from en.json.

**React warnings that were real.** `LotEntryModal` and the four action modals set state after unmounting; Carbon dropdowns flipped from uncontrolled to controlled when their selection was `undefined`.

**Translations.** Strings this PR added or touched (stock status labels, dashboard status options, QC status options, the stability and quantity validation messages, the new error codes) go through react-intl with keys in `en.json` only; the eight `qc.rules.*` keys are back in their literal form.

**Dead code.** `InventoryItemDAO.getLowStockItems` and its implementation (replaced by the shared predicate above), and nineteen frontend API wrappers with no caller left (`getAllActive`, `getByType`, `search`, `getStockLevel`, the `InventoryLotAPI` read helpers, `checkAvailability`, `getAlerts`, and the `TransactionAPI`/`UsageAPI` lookups).

**Testing.** Backend: fifteen inventory test classes, including new `InventoryLotServiceTest`, `InventoryManagementServiceAvailabilityTest`, `CodeGeneratorTest` and `InventoryItemCodeSequenceTest` (the last creates two same-named items against the real database and reads back `-001` and `-002`). Frontend: 176 passing across `components/inventory` and `storage/LocationPicker`, with new unmount-guard, sort-then-act, filter, picker and error-translation tests. The Playwright spec `storage-assign-dashboard.spec.ts` drops its "open cascade view" step, since the cascade now renders alongside search.

## Screenshots

Not applicable; wire samples will be added from the verification run.

## Related Issue

Part of the inventory stack, reviewed bottom-up. #4015, #4016, #4017 and #4018 are merged, so the open part starts from `develop`:

5. **#4019 Fix what review and hands-on testing turned up** (this PR, based on `develop`)

## Other

Found in the walkthrough and deliberately left to the PRs that own the area: lots are not yet visible under `/Storage` and a disposed lot keeps its storage assignment (#4033, lot storage parity). Follow-ups on `develop`: non-ASCII letters are dropped silently by `CodeGenerator` (`émile` becomes `MILE`); `/rest/inventory/items/search` matches name only with unescaped LIKE wildcards; `PUT /rest/inventory/lots/{id}` without `version` bypasses optimistic locking; `open`, `qc-status` and `status` still answer bodyless 400/500 on a missing lot id.


frontend/src/components/inventory/InventoryService.js
```diff
@@ -254,10 +208,6 @@ export const InventoryLotStorageAPI = {
   getLocation: (lotId) =>
     promisify(getFromOpenElisServer, `${STORAGE_BASE_PATH}/${lotId}`),
 
-  // List movement-audit rows for a lot
-  getMovements: (lotId) =>
```
mherman22, line 258: `LotDetailsPanel` still calls this one, at `LotDetailsPanel.jsx:49`, so it is not dead.

If merged as-is, opening a lot's details throws `TypeError: InventoryLotStorageAPI.getMovements is not a function` inside `fetchLotDetails`, whose `catch` only `console.error`s, so `movements` stays `[]` and the Movements tab renders as though the lot has never been moved. The transactions and usage calls run before it and still populate, so the panel looks fine and the empty tab reads as real data.

`LotDetailsPanel.test.jsx` cannot catch this: its `vi.mock` factory hand-defines `InventoryLotStorageAPI: { getMovements: vi.fn() }`, so the test never touches the real module. I checked the real one at this head and `typeof InventoryLotStorageAPI.getMovements` is `"undefined"`.

Either keep the wrapper, or remove the Movements tab in the same change.

frontend/src/components/storage/LocationPicker/LocationPickerInline.jsx
```diff
@@ -63,65 +65,52 @@ export default function LocationPickerInline({
...
+        onSelect={handleSearchSelect}
+        selectedSelection={state.selection}
+      />
+      <p className="storage-location-picker-browse-label">
+        {intl.formatMessage({
+          id: "storage.picker.orBrowse",
+          defaultMessage: "Or pick level by level",
+        })}
+      </p>
+      <CreateForm selection={state.selection} onLevelChange={setLevel} />
```
mherman22, line 97: `CreateForm` now renders unconditionally, but `OrderLabel.jsx:1330` still passes `allowCreate={false}`, and nothing in this component reads that prop any more.

If merged as-is, the Order Label step gets the full five-level cascade with an "Add new {level}" button on each row, and those buttons POST to `/rest/storage/{rooms|devices|shelves|racks|boxes}`. A step that was deliberately restricted to picking existing locations can now create them, and nothing in the diff or the tests says that was intended (the inline picker's mode tests went away with `SET_MODE`, so nothing covers the prop either).

Either honour the prop again, gating `CreateForm` or at least its add buttons on it, or drop it from the signature and update `OrderLabel`. Whichever way it goes, the doc block on this component still describes a shell that "toggles the mode", which no longer exists.

src/main/java/org/openelisglobal/storage/service/StorageLocationServiceImpl.java
```diff
@@ -1216,6 +1217,49 @@ public List<Map<String, Object>> searchLocations(String searchTerm) {
...
+            // the format under "boxType" the way devices keep "deviceType".
+            Object physicalType = box.get("type");
+            if (physicalType != null) {
+                result.put("boxType", physicalType);
+            }
+            result.put("type", "box");
+
+            // getBoxesForAPI already composes the full Room > Device > Shelf > Rack >
+            // Box path; only rebuild if it is missing.
+            Object existingPath = box.get("hierarchicalPath");
```
mherman22, line 1236: The path `getBoxesForAPI` composes uses ASCII `>`, but every other branch of `searchLocations` uses `›`, and `SearchField` splits `hierarchicalPath` on `" › "` to work out depth and to separate the ancestors from the leaf.

If merged as-is, a box matched directly by label or code renders as a single unbroken bold line (`Main Lab > Freezer 1 > Shelf A > Rack 2 > Box 7`) with no indentation and no muted ancestor breadcrumb, unlike every other result in the same list. The split yields one part, so depth is 0. On top of that, the results are sorted by `hierarchicalPath`, and `>` (U+003E) sorts before `›` (U+203A), so the box is placed away from its own ancestors rather than grouped with them, which is what the sort comment says it is for. Boxes that arrive via `expandRackSubtree` already get `›` and look correct, so the two kinds of box row render differently.

The separator needs fixing on the path that actually runs, not only in the `existingPath` fallback underneath, which never executes because `getBoxesForAPI` always sets `hierarchicalPath`. `StorageLocationServiceSearchTest.testSearchLocations_Box_IncludesAllParentIdsAndHierarchyType` currently pins the ASCII form, so that fixture moves with it.

frontend/src/components/storage/LocationPicker/LocationPickerModal.css
```diff
@@ -68,7 +68,15 @@
...
-.storage-location-picker-modal-picker .storage-location-picker-search-results {
+.storage-location-picker-browse-label {
+  margin: 1rem 0 0.5rem;
+  font-size: 0.75rem;
+  color: var(--cds-text-secondary, #525252);
+  text-transform: uppercase;
+  letter-spacing: 0.02em;
+}
+
+.storage-location-picker-search-results {
```
mherman22, line 79: Worth moving `position: relative` out of the modal scope alongside this.

The comment on `.storage-location-picker-modal-picker .storage-location-picker-search` still says the layout is kept modal-scoped so non-modal instances don't turn into overlays, which is the opposite of what this rule now does, and it will mislead whoever edits here next. More practically, `.storage-location-picker-search` only picks up `position: relative` from that modal-scoped rule and from a global rule in `order/order-workflow.scss`. The order pages pull that stylesheet in; the Notebook entry form and `/Storage/sample-items/:id/manage-location` do not, so on those two the results list is absolutely positioned against whatever ancestor happens to be positioned instead of its own wrapper.

I did not drive this in a browser, so I can't say where it actually lands on those pages, but the stale comment should go either way and the wrapper rule is one line.

src/main/resources/liquibase/3.5.x.x/072-null-blank-lot-barcodes.xml
```diff
@@ -0,0 +1,31 @@
...
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
+
+    <!--
+    inventory_lot.barcode is UNIQUE and nullable, so "no barcode" must be NULL.
+    Saving a lot with a blank barcode stored an empty string instead, and the
+    unique index then rejected the next one ("Key (barcode)=() already exists"),
+    blocking any further edit of a barcode-less lot. The write paths now
```
mherman22, line 13: "The write paths" here is one React component. `LotEntryModal` trims blank to null on both the receive and the update payload, and that is the only normalisation in the change.

`POST /rest/inventory/lots`, `PUT /rest/inventory/lots/{id}` and `POST /rest/inventory/management/receive` all still accept `"barcode": ""` and persist it, and `InventoryLot.barcode` has no normalisation of its own, so the collision comes back for any writer that isn't this modal. Because the changeset is `MARK_RAN` as soon as there is nothing to clean, it won't pick those rows up on a later upgrade either.

Normalising where the lot is persisted (the service, or the entity) would make the invariant hold for every caller and let this comment be literally true.

frontend/src/languages/en.json
```diff
@@ -1266,6 +1266,7 @@
   "label.validation.qc.banner.title": "QC failures detected — {count} QC sample(s) failed acceptance criteria",
   "label.validation.qc.failedSamples": "Failed QC Samples",
   "label.validation.qc.justification": "Justification",
+  "label.validation.viewPatient": "View Patient",
```
mherman22, line 1269: This key isn't referenced anywhere in the tree, and it belongs to the validation page rather than to inventory or storage, so it looks like it rode along by accident.

Same for the eight `qc.rules.*` entries at 7661-7674, which got rewritten from literal subscript characters to `\u` escapes. I checked that they parse to identical keys and values, and that running Prettier over the base file leaves the literal form alone, so it isn't the formatter doing it. Harmless, but the description says en.json only gained keys, and a translations file that only ever grows is what lets the next reviewer skip past it at a glance.

Issue #3654, opened 2026-05-26
`updateShipment()` Pre-sets `@Version` Field Before `merge()`, Causing `OptimisticLockException`


Agaba-derrick:
Any call to `ShipmentService.updateShipment()` will always fail with an `OptimisticLockException`.
`BaseObject.lastupdated` is annotated with `@Version`, enabling Hibernate's optimistic locking. When Hibernate performs a `merge()`, it generates a SQL `UPDATE` with a `WHERE lastupdated = <version_loaded_from_DB>` clause to detect concurrent modifications.

`ShipmentServiceImpl.updateShipment()` manually calls `setLastupdated(now)` before calling `entityManager.merge()`:

```java
// ShipmentServiceImpl.java — updateShipment()
shipment.setLastupdated(new Timestamp(System.currentTimeMillis())); // ← mutates @Version
shipmentDAO.update(shipment); // ← merge() uses the now-wrong version value
```

This overwrites the original loaded version with the current time. Hibernate then issues:

```sql
UPDATE shipment SET ... WHERE id=? AND lastupdated='<now>'
```

…but the DB holds the old timestamp → 0 rows updated → `StaleObjectStateException` → `OptimisticLockException`.

### Steps to Reproduce

1. Load any existing shipment: `shipmentService.getShipmentById(1)`
2. Modify a field: `shipment.setTrackingNumber("new-value")`
3. Call `shipmentService.updateShipment(shipment)`

### Error

```
OptimisticLockException: Row was updated or deleted by another transaction
(or unsaved-value mapping was incorrect): [Shipment#1]
Caused by: StaleObjectStateException
```

### Root Cause

`ShipmentServiceImpl.java` — `updateShipment()` at line ~103 manually pre-sets the `@Version` field, corrupting the version value Hibernate relies on for its optimistic lock `WHERE` clause.

### Affected Files

- `src/main/java/org/openelisglobal/shipment/service/ShipmentServiceImpl.java`

### Recommended Fix

Remove the manual `setLastupdated()` call from `updateShipment()` and let Hibernate manage the `@Version` field automatically:

```java
// BEFORE (buggy)
public Shipment updateShipment(Shipment shipment) {
    shipment.setLastupdated(new Timestamp(System.currentTimeMillis())); // ← REMOVE THIS
    shipmentDAO.update(shipment);
    ...
}

// AFTER (correct)
public Shipment updateShipment(Shipment shipment) {
[...]

PR #4021, opened 2026-08-09
Fix OptimisticLockException in ShipmentServiceImpl

omgitsshriharsh:
Resolved OptimisticLockException by removing manual setLastupdated overrides
# Pull Requests Requirements

[x] The PR title includes a brief description of the work done, including the Issue number if applicable.

[x] The PR title follows conventional commit label standards.

[x] The changes confirm to the OpenElis Global x3 Styleguide and Design documentation.  

[x] The changes include tests or are validated by existing tests.  

[x] I have read and agree to the Contributing Guidelines of this project.

## Summary
Fixed an OptimisticLockException that occurred whenever an existing shipment was updated.

Root cause was that ShipmentServiceImpl was explicitly calling shipment.setLastupdated(...) right before calling the DAO update, which mutated the @Version field prematurely and caused Hibernate's version check to fail. Removed these manual overrides from both updateShipment() and updateShipmentStatus() so that Hibernate can manage @Version natively. Also updated updateShipment() to return the managed instance from shipmentDAO.update().

## Screenshots
ss1
image

## Related Issue
Github #3654

## Other
Learned through debugging this issue how Hibernate's @Version annotation natively handles optimistic locking and timestamp tracking, bypassing the need for manual overrides


src/main/java/org/openelisglobal/shipment/service/ShipmentServiceImpl.java
```diff
@@ -117,7 +117,7 @@ public Shipment updateShipmentStatus(Integer id, ShipmentStatus newStatus) {
                     .orElseThrow(() -> new IllegalArgumentException("Shipment not found with ID: " + id));
 
             shipment.setStatus(newStatus);
-            shipment.setLastupdated(new Timestamp(System.currentTimeMillis()));
+            // Manual setLastupdated removed so Hibernate @Version takes over
```
mherman22, line 121:
```suggestion
```

src/main/java/org/openelisglobal/shipment/service/ShipmentServiceImpl.java
```diff
@@ -100,15 +100,15 @@ public Shipment createShipment(Shipment shipment) {
     @Override
     public Shipment updateShipment(Shipment shipment) {
         try {
-            shipment.setLastupdated(new Timestamp(System.currentTimeMillis()));
-            shipmentDAO.update(shipment);
-            logger.info("Updated shipment with ID: {}", shipment.getId());
-            return shipment;
+            // Hibernate @Version handles lastupdated automatically; capture updated instance
```
mherman22:
```suggestion
```

mherman22 (approved):
Just some clean up but otherwise this LGTM, thank you!
@omgitsshriharsh Do you mind adding some unit tests for this? Basically writing up some failing tests to sort of reproduce the bug and then make sure the tests pass with your changes here?

Agaba-derrick (changes requested):
@omgitsshriharsh  Do you wanna check  out the CI failings

src/main/java/org/openelisglobal/shipment/service/ShipmentServiceImpl.java
```diff
@@ -100,10 +100,11 @@ public Shipment createShipment(Shipment shipment) {
     @Override
     public Shipment updateShipment(Shipment shipment) {
         try {
-            shipment.setLastupdated(new Timestamp(System.currentTimeMillis()));
-            shipmentDAO.update(shipment);
-            logger.info("Updated shipment with ID: {}", shipment.getId());
-            return shipment;
+            // Hibernate @Version handles lastupdated automatically; capture updated
+            // instance
+            Shipment updated = shipmentDAO.update(shipment);
```
mherman22, line 105: This still needs a regression test before it goes in. Nothing in the tree calls `updateShipment` yet, so if someone stamps a timestamp back into this method later there is no signal at all that the bug is back.

src/main/java/org/openelisglobal/shipment/service/ShipmentServiceImpl.java
```diff
@@ -100,10 +100,11 @@ public Shipment createShipment(Shipment shipment) {
     @Override
     public Shipment updateShipment(Shipment shipment) {
         try {
-            shipment.setLastupdated(new Timestamp(System.currentTimeMillis()));
-            shipmentDAO.update(shipment);
-            logger.info("Updated shipment with ID: {}", shipment.getId());
-            return shipment;
+            // Hibernate @Version handles lastupdated automatically; capture updated
+            // instance
```
mherman22, line 104: These two lines describe the change rather than the code, so once the diff is history they only tell a reader what used to be here. The suggestion I left on this went outdated when spotless reflowed the line, so re-anchoring it.

```suggestion
```

PR #4026, opened 2026-08-10
Make the inventory lot barcode findable

mherman22:
Stack, reviewed bottom-up. #4015 to #4019 are merged, so the open part of the stack now starts from `develop`.

5. ~~#4019 Correct stock visibility, barcodes and item codes~~ merged
6. **Make the inventory lot barcode findable** — you are here

Based on `develop` (#4019 merged as 97a6651cc). Ten commits are new here.

The Barcode field on the lot modal was write-only: you could type a value and
it was stored and enforced unique, but nothing ever read it back.
`InventoryLotDAO.getByBarcode` existed only to enforce uniqueness on write
(#4019 calls it from `normalizeBarcode`); nothing ever looked a lot up by the
value. The dashboard search matched only lot number and item name, and the lot
details panel did not show the barcode at all. Nothing in the UI said whether the field meant the
manufacturer's barcode or an internal one, which matters because the UNIQUE
constraint is only correct for the internal reading — a GTIN is shared by every
lot of a product.

The barcode is now the lab's own label and the system mints it. Leave the field
blank and the service generates one from the item code and lot number, so a
human can still identify the lot when the printed label is damaged; supply one
explicitly and it is normalised instead — on receipt, and equally on the one
edit that gives a legacy lot its first barcode. A barcode a lot already holds is
never reshaped by a later save, because a label has been printed from it. `CodeGenerator` upper-cases and
collapses every run of non-alphanumerics to a hyphen, so item `PAR-500-001`
with lot `lot/2025-001` yields `PAR-500-001-LOT-2025-001`, and a taken value is
suffixed `-2`, `-3`. Where the lot number already starts with the item code, the
code is not prefixed twice.

The explicit-barcode branch mirrors `InventoryItemServiceImpl.resolveCode`:
normalise, then refuse a value another lot holds, with a localized error rather
than a raw constraint violation. The generated branch differs deliberately — an
item code draws a per-prefix database sequence, while a lot barcode is generated
from a readable seed and suffixed numerically on collision. Lookup is wired
through to `GET /rest/inventory/lots/barcode/{barcode}` (nothing in this repo
calls it yet; the dashboard search filters the loaded list client-side) and the
dashboard search now matches on barcode.

The dashboard resolves a row to its lot by id rather than by array index, so a
row action cannot target a different lot once the table is sorted. That fix
landed in #4019 together with the test that guards it; this PR adds nothing
there.

A barcode is locked once a lot HAS one. A lot that carries no barcode — every
lot predating this change — can still be given one from the edit modal, which is
the same case the server allows: `InventoryLotRestController.update` restores the
stored barcode only when there is one. Existing lots are deliberately not
backfilled: `inventory_lot.barcode` is nullable and UNIQUE, so a migration would
have to invent a unique value for every legacy row and lock it before anyone had
chosen it. A legacy lot gets its barcode the first time someone edits it —
normalised on the way in, duplicate-checked against the normalised form, and
resolvable by the hand-keyed lookup afterwards — and only then becomes
immutable.

Printing a lot label is included: a `print-label` endpoint, the label type, and
a Print label action on the dashboard row menu.

Deliberately not here: scan-to-act on the dashboard and
GS1 parsing to auto-fill lot number and expiry from a manufacturer barcode.
Each is a feature in its own right and depends on this being wired up first.


src/main/java/org/openelisglobal/inventory/service/InventoryLotServiceImpl.java
```diff
@@ -39,6 +46,64 @@ public InventoryLot getForUpdate(Long lotId) {
...
+     * value otherwise, and reject a duplicate rather than letting the UNIQUE index
+     * surface a raw constraint error.
+     */
+    private String resolveBarcode(InventoryLot lot) {
+        String supplied = lot.getBarcode();
+        String barcode = (supplied == null || supplied.trim().isEmpty())
+                ? CodeGenerator.generateFromName(barcodeSeed(lot), BARCODE_MAX_LENGTH, "LOT", this::barcodeExists)
+                : CodeGenerator.normalize(supplied, BARCODE_MAX_LENGTH);
+        if (barcodeExists(barcode)) {
+            throw new LocalizedValidationException("inventory.lot.error.duplicateBarcode",
```
mherman22, line 69: If merged as is, a technician who types a barcode that another lot already holds sees "Internal server error" instead of this message, so `inventory.lot.error.duplicateBarcode` is a key nothing ever renders.

The lot modal creates through `POST /rest/inventory/management/receive`. `InventoryManagementRestController.receiveInventory` catches `IllegalArgumentException` and then `Exception`; `LocalizedValidationException` extends `LIMSRuntimeException` extends `RuntimeException`, so it takes the second branch and the response is a 500 carrying `{"error":"Internal server error"}`. `post()` in InventoryService.js rejects with that string, and `LotEntryModal` puts `err.message` straight into the error banner. `InventoryLotRestController.create` loses it the same way, with `.build()` and no body at all.

`InventoryItemRestController.create` already has the branch that makes the sibling work (400 plus `errorCode` and `params`), so I would copy it onto both lot create paths. One caveat while you are there: `post()` keeps only `message`, dropping `errorCode` and `params`, so until that helper carries them the best the modal can show is the untranslated fallback. That half is shared with the item path and can wait. The 500 cannot.

src/main/java/org/openelisglobal/inventory/service/InventoryLotService.java
```diff
@@ -39,6 +39,12 @@ public interface InventoryLotService extends BaseObjectService<InventoryLot, Lon
      */
     InventoryLot getByLotNumber(String lotNumber);
 
+    /**
+     * Get lot by its internal barcode. Blank barcodes are stored as NULL, so a
```
mherman22, line 43: Blank barcodes are not stored as NULL any more. `resolveBarcode` generates one from the seed when the supplied value is blank, and `generateFromName` falls back to `"LOT"` when the seed has no alphanumerics, so `insert` never writes NULL or an empty string. `insert_shouldGenerateBarcodeWhenSuppliedValueIsBlank` in this PR asserts exactly that.

The second half of the sentence still holds, since the implementation short-circuits a blank argument, so it is only the storage claim that needs to go. The same belief is written into the fixture in `inventory-test-data.xml` ("Production normalizes blank to NULL on write"). What lot 1002 actually stands in for now is a row written before this change, or one that came through `PUT /lots/{id}`, which skips `resolveBarcode` entirely, and saying so would make the fixture easier to reason about.

src/main/java/org/openelisglobal/barcode/labeltype/InventoryLotLabel.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Label for an inventory lot. Carries the lot's internal barcode, with the item
+ * name and lot number above it and the expiry below, so a technician can read
+ * the container without scanning it.
+ *
+ * Dimensions are fixed rather than configuration-driven: storage location
+ * labels expose width/height as admin properties, but lots have no such
+ * property yet and the 3x1 inch default is the same stock those use.
+ */
+public class InventoryLotLabel extends Label {
```
mherman22, line 16: The description says lot label printing is deliberately not in this PR, and that two commits are new when there are six. Printing is most of what landed: this class, `InventoryLotLabelService` and its impl, the `print-label` endpoint, `InventoryLotLabelServiceIT`, the dashboard menu item and its handler, `InventoryService.printLabel`, three en.json keys and two message_en keys.

I read it and did not find anything wrong with it, which is the problem: a reviewer who trusts the description will not read it at all, and the cost of establishing that it is safe lands on whoever notices later. I would split it onto its own PR stacked on this one, since it depends on the barcode existing and nothing here depends on it. If you would rather keep one PR, the description needs rewriting so the scope it claims matches the scope it has.

frontend/src/components/inventory/InventoryDashboard.test.jsx
```diff
@@ -324,3 +325,155 @@ describe("InventoryDashboard Location column", () => {
...
+    const table = document.querySelector("table");
+    expect(within(table).getByText("LOT-100")).toBeInTheDocument();
+    expect(within(table).queryByText("LOT-200")).not.toBeInTheDocument();
+  });
+});
+
+describe("InventoryDashboard row-to-lot mapping", () => {
+  // The table is sortable, so Carbon's row order need not match the order of
+  // the lots array. Resolving the lot by row index made every row action —
+  // move, dispose, QC — target whichever lot happened to sit at that index.
```
mherman22, line 391: `lotsById` and `lotsById.get(row.id)` are already on `inventory/5-review-fixes`, at lines 396 and 657 of the base copy of InventoryDashboard.jsx, so no index-to-id change happened in this PR. I replaced the file with the base version and ran this test against it and it passed, so it guards the behaviour rather than reproducing a bug this branch fixed.

Keep the test. It is the description and this comment that need correcting: "rows were matched to lots by array index ... Now matched by id, with a regression test that fails against the old lookup" describes work that is not in this diff. The barcode-search test in the block above it does fail on base, so that one is the real thing.

src/main/java/org/openelisglobal/barcode/labeltype/InventoryLotLabel.java
```diff
@@ -0,0 +1,101 @@
...
+    public int getNumTextRowsBefore() {
+        return countRows(aboveFields);
+    }
+
+    @Override
+    public int getNumTextRowsAfter() {
+        return countRows(belowFields);
+    }
+
+    private int countRows(java.util.List<LabelField> fields) {
```
mherman22, line 68: `Label` already provides this. `protected int getNumRows(Iterable<LabelField>)` at Label.java:277 does the same newline and colspan walk against the same 20-column width, and `aboveFields` and `belowFields` are `ArrayList<LabelField>`, so `return getNumRows(aboveFields);` drops in and this method plus `MAX_COLUMNS` both go away.

BlankLabel, BlockLabel, FreezerLabel, SlideLabel, SnapshotLabel and SpecimenLabel all call the helper. StorageLocationLabel open-codes it twice, which is presumably where this shape came from, but that is an argument for using the helper rather than for adding a third copy that has to be kept in step if the column width ever changes.

PR #4033, opened 2026-08-10
feat(storage): list inventory lots beside samples, free a disposed lot's box, and fold six nav rows into one page

mherman22:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Based on `develop`. Thirty-five commits are new here.

Lots and samples already shared `sample_storage_assignment`, one audit trail,
one picker, and occupant-agnostic occupancy queries, but the integration only
ran one way. You could put a lot into a freezer from Inventory; standing in
Storage Management looking at that freezer, the lot did not exist.

The disposal gap was a live bug rather than a missing feature. Disposing a
SampleItem clears the assignment's location fields, which is what frees the
slot; `disposeLot` never touched the assignment, so a disposed lot occupied its
box forever. Because occupancy is shared, that phantom occupancy inflated
capacity figures samples also read. Disposal now releases the slot and writes a
release movement, keeping the assignment row for audit exactly as samples do.

Both halves run in one transaction. `SampleStorageService.disposeInventoryLot`
flips the lot to DISPOSED and releases the slot inside a single `@Transactional`
unit, the same shape as `disposeSampleItem`; previously they committed
separately, so a failure in the release left a DISPOSED lot still holding its
box. It lives on the storage service rather than the lot service because
`SampleStorageServiceImpl` already depends on `InventoryLotService` and the
reverse edge would have been a cycle. This fixes new disposals: rows already
orphaned by the old behaviour are not retroactively repaired.

The rest is parity. `GET /rest/storage/inventory-lots` lists every lot that has
an assignment row, with its location resolved or empty. It is not the same shape
as `GET /rest/storage/sample-items`, which starts from all sample items: this
one starts from the assignment rows, so a lot never assigned storage does not
appear. `PATCH /rest/storage/inventory-lots/{id}` edits position and notes
without a move. A new Inventory Lots page shows lot number, barcode, item,
quantity, status and location.

### Storage Management rework

Storage spent six sidenav rows on one module, five of them
(Rooms/Devices/Shelves/Racks/Boxes) levels of a single hierarchy shown as peers,
mixed in with the occupant listings. Meanwhile `/Storage`, the dashboard, had no
way in at all: changeset `029-add-storage-subnav-items.xml` stripped its
`action_url` to make the parent expandable, so it was reachable only by typing
the URL.

Storage Management is now one sidenav row leading to a page shaped like
Inventory Management: a breadcrumb, then three tabs, Dashboard, Sample Items and
Inventory Lots. The five levels are tiles on the Dashboard tab, each showing how
many of that level are active, and picking a tile swaps the single table below
it. Each tab's contents only mount when that tab is selected, because Carbon's
`TabPanel` renders hidden panels rather than unmounting them, so without that
gate every visit also fetched the other two listings.

Every previous listing URL still resolves, so `/Storage/racks` opens the
Dashboard tab with the Racks table under the tiles and existing bookmarks keep
working. The creation, edit and manage-location routes are dropped, because each
of those is now a modal. Those URLs fall through to the router's catch-all, the
same way `/Storage/samples` has since changeset `006-storage-sidenav-rename.xml`
recorded that precedent; on this build the catch-all yields the framework's own
not-found response rather than a rendered page, which is what the cited
precedent does too.

Creating, editing and managing the location of a row are all modals now. One
`EditLocationModal` covers all five levels, so `EditLocationPage`, `EditBoxPage`
and their routes are gone; its fields are the union of what those two rendered
plus the `active` checkbox the Add modal omits, and it carries back the columns
it does not display, which every update handler used to null. Managing a sample
item's location opens the same `LocationPickerModal` the results and inventory
surfaces already use, so `ManageLocationPage`, `LocationPickerPage` and the
breadcrumb component they were the last users of are deleted; the picker itself
is untouched.

The Add, Edit and Delete modals and the notification they raise live once in
`StorageResourcePage` rather than in all five level pages. Add stays disabled
until a parent is chosen for any level below room, and a rejection that fails
bean validation shows the server's own field message instead of a bare status
code, on both the create and the edit path.

The row overflow menu, its column and the Add button are admin-only. On
`develop` those were `SecureRoute`s gated on `GLOBAL_ADMIN`; here they are
conditions in the markup. The storage write endpoints carry no role check of
their own, so this gate is the only role barrier on the create and edit paths,
and it is asserted against the shape of the condition rather than a list of
roles: a user holding every non-admin role at once sees none of the controls,
and an administrator who also holds Reception sees all of them, which a gate
written as an exclusion cannot satisfy.

### Paging and ordering

The tables show five rows a page, the smallest size the sample-items endpoint
accepts. That size turned three latent ordering problems into everyday ones, so
all three are fixed here.

The five level listings return every row and take no paging parameters, so the
page is cut client-side, ordered newest first, and returns to the first page
after a create or a delete. `GET /rest/storage/sample-items` is ordered by
sample item id descending rather than by location: descending because an
ascending order opens the tab on the oldest items a lab ever received, which on
a running site are long disposed; by id rather than location because disposing
an item clears its location, which used to move its row to another page instead
of flipping its status where it sat. The lot assignment query orders by lot id
descending, because it had no ordering at any layer and an unordered scan
returns an updated row wherever the database rewrote it.

Every table returns to the first page whenever the list under it shrinks, since
a stale page index points past the end of a shorter list and the table renders
empty. That covers create, delete, the storage search, and the Inventory
Catalog's search and its type and status filters.

The sample-items search endpoint returns the same `active` and `disposed`
values the listing does; it used to hand back the raw database status id, so a
disposed sample found by search wore the Active tag. A disposal the server
refuses now raises an error notification instead of leaving Confirm silent.

### Inventory Management

The four metric tiles filter the lots table beneath them, the way the Storage
Management level tiles do. Each metric's rule lives once and is read by the
tile's count, the filter it applies and the stock-status tag on the row, so an
edit to a rule reaches all three. Low Stock is the exception and says so: it
counts items below threshold, as the backend reports them, while filtering to
those items' lots, so its number and its row count differ by design. The search
box and the type and status dropdowns return to the first page as the tiles do,
and the dashboard lists lots newest first, matching the storage table.

All five level tables also gain a search box. Rooms, devices, shelves and racks
already had a search endpoint on `develop`; boxes did not, so this adds
`GET /rest/storage/boxes/search` alongside them.

Each dropdown in the Add modal asks for what it actually selects. All three were
labelled with `storage.picker.select`, whose English text is `Select {level}`
and which carries the same placeholder in every translated bundle, and all three
formatted it with no values, so all three rendered the pattern verbatim. A box's
rack picker is read-only, because the box update handler writes back the rack
the stored row already has, so an editable picker would report success and move
nothing.

`en.json` is the only bundle this PR touches, and Transifex owns the other 29.
The old child menu rows are deactivated rather than deleted, and the changeset
has a rollback.

The Playwright specs move with the UI. Six storage specs drove the old shell, so
they were rewritten against the new journeys rather than relaxed, and each still
proves what its name claims. Two guarantees genuinely disappear and are recorded
here rather than quietly dropped: the sidenav highlight can no longer move
between sibling storage rows, because one row is left, and there is no per-level
breadcrumb crumb to assert. Tab and tile state carry the first, asserted through
ARIA.

This PR also deletes five already-dead components with their stylesheets and
tests: `StorageDashboard`, `StorageLocationModal`, `LocationTreeView`,
`LocationFilterDropdown` and `LocationAutocomplete`. Nothing on `develop`
imported any of them before this branch existed; they are cleared out here
rather than in a separate janitorial PR because this is the last PR of the
storage stack, so that block of deletions reads as unrelated to lot storage.

## Screenshots

## Related Issue

Stack, reviewed bottom-up. This is the last one open; every PR below it is
merged.

5. ~~#4019 Correct stock visibility, barcodes and item codes~~ merged
6. ~~#4026 Make the inventory lot barcode findable~~ merged
7. **Lot storage parity, and a Storage Management rework** — you are here

## Other

Two things a reviewer meets immediately, neither introduced here:

**Creating a room with the Code field blank fails when the name is long.**
`generateUniqueRoomCode` builds a code of up to 45 characters while the code
validator caps it at 10, so a name like "Verify Room 4033" is rejected with
"Code cannot exceed 10 characters". This is on `develop` and this branch does
not touch it, but deleting the `/new` pages makes the modal the only way in.

**Manage Location is offered on a disposed sample item and cannot succeed.**
A disposed sample keeps its assignment row with the location cleared, so the UI
treats the action as an assign while the server sees an existing assignment and
refuses it. The error is shown and the modal closes; the action simply should
not be offered. Also pre-existing.

Follow-ups this PR does not take on: `/rest/storage/inventory-lots` accepts no
paging parameters, so the listing wants server-side pagination once the table
grows. Turning a page during a Sample Items search refetches the whole match
set, because the listing genuinely needs its paging parameters server-side and
only the search branch ignores them. An optimistic-lock conflict on a lot `PUT`
still surfaces as a 500 rather than a 409. `InventoryManagementServiceIT` is
named so that Surefire never runs it, since this repo configures no Failsafe.
`updateDevice` never copies `ipAddress`, `port` or `communicationProtocol` from
the form, unlike `createDevice`, so every device update nulls a linked freezer's
connection details. And the Inventory Dashboard and Catalog hide their pager
when a filter empties the table, which is what strands a user whose page index
outlived its list after a save; rendering the pager on an empty page would close
that class on both.


mherman22 (commented):
The description has drifted from the head, and it is the only spec available here (OGC-1177 is not publicly readable, so I judged the change against the description alone). It describes tabs called Overview / Locations / Sample Items / Inventory Lots with the five levels behind a ContentSwitcher; the page actually has three tabs, Dashboard / Sample Items / Inventory Lots, with the levels as clickable tiles. It says the level pages are "unchanged apart from an `embedded` prop", but each one gained a search box, an AddLocationModal and a notification helper, and the five Add pages and the dashboard page are gone entirely. It says every previous URL still resolves; the five `/Storage/*/new` URLs no longer do. Whoever picks this up to test it will be testing a page that does not exist, so the description is worth a rewrite.

Separately, this is two changes sharing a branch: the lot storage parity work (the disposal fix, the two endpoints, the lots page) and the Storage Management rework, which is where nearly all of the 4,100 deleted lines live. I did get through both, but the disposal fix is three lines sitting under forty UI files, and splitting at that seam would let the backend half be reviewed and land on its own timescale.

frontend/src/components/layout/Header.test.jsx
```diff
@@ -1025,3 +1025,85 @@ describe("Header Component - M2b Enhancement Tests", () => {
...
+
+    const parent = await waitFor(() => {
+      const el = container.querySelector("#menu_storage_management");
+      expect(el).toBeTruthy();
+      return el;
+    });
+
+    // Carbon renders an expandable parent as a button inside .cds--side-nav__submenu
+    expect(parent.querySelector(".cds--side-nav__submenu")).toBeNull();
+    expect(parent.querySelector('a[href="/Storage"]')).toBeTruthy();
```
mherman22, line 1087: Both of these tests fail on this head, which is what the red frontend checkpoint is. `#menu_storage_management` is not the sidenav row in the leaf branch: `Header.jsx` puts `elementId` on the inner `<span>` and gives the `SideNavMenuItem` anchor `elementId + "_nav"`. So `parent` here is the span, `.cds--side-nav__submenu` is null trivially, and the `a[href="/Storage"]` lookup finds nothing because the anchor is the span's ancestor, not its child. The second test fails the same way on the selector at line 1097.

If this merges as-is the guard the PR adds for the collapsed Storage menu is never actually asserted, so a later change to `hasChildren` can turn Storage Management back into an expandable that opens onto nothing and no test notices.

The production code is fine, for what it is worth. I re-pointed the query at `#menu_storage_management_nav` and asserted on the element itself and the file goes 37/37:

```js
    const parent = await waitFor(() => {
      const el = container.querySelector("#menu_storage_management_nav");
      expect(el).toBeTruthy();
      return el;
    });

    expect(parent.tagName).toBe("A");
    expect(parent.getAttribute("href")).toBe("/Storage");
```

The second test needs the same swap, querying `"#menu_storage_management_nav"` instead of `"#menu_storage_management a[href='/Storage']"`.

src/main/java/org/openelisglobal/inventory/controller/rest/InventoryLotRestController.java
```diff
@@ -377,6 +377,9 @@ public ResponseEntity<InventoryLot> disposeLot(@PathVariable String id,
             String reason = request != null ? request.getReason() : null;
             String notes = request != null ? request.getNotes() : null;
             InventoryLot lot = inventoryLotService.disposeLot(Long.valueOf(id), reason, notes, sysUserId);
+            sampleStorageService.releaseInventoryLotLocation(id,
+                    "Disposal: " + (reason != null ? reason : "") + (notes != null ? " | Notes: " + notes : ""),
+                    sysUserId);
```
mherman22, line 382: This belongs in the service before it merges. `disposeLot` is `@Transactional` and commits before `releaseInventoryLotLocation` opens a second transaction, so if the release throws, the lot is already DISPOSED while its assignment still points at the box. That is the phantom occupancy this PR exists to remove, it keeps inflating the capacity figures samples read, and the caller sees only a 500 that says nothing about the split state.

`disposeSampleItem` is the sibling to copy: it lives in `SampleStorageServiceImpl` and does the status change and the slot release inside one `@Transactional` method. That class already autowires `InventoryLotService` (field at line 65), so a `disposeInventoryLot(...)` next to it needs no new wiring and the controller goes back to one call.

The placement also leaves the fix untested. Nothing under `src/test/java/org/openelisglobal/inventory/controller/rest/` touches `/dispose`, and the new `SampleStorageServiceInventoryLotIntegrationTest` calls `releaseInventoryLotLocation` directly, so these three lines can be deleted and the suite stays green. Moving it into the service is what makes "disposing a lot frees its slot" assertable at all.

frontend/src/components/storage/components/AddLocationModal.jsx
```diff
@@ -0,0 +1,302 @@
...
+  const invalid = useMemo(() => {
+    if (!form.name.trim()) return true;
+    if (level === "device" && !form.deviceType) return true;
+    if (level === "box") {
+      if (!form.code.trim()) return true;
+      if (!(Number(form.rows) >= 1) || !(Number(form.columns) >= 1))
+        return true;
+    }
+    return false;
+  }, [form, level]);
```
mherman22, line 93: Racks and boxes lost their required-parent check in the fold. `AddBoxPage` disabled Add on `!formData.parentRackId`, and `AddLocationPage` on `(type === "rack" && !formData[meta.parentField])`; this `invalid` never looks at `form.parentId`, so Create is enabled with nothing selected and the user learns about it from a 400 after submitting.

`createDevice`, `createShelf`, `createRack` and `createBox` all reject a missing parent, so gating on `parentField` covers every level below room rather than just the two the old pages guarded:

```suggestion
  const invalid = useMemo(() => {
    if (!form.name.trim()) return true;
    if (storageLevel(level)?.parentField && !form.parentId) return true;
    if (level === "device" && !form.deviceType) return true;
    if (level === "box") {
      if (!form.code.trim()) return true;
      if (!(Number(form.rows) >= 1) || !(Number(form.columns) >= 1))
        return true;
    }
    return false;
  }, [form, level]);
```

frontend/src/components/storage/StorageManagementPage.jsx
```diff
@@ -0,0 +1,223 @@
...
+                  <LevelTable embedded />
+                </TabPanel>
+
+                <TabPanel>
+                  <SampleItemsPage embedded />
+                </TabPanel>
+
+                <TabPanel>
+                  <InventoryLotsPage embedded />
+                </TabPanel>
```
mherman22, line 215: Both of these mount on every visit to `/Storage`, including when the Dashboard tab is the one showing. Carbon's `TabPanel` only toggles `hidden`; it renders its children unconditionally (`@carbon/react` 1.104.1, `Tabs.js` line 739). I rendered the page at `/Storage` and it issues four requests: `dashboard/location-counts`, `rooms?page=0&size=25`, `sample-items?page=0&size=25` and `inventory-lots?page=0&size=25`. The last two are for tabs nobody is looking at, and `sample-items` is `getAllSamplesWithAssignments`, which walks every sample item in the database.

Rendering each panel only once its tab has been selected would keep the landing page down to one list. Worth noting the page's own test cannot catch this, since it mocks all seven child pages out.

frontend/src/components/storage/pages/InventoryLotsPage.jsx
```diff
@@ -0,0 +1,241 @@
...
+  const [page, setPage] = useState(1);
+  const [pageSize, setPageSize] = useState(25);
+  const [searchTerm, setSearchTerm] = useState("");
+
+  const { items, totalItems, loading } = useStorageTableData({
+    listUrl: "/rest/storage/inventory-lots",
+    page,
+    pageSize,
+    searchTerm,
+  });
```
mherman22, line 38: The comment at line 99 says a round trip per keystroke is the thing this avoids, but passing `searchTerm` into `useStorageTableData` is what causes one. `buildUrl` lists `searchTerm` in its deps, so `fetchData` gets a fresh identity on every character and the effect refires. I typed four characters into the search box in a render test and got four extra GETs, every one of them `/rest/storage/inventory-lots?page=0&size=25`, identical to the first. The filtering below is client-side and does not need any of them.

```suggestion
  const { items, totalItems, loading } = useStorageTableData({
    listUrl: "/rest/storage/inventory-lots",
    page,
    pageSize,
  });
```

frontend/src/components/storage/pages/StorageResourcePage.jsx
```diff
@@ -123,17 +134,46 @@ export default function StorageResourcePage({
...
+            id="storage-resource-search"
+            size="md"
+            labelText={intl.formatMessage({
+              id: searchPlaceholderId || "label.search",
+              defaultMessage: "Search",
+            })}
+            placeHolderText={intl.formatMessage({
+              id: searchPlaceholderId || "label.search",
+              defaultMessage: "Search",
+            })}
```
mherman22, line 164: `placeHolderText` is not a prop on Carbon's `Search`, which takes `placeholder`, so none of these translated placeholders reach the user. I rendered `<Search placeHolderText="SHOULD-SHOW" />` against `@carbon/react` 1.104.1 and the input came back with `placeholder="Search"` (Carbon's default) plus a stray `placeholdertext` attribute leaked onto the DOM by the rest-spread. So `storage.search.rooms.placeholder` and its four siblings, and the `storage.search.boxes.placeholder` this PR adds, all render as the generic default. Every other search box in the codebase uses `placeholder=`. `InventoryLotsPage.jsx:174` and `SampleItemsPage.jsx` carry the same typo.

```suggestion
            placeholder={intl.formatMessage({
              id: searchPlaceholderId || "label.search",
              defaultMessage: "Search",
            })}
```

frontend/src/App.jsx
```diff
@@ -928,17 +920,20 @@ export default function App() {
...
                     <RouteErrorBoundary {...routeErrorStorage}>
-                      <StorageDashboard />
+                      <StorageManagementPage />
                     </RouteErrorBoundary>
                   )}
                   role={[Roles.RECEPTION, Roles.RESULTS, Roles.GLOBAL_ADMIN]}
                 />
+                {/* Every per-resource URL resolves to the same tabbed page, so
+                    existing bookmarks and menu rows keep working. The slug union
+                    keeps /new and /:id/edit matching their own routes. */}
```
mherman22, line 930: The `/new` half of the last sentence is not true any more: this PR deletes all five `/Storage/*/new` routes, so those URLs fall through instead of matching a route of their own. Nothing links to them (I grepped the tree and the only `/Storage/...` hrefs left are the listing and `:id/edit` ones), so the behaviour is fine, but the comment will send the next reader hunting for routes that do not exist. Dropping the last sentence is enough.

frontend/src/components/storage/storageLevels.js
```diff
@@ -0,0 +1,83 @@
+export const STORAGE_LEVELS = [
```
mherman22, line 1: The fold dropped the one comment that explained why `listParam` and `parentField` both exist. `CreateForm.jsx` carried it above its own `LEVELS`:

```
// Backend naming inconsistency captured here so callers don't branch:
//   - GET list parent param: "{parent}Id"   (e.g. roomId, deviceId)
//   - POST body parent field: "parent{Parent}Id" (e.g. parentRoomId)
//   - Identifier field: "name" for Room/Device, "label" for Shelf/Rack/Box
```

Without it the table reads as if it holds two names for one thing, which is the edit that comment was there to prevent. Worth bringing across now that this is the single source for all five callers.

frontend/src/components/storage/storageLevels.js
```diff
@@ -0,0 +1,83 @@
...
+    addTitleId: "storage.add.box",
+    addTitleDefault: "Add Box",
+  },
+];
+
+export const storageLevel = (key) =>
+  STORAGE_LEVELS.find((level) => level.key === key) || null;
+
+/** Keys in hierarchy order, e.g. for cascading clears. */
+export const STORAGE_LEVEL_KEYS = STORAGE_LEVELS.map((level) => level.key);
```
mherman22, line 83: `STORAGE_LEVEL_KEYS` has no callers. The five level pages, `EditLocationPage`, `AddLocationModal` and `CreateForm` all go through `storageLevel()` or `STORAGE_LEVELS` directly, and the cascading-clear use the comment describes does not exist anywhere yet. I would drop both lines and add them back alongside the caller that needs them.

frontend/src/components/storage/StorageManagementPage.jsx
```diff
@@ -0,0 +1,223 @@
...
+import { AlertDialog } from "../common/CustomNotification";
+import { getFromOpenElisServer } from "../utils/Utils";
+import SampleItemsPage from "./pages/SampleItemsPage";
+import InventoryLotsPage from "./pages/InventoryLotsPage";
+import RoomsPage from "./pages/RoomsPage";
+import DevicesPage from "./pages/DevicesPage";
+import ShelvesPage from "./pages/ShelvesPage";
+import RacksPage from "./pages/RacksPage";
+import BoxesPage from "./pages/BoxesPage";
+import "./StorageDashboard.css";
```
mherman22, line 25: Deleting `StorageDashboard.jsx` orphaned most of this stylesheet. Nothing in the tree carries the `storage-dashboard` class any more, so 26 of its 33 selectors are dead, and this import now exists for the `.storage-metric*` rules appended at the end (`.metric-value` at line 45 is the one unscoped survivor that still applies to the tiles). Pruning the dead rules and renaming the file after the page that owns it would save the next editor the trace I just did.

frontend/src/languages/en.json
```diff
@@ -6911,6 +6911,8 @@
   "storage.nav.dashboard": "Storage Dashboard",
   "storage.nav.devices": "Devices",
   "storage.nav.devices.tooltip": "Manage storage devices",
+  "storage.nav.inventoryLots": "Inventory Lots",
+  "storage.nav.inventoryLots.tooltip": "View inventory lots and where they are stored",
```
mherman22, line 6915: These two were added for the Inventory Lots sidenav row, and a later commit in this PR replaced `073-storage-inventory-lots-menu.xml` with the collapse changeset, so the row is never created and nothing reads either key. The page labels itself from `storage.tab.inventoryLots` and `storage.breadcrumb.inventorylots` instead. Same leftover in `073-storage-menu-collapse.xml`: `menu_storage_inventory_lots` is in the `is_active` update but matches no row in any changeset, and it is missing from the rollback's list, so the two are worth cleaning together.

frontend/src/components/storage/pages/RoomsPage.jsx
```diff
@@ -1,17 +1,49 @@
...
   const [pageSize, setPageSize] = useState(25);
   const [deleteTarget, setDeleteTarget] = useState(null);
+  const [addOpen, setAddOpen] = useState(false);
+  const { setNotificationVisible, addNotification } =
+    useContext(NotificationContext);
+
+  // Name the level being acted on — "Rack created", not a generic
+  // "Storage location created" that reads identically for all five.
+  const level = storageLevel("room");
+  const notify = (kind, messageId, defaultMessage) => {
```
mherman22, line 26: This `notify` helper is byte-identical in all five level pages. I hashed the block in RoomsPage, DevicesPage, ShelvesPage, RacksPage and BoxesPage and got the same md5 from each, and the `AddLocationModal` plus `history.replace` block further down is the same story. Everything in it is already parameterised by `storageLevel(key)`, so the next change to the create or delete feedback is five edits rather than one, in a PR whose point is to stop repeating these five pages.

`StorageResourcePage` already receives `onAddRequested` and `onDeleteRequested`, so it is the natural home; a small `useLocationLevelFeedback(levelKey)` hook would do just as well if you would rather keep the modal in the page.

frontend/src/components/storage/components/EditLocationModal.jsx
```diff
@@ -0,0 +1,358 @@
...
+    setForm(null);
+    setError(null);
+    setLoading(rowKey !== null);
+  }
+
+  useEffect(() => {
+    if (!open || !meta || id == null) return;
+    getFromOpenElisServer(
+      `/rest/storage/${meta.endpoint}/${encodeURIComponent(String(id))}`,
+      (response) => {
```
mherman22, line 58: This needs a guard before it merges. The callback writes into the form without checking that the row it was issued for is still the one open, and the effect has no cleanup, so a response can land after the dialog was closed and reopened on a different row.

If merged as-is, an admin who opens Edit on row A, closes it, and opens Edit on row B before A's GET returns sees A's name and code sitting in B's form, and pressing Save PUTs those values to row B, silently renaming and re-coding the wrong location with no error shown. I reproduced it by firing row B's callback and then row A's: the Name field held row A's value while the `id` prop was already B.

`getFromOpenElisServer` takes an `AbortSignal` as its third argument, and thirteen components already use it for exactly this; `HistoricalTrends.jsx:118` spells the idiom out in a comment. Capturing `rowKey` in the closure and dropping the response when it no longer matches works too, and has the advantage of also covering the `setLoading(false)` below, which today clears the spinner for a row whose own request has not come back.

The pages this replaces did not have the problem, since each edit was a fresh route mount.

frontend/src/components/storage/components/EditLocationModal.jsx
```diff
@@ -0,0 +1,358 @@
...
+        if (response && !response.error) {
+          setForm({
+            name:
+              response[meta.nameField] || response.name || response.label || "",
+            code: response.code || "",
+            description: response.description || "",
+            parentId: meta.parentField
+              ? String(response[meta.parentField] || "")
```
mherman22: At the shelf level this never preselects. `meta.parentField` is `parentDeviceId`, but the shelf GET does not return that key: `StorageShelfResponse` exposes `deviceId` (grep gives 4 hits for `deviceId`, 0 for `parentDeviceId`), and `toShelfResponse` calls `setDeviceId`. So the Device picker on an Edit Shelf modal reads "Select device" even when the shelf sits in one. Saving is unaffected, because `updateShelf` treats a blank parent as "keep the existing one", so this is cosmetic.

It came across verbatim from `EditLocationPage` rather than being introduced here, but this is the only edit path now. `response[meta.parentField] || response.deviceId` covers it. Device, rack and box all return their own `parent*Id` and are fine.

frontend/src/components/storage/components/EditLocationModal.jsx
```diff
@@ -0,0 +1,358 @@
...
+          intl.formatMessage({
+            id: "storage.edit.error.saveFailed",
+            defaultMessage: "Save failed",
+          }),
+        );
+      }
+      if (!response.ok) {
+        const body = await response.json().catch(() => ({}));
+        throw new Error(
+          body.error ||
```
mherman22, line 162: Bean-validation failures miss this chain and reach the user as "Save failed (HTTP 400)". `ControllerSetup.handleMethodArgumentNotValid` returns `{timestamp, status, errors: {field: message}}`, so neither `error` nor `message` matches and the server's own wording is thrown away. It is reachable from here because Save is not gated on a non-empty name, so clearing Name and saving tells the user nothing about what went wrong.

Preferring `body.error` is the right call and is a real improvement on the deleted pages, since every explicit failure the storage controller raises uses that key. Adding `body.errors` to the chain, or gating Save on a non-empty name, closes what is left.

frontend/src/components/storage/components/EditLocationModal.jsx
```diff
@@ -0,0 +1,358 @@
...
+                  id: "label.description",
+                  defaultMessage: "Description",
+                })}
+                value={form.description}
+                onChange={(e) => update("description", e.target.value)}
+              />
+            )}
+
+            {/* Boxes render their rack read-only: updateBox reads the parent off
+                the stored row and never calls setParentRack, so an editable
```
mherman22: `updateBox` does call `setParentRack`, at `StorageLocationRestController.java:1295`, using the parent read off the stored row. The conclusion is right, the submitted `parentRackId` really is ignored, but the mechanism as stated is not, and whoever greps `setParentRack` after reading this will end up doubting the rest of the comment. Saying the submitted rack is ignored is both shorter and true.

frontend/src/components/storage/pages/StorageResourcePage.jsx
```diff
@@ -70,69 +113,88 @@ export default function StorageResourcePage({
...
-      if (editHref) {
-        nextRow.actions = (
-          <Link to={editHref(rawItem)}>
-            <FormattedMessage id="label.edit" defaultMessage="Edit" />
-          </Link>
-        );
-      }
-      if (editHref || (isGlobalAdmin && onDeleteRequested)) {
-        nextRow.menuActions = (
+      if (!isGlobalAdmin) return row;
```
mherman22, line 110: The two behaviour changes in this file, the menu becoming admin-only and the update notification, have no test that would fail without them. Every new test renders the modal directly instead. `RacksPage.test.jsx` already has the creation analogue asserting "Rack created", so an update counterpart plus a case for a non-admin getting no actions column would close it.

No Playwright spec covers editing a device, shelf or rack either, only rooms and boxes. A shelf one would have caught the preselect problem I left on the parent read above.

frontend/src/components/storage/components/EditLocationModal.jsx
```diff
@@ -0,0 +1,358 @@
...
+    setError(null);
+
+    const payload = {
+      [meta.nameField]: form.name,
+      code: form.code || null,
+      active: form.active,
+    };
+    if (meta.parentField) payload[meta.parentField] = form.parentId || null;
+    if (level === "room") payload.description = form.description || null;
+    if (level === "device") payload.type = form.deviceType || null;
```
mherman22: Was it deliberate to leave the device-only fields out of this payload? `updateDevice` assigns `temperatureSetting` and `capacityLimit` straight from the form, with an explicit `: null` when the form omits the temperature, so editing a device's name through this modal clears both. `updateShelf` does the same to `capacityLimit`, and `updateBox` to `type` and `positionSchemaHint`.

`EditLocationPage` and `EditBoxPage` left them out too, so this commit does not introduce it and I would not hold the PR for it. I am asking because the modal is now the only way to edit a device, and the loss is silent.

PR #4084, opened 2026-08-18
fix: resolve acting user properly in storage module controllers

### LINKED ISSUE #4082: Storage module: hardcoded/inconsistent user ID resolution in LabelManagementRestController and StorageLocationRestController
LabelManagementRestController.getCurrentUserId() was just a stub:

private String getCurrentUserId() {
    // Placeholder: should get from Spring Security context
    // For now, return default system user
    return "1";
}

This value gets passed into trackPrintHistory(...), so every label print gets recorded as user 1 no matter who actually printed it.

Separately, StorageLocationRestController was still using the old getSysUserId(request) pattern that #3356 replaced everywhere else. This one wasn't hardcoded or broken exactly, just inconsistent with the rest of the module — 15 methods were passing around an HttpServletRequest just to feed it into checkAdminStatus(request).

After fixing:

- LabelManagementRestController.getCurrentUserId() resolves the real user, and if it can't (no auth context), the print endpoint returns 401 instead of silently writing a bad value into the print history table
- StorageLocationRestController.checkAdminStatus() and the 6 places that stamp sysUserId on new rooms/devices/shelves/racks/boxes should use the no-arg getSysUserId() like the rest of the module already does, and the unused HttpServletRequest param goes from all 15 methods

Working on this, will put up a PR shortly.

kavya685:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
Closes #4082

`LabelManagementRestController.getCurrentUserId()` was a stub that always returned `"1"`, so every label print got recorded in print history under the system user instead of whoever actually printed it. It now resolves the real user through `UserContextHolder`, and the print endpoint returns 401 if there's no auth context instead of silently writing a bad value.

`StorageLocationRestController` wasn't hardcoded, but it was still on the old `getSysUserId(request)` pattern that #3356 replaced everywhere else — 15 methods carrying an `HttpServletRequest` around just to feed `checkAdminStatus(request)`. Switched it over to the no-arg `getSysUserId()` like the rest of the module, and dropped the now-unused `HttpServletRequest` param from all 15 signatures.

## Screenshots

## Related Issue
Fixes #4082

## Other

### DIFF
--- a/src/main/java/org/openelisglobal/storage/controller/LabelManagementRestController.java
+++ b/src/main/java/org/openelisglobal/storage/controller/LabelManagementRestController.java
@@ -6,6 +6,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import org.openelisglobal.common.rest.BaseRestController;
+import org.openelisglobal.common.util.UserContextHolder;
 import org.openelisglobal.storage.dao.StorageDeviceDAO;
 import org.openelisglobal.storage.dao.StorageRackDAO;
 import org.openelisglobal.storage.dao.StorageShelfDAO;
@@ -43,6 +44,9 @@ public class LabelManagementRestController extends BaseRestController {
     @Autowired
     private StorageRackDAO storageRackDAO;
 
+    @Autowired
+    private UserContextHolder userContextHolder;
+
     /**
      * Generate and return PDF label POST /rest/storage/{type}/{id}/print-label
      * Validates code exists before printing, returns error if missing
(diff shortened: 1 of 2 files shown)


Josephkagimu1 (commented):
Hi @kavya685  , The fix for getCurrentUserId() is correct — UserContextHolder.requireSysUserId() with a 401 guard is the right pattern. The StorageLocationRestController cleanup is clean and consistent with PR #3356.

One gap: no test for the behavioral change in LabelManagementRestController. Since getCurrentUserId() went from hardcoded "1" to the real authenticated user, a test verifying print history is recorded under the correct user ID (and returns 401 when no auth context) would prevent regressions.

PR #4111, opened 2026-08-20
fix: protect label_preset from fixture-load truncation in test suite

### LINKED ISSUE #3699: Refactor: Standardise `BaseWebContextSensitiveTest` — Phase 1: Bug Fixes

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
When the full test suite ran, another test class (e.g. LabelServiceTest) called executeDataSetWithStateManagement, which truncated the label_preset table as part of its fixture setup. This wiped the Liquibase-seeded "Specimen Label" system preset (including its is_universal = true flag backfilled by changeset 032). The aggregation test then ran and couldn't find the preset, causing the failure.

## Screenshots

Screenshot from 2026-08-21 01-50-56

## Related Issue

Closes https://github.com/DIGI-UW/OpenELIS-Global-2/issues/3699


src/test/java/org/openelisglobal/BaseWebContextSensitiveTest.java
```diff
@@ -95,7 +95,15 @@ public abstract class BaseWebContextSensitiveTest extends AbstractTransactionalJ
      * {@code nextval()} returns {@code 4}) as it does for every sequence in the
      * schema on a fresh test DB.
      */
-    private static final String[] PROTECTED_SEED_TABLES = { "reference_tables", "requester_type" };
+    // label_preset is protected so the Liquibase-seeded system presets (Order
+    // Label,
```
HerbertYiga: do we need a list of all these added comments?

HerbertYiga (approved):

PR #4153, opened 2026-09-01
fix(userrole): correct misleading error messages and widen exception handling in UserRoleDAOImpl

### LINKED ISSUE #2698: Improve error handling and diagnostics in UserRoleDAOImpl – silent failures hide real problems
### Describe the problem / current behavior

Many methods in `org.openelisglobal.userrole.daoimpl.UserRoleDAOImpl` catch exceptions (HibernateException / RuntimeException), log them via `LogEvent.logError(e)`, and then either:

- silently return partial/empty results, or
- re-throw a generic `LIMSRuntimeException` **without any contextual information** (userId, roleName, etc.)

This makes it very hard to diagnose permission / role lookup failures in production.

Examples of problematic patterns:

1. `getRoleIdsForUser(String userId)`
   - Catches → logs → returns empty list → caller has no idea why user appears to have no roles

2. `userInRole(String userId, String roleName)` and the Collection variant
   - Catch → log → return `false` → looks like user simply doesn't have the role

3. `deleteLabUnitRoleMap(LabUnitRoleMap roleMap)`
   - Silent delete failure possible

When something goes wrong (DB connectivity, cast issues, constraint violations, etc.), support tickets become very hard to investigate because the only clue is a generic log line without context.

### Why this matters

- Security-sensitive area (roles & permissions)
- Silent failures lead to mysterious symptoms: "user can't access feature X but has the role assigned"
- Wastes a lot of time for implementers / support teams
- Harder to create good bug reports / tickets from the field

### Suggested improvement

Add **contextual information** to both logs **and** the thrown `LIMSRuntimeException` so that:

- Logs are immediately actionable
- Exceptions contain enough info to understand what failed without reading code

dharmendra26-wiz:
Fixes #2698 (partially — scoped to UserRoleDAOImpl's DAO layer, per @mozzy11's feedback on the prior attempt in #2701, which confirmed the issue was still valid on current develop and asked for a fresh PR with a green build and tests).

## What was wrong
While investigating #2698, I found two copy-paste bugs and one exception-handling gap in `UserRoleDAOImpl`, verified by tests rather than just reading the code:

1. `deleteLabUnitRoleMap()` and `getUserIdsForRole()` both had error messages that incorrectly said `"Error in UserRoleDAOImpl userInRole()"` — copy-pasted from the `userInRole()` methods and never updated. This would send anyone debugging a real failure to the wrong method.

2. `deleteLabUnitRoleMap()` only caught `HibernateException`, but deleting a detached/non-existent entity through `EntityManager` throws `jakarta.persistence.EntityNotFoundException`, which does not extend `HibernateException`. This meant a real failure case bypassed the logging/wrapping entirely and propagated as a raw, unlabeled exception. Widened the catch to `RuntimeException`.

## What changed
- Fixed both incorrect error message strings to reference their actual method names.
- Widened `deleteLabUnitRoleMap()`'s catch clause from `HibernateException` to `RuntimeException` to also cover `EntityNotFoundException`.
- Added `UserRoleDAOImplTest` with a test that reproduces the `EntityNotFoundException` path (via a Hibernate proxy to a non-existent row) and asserts the wrapped exception now correctly references `deleteLabUnitRoleMap()`.

## Testing
- `mvn test -Dtest=UserRoleDAOImplTest` — 1/1 pass
- `mvn test -Dtest=UserRoleServiceTest,UserRoleDAOImplTest` — 8/8 pass, confirming no regression in existing UserRole tests

This is a smaller, more targeted scope than #2701 — no behavior change for existing callers (both methods still throw `LIMSRuntimeException`, just now with correct messages and broader coverage), so it should be low-risk to review.

### DIFF
--- a/src/main/java/org/openelisglobal/userrole/daoimpl/UserRoleDAOImpl.java
+++ b/src/main/java/org/openelisglobal/userrole/daoimpl/UserRoleDAOImpl.java
@@ -104,9 +104,9 @@ public class UserRoleDAOImpl extends BaseDAOImpl<UserRole, UserRolePK> implement
     public void deleteLabUnitRoleMap(LabUnitRoleMap roleMap) {
         try {
             entityManager.unwrap(Session.class).delete(roleMap);
-        } catch (HibernateException e) {
+        } catch (RuntimeException e) {
             LogEvent.logError(e);
-            throw new LIMSRuntimeException("Error in UserRoleDAOImpl userInRole()", e);
+            throw new LIMSRuntimeException("Error in UserRoleDAOImpl deleteLabUnitRoleMap()", e);
         }
     }
 
@@ -121,7 +121,7 @@ public class UserRoleDAOImpl extends BaseDAOImpl<UserRole, UserRolePK> implement
             userIds = query.list();
         } catch (HibernateException e) {
             LogEvent.logError(e);
-            throw new LIMSRuntimeException("Error in UserRoleDAOImpl userInRole()", e);
+            throw new LIMSRuntimeException("Error in UserRoleDAOImpl getUserIdsForRole()", e);
         }
         return userIds;
     }
--- a/src/test/java/org/openelisglobal/userrole/UserRoleDAOImplTest.java
+++ b/src/test/java/org/openelisglobal/userrole/UserRoleDAOImplTest.java
@@ -1,0 +1,48 @@
+package org.openelisglobal.userrole;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.PersistenceContext;
+import org.hibernate.Session;
+import org.junit.Test;
+import org.openelisglobal.BaseWebContextSensitiveTest;
+import org.openelisglobal.common.exception.LIMSRuntimeException;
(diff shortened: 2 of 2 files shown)


src/main/java/org/openelisglobal/userrole/daoimpl/UserRoleDAOImpl.java
```diff
@@ -121,7 +121,7 @@ public List<String> getUserIdsForRole(String roleName) {
             userIds = query.list();
         } catch (HibernateException e) {
             LogEvent.logError(e);
-            throw new LIMSRuntimeException("Error in UserRoleDAOImpl userInRole()", e);
+            throw new LIMSRuntimeException("Error in UserRoleDAOImpl getUserIdsForRole()", e);
```
mherman22: There's a third one of these in the same file that this didn't catch. Line 55, inside `getRoleIdsForUser`, throws `Error in UserRoleDAOImpl getUserRolesForUser()`, and `getUserRolesForUser` doesn't exist anywhere in the tree, so it points a debugger at something that isn't even a method. Same one line change as the two here.

Worth folding in rather than leaving, since the next reader will assume a PR titled "correct misleading error messages" got all of them.

dharmendra26-wiz:
"Hey @Agaba-derrick , the CI checks are all green and the branch is up to date. Could you please take a look and review this when you have a chance? Thank you!"

PR #4157, opened 2026-09-02
test(eqa-v2): Playwright coverage for the participant and provider EQA lanes (OGC-613)

samuelmale:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Playwright end-to-end coverage for the EQA module. Tests only — no production
code changes, no liquibase, no new i18n keys, no new dependencies, and no
edits to shared config (the `foundational/core` glob picks new specs up, and
the setup project already matches any setup file).

Before this, three EQA tests existed and touched four of the module's eighteen
routes. This adds fourteen tests across nine specs, and every V2 surface a
provider, participant or QA officer uses is now driven in a browser.

**Participant lane** — `eqa-participant-smoke.spec.ts`: a planned cycle is
received through Add Order with the EQA box (which records the panel receipt
and moves the cycle in one transaction), results are entered and validated,
the live progress count and entry tag are asserted, the cycle is reviewed and
submitted, and the order appears on the EQA orders register with a derived
status of complete.

**Provider lane** — `eqa-provider-cycle.spec.ts`: scheme board, the five-step
wizard (including the distribution-date and submission-deadline labels and the
pre-selected roster), the prep gate blocked then cleared, pack list and label
documents, dispatch to five participants, five deliveries — after which the
cycle **opens submissions with no human action**, the complementary edge to
`eqa-open-submissions.spec.ts`, which owns the manual override. Then scoring,
the CSV and FHIR score egress paths, a repeat panel from the reserve, report
comments attached and removed, the timeline actor rows, and the follow-up
register showing the participant that failed.

**Oversight** — `eqa-followup-split.spec.ts` asserts each follow-up register
holds only its own rows (a row about another lab must never enter this lab's
non-conformity workflow) and drives triage on both.
`eqa-oversight-dashboards.spec.ts` covers coverage, recent cycles and analyst
competency — the pages an accreditation assessor is shown.
`eqa-receipt-conditions.spec.ts` covers overdue and arrived-damaged, neither
of which a healthy dispatch produces.

**Also** — `eqa-enrollment-crud.spec.ts` (the write path other specs were
seeding around), `eqa-role-gating.spec.ts` (the participant-only board, via a
second authenticated session without the provider grant),
`eqa-in-house-unblind.spec.ts` (sealed targets and the label sheet), and
`eqa-v1-routes.spec.ts` (shallow mount checks for the five first-generation
screens still on the menu).

## Screenshots

Not applicable — this PR adds tests and changes no UI.

## Related Issue

OGC-613.

## Other

**Verification.** `npm run lint`, `npm run pw:guard` and `npm run format` are
clean. The full EQA suite ran green three times consecutively (16/16, ~1.6
minutes each) and leaves no seeded rows behind.

That took finding the cause of what looked all session like flaky tests. The
Docker VM was out of memory — 9.2 GB of 9.9 used, 491 MB available, swap
100% full — because a second full OpenELIS stack was running alongside the
first. Under Playwright load the webapp container restarted every few
minutes, taking down whatever was mid-run; the earlier full-suite attempt lost
83 tests that way, including untouched pre-existing specs. With the second
stack paused, available memory went to 3.9 GB, restarts stopped entirely, and
the suite time fell from 7.7 minutes to 1.6. **CI runs a dedicated stack, so
this should not recur there — but it is worth watching this PR's own run.**

**Seeding notes** for whoever maintains these. Statistics pool every result in
a distribution rather than grouping per test, so with n rows of which n-1 are
identical the odd one out scores exactly (n-1)/sqrt(n): fifteen rows reach
3.61 and an unacceptable verdict, where five cap at 1.79 and can never leave
acceptable. Participant results are unique on round, enrollment and analyte
together. The competency dashboard gets its own analyst because this stack's
admin already carries eleven samples and nine failures. And the follow-up
queue needs an organization named for the site — the backend resolves this
lab from the SiteName setting, falling back to "This laboratory", and returns
nothing when no organization carries that name, in which case every follow-up
lands in the provider register and the queue can never populate.

### Defects found, none fixed here

1. **A cycle can never be closed.** No control on either the provider or the
   participant side issues the SCORED to CLOSED transition, and no scheduler
   does either — the edge is legal in both state machines and reachable only
   by calling the transition endpoint directly. Every cycle therefore stays
   scored forever, and the scheme board keeps counting it as open.
2. **In-house unblinding returns 500.** The service reads the panel's cycle
   after the row-locking call that fetched the panel has returned, so the
   entity is detached and Hibernate cannot initialise the association. This
   happens before any seeded value is read. The spec asserts the action is
   offered but does not drive it, so it reports the bug instead of pinning it.
3. **The receipts tab shows stale rows.** `ReceiptMonitor` fetches on mount
   only while every workbench tab panel mounts with the page, so after a
   dispatch it still reports every participant as not shipped until a reload —
   which a provider would read as a failed dispatch. Adding `cycleStatus` to
   that callback's dependencies fixes it at the root.
4. **A non-array reply crashes the in-house page.** Panel target values are
   stored encrypted, and a value that fails to decrypt makes the panels
   endpoint answer 500; the page guards only against a null reply, so the
   error body reaches its tile arithmetic and the page dies with an unhandled
   type error. The same `data || []` shape appears on sibling EQA pages.

Deliberately left uncovered, with reasons: the four-step in-house wizard (its
gates are unit-tested against the rules the server enforces, and driving it
needs fixture data not guaranteed outside this stack); the deadline digest
(scheduler-driven, so integration tests own it); and scoring statistics
(covered by backend tests).

### DIFF
--- a/frontend/playwright/helpers/seed-eqa-data.ts
+++ b/frontend/playwright/helpers/seed-eqa-data.ts
@@ -311,11 +311,26 @@ export function seedProviderScheme(runTag: string): ProviderSchemeSeed {
   };
 }
 
+/** Reported results planted per organization, so a spec can assert the score
+ * cell ("{unacceptable} unacceptable of {RESULTS_PER_ORGANIZATION}"). */
+export const RESULTS_PER_ORGANIZATION = 3;
+
 /**
- * Plant the score container: one eqa_distribution on the cycle plus one
- * reported eqa_result per organization, clustered around 100 so every
- * verdict is ACCEPTABLE. scoreCycle counts non-null result_value rows on the
- * cycle's distribution and refuses below MIN_PARTICIPANTS_FOR_STATS (5).
+ * Plant the score container: one eqa_distribution on the cycle plus reported
+ * eqa_result rows, three tests per organization. scoreCycle refuses below
+ * MIN_PARTICIPANTS_FOR_STATS (5) reported rows, so five organizations give a
+ * comfortable fifteen.
+ *
+ * The first organization's first result is planted far from the pack, which
+ * makes it UNACCEPTABLE and every other row ACCEPTABLE. The arithmetic is
+ * worth stating, because a smaller seed cannot reach the verdict at all:
+ * statistics pool every result in the distribution (not per test), so with n
+ * rows of which n-1 are identical the odd one out scores exactly
+ * (n-1)/sqrt(n) whatever its magnitude — 3.61 at n=15, over the
+ * unacceptable threshold of 3.0, where five rows would cap at 1.79 and never
+ * leave ACCEPTABLE. One unacceptable participant is what makes scoring
+ * enqueue a follow-up, which is the behaviour under test.
+ *
  * Rows are swept by the scheme-scoped restore().
  */
 export function seedReportedResults(
@@ -328,13 +343,21 @@ export function seedReportedResults(
     "scheme id",
   );
   // eqa_result.test_id references test(id) — the panel's own rows key on
-  // analyte, not test — so borrow the lowest existing test the way
-  // seed-qc-sigma-data.ts does. Which test it is does not matter: the row
-  // only has to satisfy the FK and the (distribution, org, test) key.
-  const testId = asInt(
-    psql(`SELECT id FROM ${SCHEMA}.test ORDER BY id LIMIT 1`),
-    "test id",
-  );
+  // analyte, not test — so borrow existing tests the way
+  // seed-qc-sigma-data.ts does. Which tests they are does not matter: the
+  // rows only have to satisfy the FK and the (distribution, org, test) key.
+  const testIds = psql(
+    `SELECT string_agg(id::text, ',') FROM (SELECT id FROM ${SCHEMA}.test ORDER BY id` +
+      ` LIMIT ${RESULTS_PER_ORGANIZATION}) t`,
+  )
+    .split(",")
+    .map((id) => asInt(id, "test id"));
+  if (testIds.length < RESULTS_PER_ORGANIZATION) {
+    throw new Error(
+      `Need ${RESULTS_PER_ORGANIZATION} tests in the catalog, found ${testIds.length}`,
+    );
+  }
+
   const distributionId = asInt(
     psql(
       `INSERT INTO ${SCHEMA}.eqa_distribution (id, fhir_uuid, eqa_program_id, distribution_name,` +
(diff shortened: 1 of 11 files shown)


samuelmale:
Thanks — this was a good catch list. Nine of the nine blocking findings hold; eight are fixed, one is answered differently than suggested and I've explained why. Two of the findings turned out to be worse than reported, and chasing them surfaced three product defects.

## Fixed

**1. Dismissal recorded no competency event.** Correct, and worse: `EQAProviderScoringServiceImpl.snapshotRows` omits `participantResultId` entirely, so *every* provider-register dismissal writes zero events — only the participant path supplies it. The seed now plants a real `eqa_participant_result` with an assigned analyst and names its id in the summary, and the spec asserts the competency history shows "Dismissed — equipment" / "Excused". Verified by inversion: removing the id from the summary makes the spec fail.

**2. Unblinding untested, sealed target unprovable.** Correct on both halves. The panel is now created through `POST /rest/eqa/panels`, so the target passes through the encrypting converter, and confidentiality is asserted from two sessions — a reader holding the unblind privilege gets the target back, a bench user gets the blind code with the target withheld. That second assertion is only meaningful because the first proves the same call does return it. Unblinding is now `test.fixme` with the steps written out, not a button-exists check.

**3. FHIR failure accepted as success.** Fixed — the assertion requires "Scores returned over FHIR." It passes on this stack.

**4. Prep gate never proven blocked.** Fixed — ready-to-ship is asserted disabled with both blockers named ("needs N aliquots, has 0", "has not passed homogeneity QC") before any data is supplied, and enabled only after.

**6. Participant journey ran as administrator.** Correct, and this was the most valuable finding. Running it as a bench user immediately exposed two blocks: **result validation answers 401** and **the transition behind "Review & submit" answers 403**, because that endpoint requires `qa.manage.eqa`, which Reception and Results do not hold. The journey is now split by session — the laboratory receives the panel, records the receipt and enters results; validation and submission run as someone who holds those privileges. Enrollment CRUD also runs as the bench user now and passes, which proves the participant grant works. Result entry stays on the REST helpers by an earlier explicit decision, but now under the participant's own session, so it exercises their permissions.

**7. Fixture corrupted `login_user_seq`.** Correct, and it had already happened: `last_value` was 3 against `max(id)` 4, so the next application-created user would have collided. I repaired the sequence on the dev stack and rewrote the fixture as a single `DO` block using `nextval` after a `setval` repair. Its completeness check now covers every role and the lab-unit mapping — the early return had left a half-granted user that no rerun repaired, which is exactly how I discovered that a bench user with no lab-unit mapping is offered **no sample types at all** and cannot enter an order.

**8. Cleanup failures invisible.** Fixed — every statement is still attempted, but failures are collected and raised after cleanup completes, so a run cannot go green while leaving rows behind. Panel contents are also removed by their panel-sample link, not only by box, which is what had stranded orphans.

**9. V1 checks could pass on sidebar text.** Correct — fixed by scoping to `getByRole("main")` with exact matching.

## Answered differently

**5. Add Order deep-link preload.** The requirement is real (FR-V2.2-12, `eqa-kb/eqa-v2-epic-and-stories.md:1020`) and unimplemented — the link passes only `?isEQA=true`. But the suggested one-line fix isn't available: passing the cycle id would work, while preloading the *programme* has nothing to preload from, because that select is keyed on this laboratory's free-text `eqa_lab_program_enrollment` rows, which carry no foreign key back to `eqa_program`. Delivering the requirement needs that link, which is a schema decision rather than a test change. The spec now asserts the programme select **starts empty** before choosing it, so the gap is recorded rather than masked.

## Hardening applied

Scores CSV is read, not just counted as downloaded — header checked, row count and verdicts matched against what was planted. Enrollment editing asserts exactly one row still carries the run tag. The provider register now asserts the self-lab row is absent (the missing direction).

Two I did not take: scoring still borrows catalog tests, because a panel sample stores an analyte rather than a test, so there is no panel-related test id to use. And KPI tiles are still asserted for presence rather than value — on a shared database those totals include pre-existing data, so an exact figure would be fixture-dependent; the deterministic numbers (progress counts, "1 unacceptable of 3", CSV rows) are asserted exactly.

## CI

You are right that no E2E ever runs for this PR. `03 - E2E` triggers only on `pull_request.branches: [develop, fix/madagascar-accession-results-file-e2e]`, and the `workflow_run` chain that actually runs Playwright hangs off it — so a PR based on `feat/eqa-next` gets nothing, fork or not. Adding `feat/eqa-next` to that allowlist is a one-line change, though note `e2e-tests.yml` also pins `github.repository == 'DIGI-UW/OpenELIS-Global-2'`. That is a decision for the epic owner; I have not touched CI in this PR.

Local evidence: the full EQA suite ran green three consecutive times, 18 tests each, ~1.6 minutes, leaving no seeded rows. Worth knowing that what looked like flakiness all along was the Docker VM out of memory (9.2 GB of 9.9, swap full) with a second OpenELIS stack running; with it paused, restarts stopped and suite time fell from 7.7 minutes to 1.6.

## Product defects found, none fixed here

1. A cycle can never be closed — no control on either machine issues `SCORED → CLOSED`.
2. In-house unblinding returns 500 (detached entity read of the panel's cycle).
3. Provider-register dismissals record no competency event, because the snapshot omits the result id.
4. A participating laboratory cannot validate its own results (401) or submit its own cycle (403).
5. The receipts tab shows stale rows after dispatch until a reload.
6. A non-array reply crashes the in-house page — `data || []` guards null only, and a 500 body reaches the tile arithmetic.

PR #4160, opened 2026-09-02
feat(eqa-v2): absorb V1 orders and distributions into cycles (OGC-608)

samuelmale:
# Pull Requests Requirements

- [x] The changes confirm to the OpenElis Global x3 [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide) documentation.
- [x] I have read and agree to the Contributing [Guidelines](https://openelis-global.org/community/get-involved/) of this project.

## Summary

EQA V2 absorbs the V1 order list and distribution pages. Every legacy distribution gains a synthetic provider cycle, so V1 data shows up on the cycle-centric V2 pages, and the retired V1 URLs redirect to their V2 homes.

**Backend — liquibase `qa/022` (ids `qa-072`–`qa-076`)**

- `qa-072` extends the cycle-transition trigger-event CHECK with `V1_BACKFILL`; `EQATriggerEvent` gains the same value so the cycle history can read the rows.
- `qa-073` backfills one **closed, historical** cycle per **COMPLETED** distribution that has no cycle link. The cycle copies the distribution's name and dates (planned and actual), takes the next cycle number on its scheme, and gets one `V1_BACKFILL` audit row. Active distributions (DRAFT, PREPARED, SHIPPED) are **not** migrated: V1 never stored the roster or panel samples a live V2 cycle needs, so there is nothing to convert, and a cycle created without them cannot pass the provider workbench gates. The statement lives in `022-v1-distribution-cycle-backfill.sql`; the changeset runs it through `sqlFile` and the test runs the same file.
- The rollback lives in `022-v1-distribution-cycle-rollback.sql`. It **refuses to run** when any backfilled cycle has activity recorded after the migration: a transition other than the backfill row, or a round, panel, roster entry, receipt, result, follow-up or shipping box attached. Otherwise it removes exactly the cycles, audit rows and links the backfill created.
- `qa-075` has no change of its own: its `onFail="WARN"` precondition writes a startup warning when active (non-COMPLETED) distributions without a cycle exist, so a missed pre-upgrade cleanup is visible in the log. WARN rather than HALT so a stray test row cannot block a boot.
- `qa-074` deactivates the superseded Orders and Distribution menu rows. My Programs stays active.
- `qa-076` flattens the "EQA Tests" menu group, which held only My Programs once Orders left: My Programs moves directly under EQA (between the management group and My Cycles) and the empty group is deactivated.
- The include follows `qa/037` on purpose: `qa-072` re-declares the CHECK that `qa/024` and `qa/028` extended.

**Decision record:** `docs/eqa/v1-distribution-absorption.md` records the one-completed-distribution / one-closed-cycle rule and amends FR-V2.1-03 ("legacy distribution rows remain unlinked"), which now applies to active distributions only. It also carries the operator note: complete or discard active V1 distributions before upgrading, because the V1 pages are removed.

**Frontend**

- `/qa/eqa/orders` and `/EQAOrders` redirect to My Cycles. `/qa/eqa/distribution`, `/qa/eqa/distribution/create` and `/EQADistribution*` redirect to the provider scheme list. The old URLs keep working for one release.
- Program Administration loses its Participants tab (its enrollment write had been failing silently). The standalone `/qa/eqa/participants` page remains.
- The now-unreachable V1 pages are deleted: orders list, distribution dashboard, distribution wizard, participants tab, plus their tests and the 89 `en.json` keys only they used.
- The QA Overview "EQA due" row opens My Cycles.

**Not migrated:** legacy `eqa_result` scores stay read-only in the V1 table. They predate cycles and panels, so a copy into `eqa_participant_result` would be lossy.

## Tests

- `EQAV1AbsorptionMigrationTest` (6 tests, real Postgres via Testcontainers; the menu test also asserts the flattened My Programs placement) runs the changeset's own SQL files against seeded legacy rows: a COMPLETED distribution becomes a CLOSED cycle with exact field values and one audit row; DRAFT/PREPARED/SHIPPED rows stay unlinked and produce no cycle; numbering continues after an existing cycle and starts at 1 on a fresh scheme; a pre-linked distribution is untouched; a second run changes nothing; the provider scheme read (`getProviderSchemes`) lists the migrated cycle as CLOSED with an active-cycle count of 0; the rollback removes exactly what the backfill created; the rollback refuses (and leaves everything in place) when a backfilled cycle has a later manual transition.
- EQA and QA-overview Vitest suites: 168/168.

**Why the test runs the SQL files rather than the Liquibase lifecycle:** the changelog runs once against an empty database at context start, so there are no legacy rows for it to touch in any test. `ComponentBackfillMigrationTest` (`041-component-backfill.sql`) established the pattern of executing the changeset's `sqlFile` directly; this test follows it for both the forward and the rollback statement.

## Live verification

On the dev stack, boot applied the three changesets. The one legacy distribution on the dev stack (`Round 1 - HIV VL 2026`) is DRAFT, so after the rework it stays unlinked and no cycle is created; the COMPLETED path is covered by the integration test. In the browser: all six redirects land on the new pages, Program Administration shows only EQA Programs and System Settings, the provider scheme list renders, and every active EQA menu link renders its page with no 404 or failed EQA API call.

## Related Issue

[OGC-608](https://uwdigi.atlassian.net/browse/OGC-608)

## Other

- A development database that applied the first version of `qa-073` needs its `databasechangelog` row removed before this build starts, because the changeset checksum changed.
- The "System Settings per scheme" fold is not part of this change; it is separate work.
- The V1 Results & Analysis page (`/qa/eqa/results`) is a similar leftover and is tracked separately rather than folded in here.

[OGC-608]: https://uwdigi.atlassian.net/browse/OGC-608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samuelmale:
Rework pushed as `7559a53b0` in response to the review.

- **Active distributions:** the backfill now migrates only `COMPLETED` distributions, as `CLOSED` historical cycles. DRAFT, PREPARED and SHIPPED rows stay unlinked. V1 never persisted a roster or panel samples, so those rows cannot be converted into operable V2 cycles; the ADR (`docs/eqa/v1-distribution-absorption.md`) tells operators to complete or discard them before upgrading.
- **Rollback:** moved to `022-v1-distribution-cycle-rollback.sql`. It refuses when any backfilled cycle has a non-backfill transition or a dependent round, panel, roster entry, receipt, result, follow-up or shipping box, and otherwise removes exactly what the backfill created.
- **Test:** now also runs the rollback file (clean and refused paths), asserts active statuses produce no cycle, and checks the provider scheme read lists the migrated cycle as closed history with `activeCycleCount` 0. Running the SQL files directly follows the `041-component-backfill.sql` precedent; the changelog itself runs once against an empty database at context start, so a lifecycle test has no legacy rows to act on.
- **Spec:** the ADR amends FR-V2.1-03.

Dev databases that applied the first version need the `qa-073` changelog row deleted before starting this build (checksum changed).

samuelmale:
Two small follow-ups pushed as `191f20274`:

- `qa-075` makes the ADR's operator note enforceable: a change-less changeset whose `onFail="WARN"` precondition logs a startup warning when active (non-COMPLETED) distributions without a cycle exist. WARN, not HALT, so a stray test row cannot block a boot.
- `qa-076` flattens the "EQA Tests" menu group, which held only My Programs once Orders was retired; My Programs now sits directly under EQA. Asserted in the migration test.

The V1 Results & Analysis page (`/qa/eqa/results`) is a comparable leftover; it is tracked as its own item rather than widened into this PR.

PR #2855, opened 2026-02-16
issue-2854: Initialize RESULT_EDIT_ROLE_ID in LogbookResultsRestController

ketanop321:
## Pull Requests Requirements
- [✅] The PR title includes a brief description of the work done, including the Issue number if applicable.
- [✅] The PR title follows conventional commit label standards.
- [✅] The changes confirm to the OpenElis Global [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838) documentation.
- [✅] The changes include tests or are validated by existing tests.
- [✅] I have read and agree to the [Contributing Guidelines](CONTRIBUTING.md) of this project.

## Summary

`LogbookResultsRestController.RESULT_EDIT_ROLE_ID` was declared but **never initialized** in the constructor. It remained `null` for the entire application lifecycle.

When the admin configuration `roleRequiredForModifyResults` is set to `"true"`, the `userNotInRole()` method uses this field to check authorization:

```java
return !roleIds.contains(RESULT_EDIT_ROLE_ID); // contains(null) → always false → !false → always true

PR #4186, opened 2026-09-04
issue-2854: Resolve Results role id for logbook result edit checks

### LINKED ISSUE #2854: Bug: RESULT_EDIT_ROLE_ID never initialized in LogbookResultsRestController — breaks role-based result editing
Description

[LogbookResultsRestController.RESULT_EDIT_ROLE_ID]   is declared at line 121 but never assigned a value in the constructor. It remains null for the entire application lifecycle.

This field is used in [userNotInRole()]  (line 925) to check if a user has permission to edit results:
return !roleIds.contains(RESULT_EDIT_ROLE_ID); // contains(null) → always false → !false → always true 

Impact

When the admin configuration roleRequiredForModifyResults is set to "true":

All non-admin users are locked out of editing lab results via the React UI
Even users with the "Results" role are blocked
This affects the primary result entry workflow at /rest/LogbookResults
Root Cause

The constructor only initializes REFERRAL_CONFORMATION_ID but not RESULT_EDIT_ROLE_ID. The correct pattern exists in AccessionResultsController (line 54-60), which properly looks up the "Results" role via RoleService. The LogbookResultsRestController even has an unused @Autowired RoleService roleService field, suggesting the initialization was intended but accidentally omitted.

Proposed Fix

Add RoleService to the constructor and initialize RESULT_EDIT_ROLE_ID (matching AccessionResultsController)
Make the field final (controllers are singletons — Constitution Principle IV)
Remove the unused @Autowired RoleService roleService field

Souravrajvi0:
## Summary

Fixes role-based result editing for logbook results when `roleRequiredForModifyResults` is enabled.

Closes #2854. Supersedes the closed #2855, which initialized an unused field in `LogbookResultsRestController` while the live code path goes through `ResultUtil.userNotInRole()`.

## Root cause

`LogbookResultsRestController` calls `ResultUtil.userNotInRole(request)`, but `ResultUtil.RESULT_EDIT_ROLE_ID` was never initialized. With role-based editing enabled, `roleIds.contains(null)` always returned `false`, so every non-admin was treated as lacking the Results role and results were locked.

## What changed

- Lazily resolve the Results role id in `ResultUtil` (same pattern as `confirmationReferralTypeId()`)
- Remove the unused `RESULT_EDIT_ROLE_ID` dead field from `LogbookResultsRestController`
- Add integration tests for users with and without the Results role

## Testing

```bash
mvn test -Dtest=ResultUtilUserRoleTest
```

Results: 2 tests, 0 failures.


Josephkagimu1 (commented):
Nice work — this fixes the root cause correctly. ✅

Moving the check into ResultUtil.userNotInRole() is the right call since that's the live code path, and I like that it fails closed (missing role → deny) rather than silently allowing edits. Tests covering both branches look good too.

A few things to look at:
- Add an @After to clear the SecurityContextHolder so the test security context doesn't leak into other tests.
- A LogEvent warning when getRoleByName returns null would help admins debug a misconfigured role — right now every non-admin is silently locked out with no trace in the logs.
- Since RESULT_EDIT_ROLE_ID is removed, is the controller's @Autowired UserRoleService field still used? If not, happy to see it removed here if you consider it in-scope, or in a follow-up PR if you'd rather keep this focused.

PR #4221, opened 2026-09-07
fix: refresh database and FHIR runtime images off Bullseye

pmanko:
## Why

The normal end-to-end image build fails before tests start because both the PostgreSQL and HAPI images use Bullseye security metadata that expired on September 7. Evidence: [failed Shared Build](https://github.com/DIGI-UW/OpenELIS-Global-2/actions/runs/34163136921/job/101870584318). This shared infrastructure failure blocked every PR until #4209 landed a stopgap (bullseye packages from archive.debian.org with the expiry check relaxed); this PR is the durable fix and removes that stopgap.

## Changes

- PostgreSQL 14.4 -> 14.24 on Debian Bookworm; retain the PostgreSQL major version and existing initialization scripts.
- Run the unchanged HAPI 6.6.0 WAR on Tomcat 9.0.121 / Java 17 / Ubuntu Noble. The old HAPI image supplies the WAR only, not the final runtime OS.
- Preserve the checked-in FHIR configuration, mutual TLS settings, UID 8443, and certificate group 8443.
- Add a focused, isolated container smoke test and document existing-database deployment precautions.

No CI workflow changes, manually posted statuses, alternate end-to-end path, analyzer changes, or disabled package authenticity/expiration checks. This is a standalone PR targeting develop, not a new analyzer milestone.

## Validation

Passed locally using linux/amd64 images in OrbStack:

- Clean PostgreSQL image build, including apt update/install.
- Clean HAPI runtime image build, including apt update/install.
- `bash scripts/smoke-test-service-images.sh`: fresh OpenELIS database initialization; FHIR R4 metadata over mutual TLS; runtime UID 8443; rejection without a client certificate; synthetic Patient write/read surviving both services restarting.
- Original and replacement HAPI WAR SHA-256 match: `6b4d61ef8394a8abe47607f99812aba80a67da2d109ed6fc081c95d3a71e200f`.
- Full `mvn spotless:apply` and uncached `mvn clean spotless:check`, shell syntax check, and `git diff --check`.
- Inspected database/FHIR logs and response evidence. The run contains Hibernate schema-initialization constraint-drop messages and connection warnings during the deliberate simultaneous service shutdown; startup and post-restart persistence assertions passed.
- Smoke-test containers, network, and certificate volume were cleaned up; existing local stacks were untouched.

The existing normal GitHub end-to-end workflow remains the application gate.

## Deployment Risks

This is not a full HAPI application upgrade or PostgreSQL major-version upgrade. Existing database volumes need backup and rehearsal: changing operating-system locale libraries can require rebuilding text-ordering indexes. See `docs/service-image-maintenance.md`. Fresh-database smoke evidence does not establish existing-site collation safety.

No deployment or merge is performed by this PR.

## Why this is still needed after #4209

#4209 kept the CI build alive by pointing the two Debian 11 (bullseye) images at `archive.debian.org` and telling apt to accept the archive's expired Release files. That was the right move for the night the pipeline broke, and it is the wrong place to stay:

- **End of life means no more fixes, not just a broken mirror.** Debian 11's LTS ended on 2026-08-31. From that date no security update is published for anything in the image: glibc, OpenSSL, libpq, curl, the JVM's native dependencies. The expired Release file that broke CI on 2026-09-07 was the first visible symptom; every CVE from now on is the invisible one. Keeping `postgres:14.4` and `hapiproject/hapi:v6.6.0-tomcat` means shipping a database and a FHIR server on an operating system nobody patches.
- **The archive is a museum, not a mirror.** `archive.debian.org` serves frozen snapshots on a best-effort basis: the `bullseye-security` suite is not there yet (404), package pools get pruned (the current `security.debian.org` pool already 404s on `libcurl4 deb11u16`), and `Acquire::Check-Valid-Until=false` only papers over the fact that the metadata is no longer maintained. It can break again on any day, with the same symptom.
- **`postgres:14.4` is 20 minor releases behind.** 14.24 carries two years of data-corruption and security fixes for the same major version; nothing in OpenELIS depends on 14.4 specifically.

This PR moves both images onto supported bases and stops depending on the archive: PostgreSQL 14.24 on Debian 12 (bookworm), and the unchanged HAPI 6.6.0 WAR on Tomcat 9 / Java 17 / Ubuntu 24.04. The CI build compose needs no change: it never mounted `server.xml` (the image bakes `tomcat/hapi_server.xml`) and already passes the `-Dhapi.ssl.*` properties that file expects.

## Existing databases: what changes and what to do

Data compatibility is intact: 14.4 → 14.24 is a minor upgrade of the same major, in place, no dump or `pg_upgrade`. What does change is the operating system's collation library (glibc 2.31 → 2.36), and PostgreSQL 14 does not warn about the database's default collation. Measured on a copy of a real OpenELIS database (en_US.utf8, 914 indexes, 11 MB):

- `pg_collation` records `en_US.utf8` at version 2.31; the new image reports 2.36.
- `REINDEX DATABASE clinlims` as superuser took 1.0 s on that copy; it scales with index size.
- `ALTER COLLATION "en_US.utf8" REFRESH VERSION` (and `"en_US"`) then records 2.36.

Deployment step for every existing site (testing included): back up, start on the new image, run `REINDEX DATABASE clinlims` and the two `REFRESH VERSION` statements during the maintenance window, then open the site. `docs/service-image-maintenance.md` carries the same instructions.

## Verified

- Both images build from this branch (no archive, no relaxed apt checks).
- `scripts/smoke-test-service-images.sh`: fresh database initialization, non-root FHIR (uid 8443), mutual TLS, R4 metadata, Patient write/read across a restart of both services.
- OpenELIS dev stack on these images with fresh volumes: Liquibase initialised PostgreSQL 14.24, OpenELIS wrote to HAPI over mutual TLS at start-up (5 Questionnaires, 1 QuestionnaireResponse in `hfj_resource` after the checks), admin programme save and Add Order questionnaire 9/9, browser patient search on the Results page 15/15.
- OpenELIS dev stack on these images against a byte-for-byte copy of an existing 14.4 volume (in-place, no conversion): data intact (samples, patients, programmes, 1089 changesets), clean start, browser patient-search flow 15/15 before remediation, `REINDEX DATABASE clinlims` 1.04 s and `REFRESH VERSION` to 2.36 applied in place, 15/15 again afterwards.

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -1,5 +1,5 @@
-# Use the official PostgreSQL 14.4 image as base
-FROM postgres:14.4
+# Keep the PostgreSQL major version on a supported operating-system base.
+FROM postgres:14.24-bookworm
 
 RUN apt-get update && apt-get install -y gettext-base && rm -rf /var/lib/apt/lists/*
 # Copy initialization scripts and data files into the container
--- a/docs/service-image-maintenance.md
+++ b/docs/service-image-maintenance.md
@@ -1,0 +1,47 @@
+# Database and FHIR Image Maintenance
+
+The database image uses PostgreSQL 14.24 on Debian Bookworm. The FHIR image runs
+the unchanged HAPI 6.6.0 WAR on Tomcat 9 with Java 17 on Ubuntu Noble. The old
+HAPI image supplies only the WAR during the build, not the final operating
+system, Tomcat installation, or entrypoint. Configuration stays at
+`/opt/openelis/config/hapi_application.yaml`; TLS settings and runtime user 8443
+are preserved. The Tomcat installation is now `/usr/local/tomcat`.
+
+This addresses the expired Bullseye package metadata that prevented clean image
+builds. It does not constitute a HAPI application upgrade or a PostgreSQL
+major-version upgrade. Both still need their own planned upgrades; pinning
+versions does not replace regular dependency maintenance.
+
+## Validate Images
+
+Run from the repository root with Docker and jq available:
+
+```bash
+docker build --platform linux/amd64 -f db/Dockerfile -t openelis-base-refresh-db:local .
+docker build --platform linux/amd64 -f fhir/Dockerfile -t openelis-base-refresh-fhir:local .
+bash scripts/smoke-test-service-images.sh
+```
+
+The smoke test creates only uniquely named containers, a network, and a
+certificate volume. It exposes no host ports, removes its resources on exit, and
+prints the directory containing its logs and response evidence. It checks fresh
+OpenELIS database initialization, non-root FHIR startup, mutual TLS, FHIR R4
+metadata, and saving/reading a synthetic patient across both services' restart.
+It is an infrastructure test, not browser acceptance or a new CI pipeline. The
+existing end-to-end workflow remains the application gate.
+
+## Existing Deployments
+
+Back up and rehearse the update on a copy before replacing an existing database
+container. Keep the database volume; do not reset production data. An
+operating-system change also changes locale libraries. Existing indexes that
+depend on locale-sensitive text ordering may need rebuilding before normal
+traffic resumes. Review both the database's default locale and any explicit
+column/index collations. Refreshing a recorded collation version alone does not
+rebuild affected indexes. Follow the PostgreSQL
+[collation maintenance guidance](https://www.postgresql.org/docs/14/sql-altercollation.html).
+
+The fresh-database smoke test does not establish that an existing site's indexes
+are safe after that library change. Site-specific backup, index maintenance,
+restore, and rollback validation are deployment responsibilities. Do not disable
+package signature or expiration checks to retain an obsolete base image.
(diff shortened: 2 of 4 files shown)


mozzy11 (approved):
Reviewed against the failing E2E builds: bullseye-security's Release file expired at 21:13 UTC on 2026-09-07 (LTS ended 2026-08-31), so every `03 - E2E` Shared Build since then dies in `apt-get update`. This PR is the right fix and CI is fully green on it.

- Moving Tomcat from `/opt/bitnami/tomcat` to `/usr/local/tomcat` leaves `install/installerTemplate/linux/templates/docker-compose.yml:121` mounting `hapi_server.xml` into the old path, so a site's customised server.xml would be silently ignored once this image ships. Worth updating that mount (or symlinking the old path) in this PR or a follow-up.
- `projects/analyzer-harness/docker-compose.dev.yml` and `dev.docker-compose.yml` still run the upstream `hapiproject/hapi:v6.6.0-tomcat` image directly, so they are unaffected.
- Every open PR (including #3560 and #3563, which are otherwise approved with auto-merge armed) is blocked on this, so merging it soon unblocks the queue.

PR #4240, opened 2026-09-09
docs(eqa): restore the CPHL cutover checklist and programme registry template (OGC-935)

samuelmale:
## Why this is open again

The CPHL cutover checklist and the programme-registry CSV were written for #4146 and
dropped when the configuration moved under `volume/`, so the cutover instrument — the one
carrying CPHL's sign-off block — was never on the branch. #4169 restored them and was
closed unmerged with no review and no comment, which left the finding and its register
row open. Both files come back here, rebased onto the current tip.

## The restore is faithful, and that is checkable

`upstream/task/T-30` no longer exists. Both files were recovered from #4146's **first**
commit `6a96d9852` — its second commit, which moved the CPHL configuration under
`volume/`, is what dropped them.

Against that commit:

- `docs/eqa/cphl-eqa-programs.csv` is **byte-identical**;
- `docs/eqa/cphl-cutover.md` differs only in the results-collection row (rewritten to name
  what shipped after it was written), the loading section below, and prettier's table
  padding. Nothing else was lost or altered in the recovery.

## What is different from #4169

The loading section is the part a reader follows literally, and a wrong path or a wrong
endpoint there is worse than no document. So every claim in it was checked against
`EQAProgramConfigurationHandler` and `ConfigurationInitializationService`, and then
against a running instance that already carries this registry.

**Confirmed, not changed:**

- the configuration path `…/configuration/backend/eqa-programs/cphl-eqa-programs.csv` —
  the deployed file on that instance is byte-identical to the copy in this PR
  (`sha256 c9a4dcbb…`), and the domain's checksums file records that same hash;
- the six-row SQL check — it returns exactly six rows;
- the note that a scheme reaches the provider board only on its first **active**
  enrollment. `findProviderSchemeRows` inner-joins active enrollments, and on that
  instance all six programmes exist while only the two with enrolled laboratories appear
  on the board. Four deliberately absent, two present.

**Added, each confirmed live:**

- the reload endpoint takes an administrator, and `{"domains": ["eqa-programs"]}` scopes
  it to this registry instead of reloading every configuration domain;
- what the reload answers, so a reader can tell "found and already loaded" from an empty
  `files` list, which points at the path rather than at the file. The scoped call returned
  `{"domain":"eqa-programs","fileName":"cphl-eqa-programs.csv","status":"SKIPPED","skippedReason":"checksum matches"}`;
- a **blank cell leaves the existing value alone** rather than clearing it, so the file
  cannot be used to empty a field;
- an unchanged file is skipped on its checksum, so re-applying one as it stands needs
  `{"force": true}`;
- a value edited on **Schemes & Programs** is **overwritten** the next time the file gives
  that column a value — which is what decides, per column, whether the CSV or the screen
  is the source of truth;
- a row the server rejects is logged and skipped without stopping the rest of the file.

The reload driven for this was scoped to `eqa-programs` and answered `SKIPPED`, so it
changed no row on that instance.

## What the verification does not cover

The instance it was verified on is the **development** stack, which mounts
`./volume/configuration`. So the six programmes loading there is not evidence that a
production deployment would load them, and the `./configuration` mount in
`docker-compose.yml` is the one part of §4 that no instance here exercises.

What the verification does establish is stack-independent, because it concerns the loader
and the container path rather than which host directory is mounted: the domain name, the
file's location under the config directory, the checksum skip, the upsert behaviour, the
six-row check, and the board's enrollment gate.

That asymmetry is the same trap as §4's, seen from the other side, and it is why the
template has to travel with the checklist rather than be read out of the repository.

## Considered and deliberately not added

A step to assign each scheme's tests after loading. `EQAProgramTest` is read only by the
programmes dialog — nothing in the ordering, blinding, or scoring path consumes it — so it
is not a precondition for the end-to-end run in §6, and adding it would send CPHL chasing
a step they do not need.

## Scope

Docs only, no code and no schema. The §2 table is re-padded by prettier and otherwise
untouched.

### DIFF
--- a/docs/eqa/cphl-cutover.md
+++ b/docs/eqa/cphl-cutover.md
@@ -1,0 +1,117 @@
+# CPHL EQA cutover checklist — Access/Excel to OpenELIS
+
+**Scope.** OGC-935 (PNG Phase II, contract item 18). This checklist confirms
+that OpenELIS replaces the two systems CPHL Port Moresby uses to run its
+national EQA programme: the Access database that produces performance reports,
+and the Excel workbooks that track sample distribution to provincial
+laboratories. Cutover is complete when every row in §2 and §3 is checked and a
+CPHL stakeholder signs §6.
+
+**How to use this document.** Work through it on a CPHL instance with the
+programme registry loaded (§4). Each unchecked row is either a task or a
+question for CPHL — there are no other states.
+
+---
+
+## 1. Programme registry (confirm with CPHL)
+
+The registry file (`docs/eqa/cphl-eqa-programs.csv`, loaded per §4) defines six
+provider-side schemes, one per ePT-validated test domain (AC-V2.1-23). The
+domains are the best available registry; CPHL confirms or corrects each row at
+sign-off. Every value is editable afterwards on **Schemes & Programs**.
+
+| Programme                            | Test section      | Frequency | CPHL confirms                |
+| ------------------------------------ | ----------------- | --------- | ---------------------------- |
+| CPHL National HIV Serology EQA       | Serology          | Quarterly | ☐ name ☐ section ☐ frequency |
+| CPHL National HIV Viral Load EQA     | Molecular Biology | Quarterly | ☐ name ☐ section ☐ frequency |
+| CPHL National EID EQA                | Molecular Biology | Quarterly | ☐ name ☐ section ☐ frequency |
+| CPHL National HIV Recency EQA        | Serology          | Quarterly | ☐ name ☐ section ☐ frequency |
+| CPHL National COVID-19 Molecular EQA | Molecular Biology | Quarterly | ☐ name ☐ section ☐ frequency |
+| CPHL National TB Microscopy EQA      | Microbiology      | Quarterly | ☐ name ☐ section ☐ frequency |
+
+Open questions for CPHL:
+
+- ☐ Are there programmes CPHL runs that are missing from this list (for example
+  syphilis, malaria microscopy, CD4)?
+- ☐ Which international or regional schemes does CPHL _participate in_ (as
+  opposed to provide)? Those are entered as `INTERNATIONAL_PT` schemes with the
+  external provider named — none are seeded, because no authoritative list
+  exists in the project record.
+- ☐ The provincial laboratory register (name, province, contact, delivery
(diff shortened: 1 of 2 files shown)


samuelmale:
Closing: the decision is that these docs are not necessary in this repository.

This is the second time this content has been closed unmerged, so the reason is recorded
here rather than left to be rediscovered. #4169 was closed on 2026-09-03 "pending a
decision on whether those docs belong on this branch at all", and that decision has now
been made — implementation documentation for a single deployment does not belong in the
product repository.

**Not the reason:** the `docs/eqa/cphl-eqa-programs.csv` copy being redundant. It is not.
The base carries a byte-identical copy at `volume/configuration/backend/eqa-programs/`,
but `dev.docker-compose.yml` mounts `./volume/configuration` while `docker-compose.yml`
mounts `./configuration`, a directory this repository does not carry. So on a production
Compose deployment the `volume/` copy is never read. Anyone reopening this on the grounds
that the CSV already ships should know it does not ship where a deployment reads it.

**Worth keeping from the work, wherever the cutover instrument ends up living.** Each of
these was checked against `EQAProgramConfigurationHandler` /
`ConfigurationInitializationService` and then against a running instance:

- the registry loads from `<config dir>/eqa-programs/*.csv`, domain name `eqa-programs`;
- `POST /rest/configuration/domains/reload` takes an administrator; an empty body reloads
  every domain, and `{"domains": ["eqa-programs"]}` scopes it to this one;
- an unchanged file is skipped on its checksum — the response reads
  `{"status":"SKIPPED","skippedReason":"checksum matches"}` — so re-applying one as it
  stands needs `{"force": true}`;
- rows upsert by programme name, and a **blank cell leaves the existing value alone**
  rather than clearing it, so a value edited on **Schemes & Programs** is overwritten the
  next time the file fills that column;
- an unknown `testSection` logs and leaves the link empty; a row the server rejects is
  logged and skipped without stopping the file;
- a scheme reaches the provider board only on its first **active** enrollment
  (`findProviderSchemeRows` inner-joins them) — verified live, where all six programmes
  existed and only the two with enrolled laboratories appeared;
- **being present in the repository does not mean an instance has loaded it.** The reload
  response and the row check are what settle that.

Finding F-10 and the register row it blocks now need a disposition on that basis rather
than a fix: the missing artifact is deliberate, not a defect.

PR #4264, opened 2026-09-10
fix(tomcat): the app deploys once on Docker, so each scheduled job runs once

### LINKED ISSUE #4262: Docker server.xml deploys the app twice: every @Scheduled job runs twice (doubled Modbus traffic, duplicate alerts and e-mails)
## The Docker `server.xml` deploys the app twice, so every `@Scheduled` job runs twice

`tomcat/oe_server.xml:139` pairs `autoDeploy="true"` — with `deployOnStartup` left unset, so it
defaults to `true` — with an **explicit** context for the same WAR:

```xml
<Host name="localhost" appBase="webapps"
      unpackWARs="true" autoDeploy="true">                        <!-- line 139 -->

    <Context docBase="ROOT" path="/api" />
    <Context docBase="OpenELIS-Global"
             path="/api/OpenELIS-Global/" logEffectiveWebXml="true">
```

Tomcat therefore deploys `webapps/OpenELIS-Global.war` **twice**: once at `/OpenELIS-Global`, derived
from the filename by the host's own deployer, and once at the declared `/api/OpenELIS-Global/`. Two
web application contexts, two Spring `ApplicationContext`s, and two of every singleton inside them —
including two schedulers.

`install/installerTemplate/linux/templates/oe_server.xml:110` already gets this right
(`autoDeploy="false" deployOnStartup="false"` alongside the same explicit contexts), so the installer
path is fine. Only the Docker-facing copies diverged.

### Where it lands

| file | consumed by | `autoDeploy` | `deployOnStartup` | explicit OE `<Context>` | double-deploys? |
|---|---|---|---|---|---|
| `tomcat/oe_server.xml` | `Dockerfile:128` → the published image | `true` | unset (⇒ true) | yes | **yes** |
| `volume/tomcat/oe_server.xml` | `dev.docker-compose.yml:68`, `build.docker-compose.yml:72` | `true` | unset (⇒ true) | yes | **yes** |
| `dev/tomcat/oe_server.xml` | — | `true` | unset | no | no (one context) |
| `demo/oe_server.xml` | — | `false` | `false` | yes | no |
| `install/installerTemplate/.../oe_server.xml` | the installer | `false` | `false` | yes | no |

So this affects **every Docker deployment**: the published image, the dev stack developers work
against, and the `build.docker-compose.yml` stack the Playwright E2E workflows run against in CI.

The published image ships exactly the repo's file — byte-identical, so there is no post-build
patching that rescues it:

```
repo  tomcat/oe_server.xml                  8ed6a9c1e45bb803f780a0e7c0507c24972be22f33903d19440de7005171cc71
image /usr/local/tomcat/conf/server.xml     8ed6a9c1e45bb803f780a0e7c0507c24972be22f33903d19440de7005171cc71
```

### Measured on `develop` @ `5c1c33dc4`

WAR built from `develop` under JDK 21 and served by the published
`itechuw/openelis-global-2:develop` image with the **shipped** `server.xml` in place. Both contexts
come up:

```
$ ls /usr/local/tomcat/webapps/
api   api#OpenELIS-Global   OpenELIS-Global   OpenELIS-Global.war   ROOT   ROOT.war
      ^^^^^^^^^^^^^^^^^^^   ^^^^^^^^^^^^^^^
      the declared context  the filename-derived one
```

The clearest observable is the cold-storage Modbus poller, because it writes a timestamped row on a
fixed cadence. With the shipped config, readings for a single device arrive as two interleaved series
from two independent schedulers:

```
recorded_at                    temp
2026-09-10 08:17:39.975  -20.5      <- scheduler A
2026-09-10 08:18:04.232  -20.5      <- scheduler B
2026-09-10 08:18:10.655  -20.5      <- scheduler A
2026-09-10 08:18:34.669  -20.5      <- scheduler B
```

With a single context, the same device produces one clean ~30 s series. Same WAR, same image, same
database — only `autoDeploy`/`deployOnStartup` differ.

### Why it matters beyond cold storage

There are **19 `@Scheduled` methods across 13 classes** on `develop`, and every one of them runs twice
on a Docker deployment:

```
coldstorage/service/impl/ModbusPollingService.java
dataexchange/fhir/service/FhirApiWorkFlowServiceImpl.java
eqa/scheduler/EQADeadlineAlertScheduler.java
fhir/transormation/controller/FhirTransformationController.java
organization/service/OrganizationImportServiceImpl.java
provider/service/ProviderImportServiceImpl.java
sample/scheduler/RequiredByAlertScheduler.java
scheduler/independentthreads/MalariaResultExporter.java
scheduler/independentthreads/ResultExporter.java
scheduler/independentthreads/TestUsageBacklog.java
…
```

Several of those are **outward-facing** — FHIR workflow polling, the result exporters, organization
and provider imports — so the doubling is not merely wasted work; it is duplicated traffic to and from
other systems, and duplicated alerting to people.

Confirmed consequences for cold storage specifically, which is where I measured it:

- double the Modbus traffic to every monitored freezer
- **two** `FREEZER_OFFLINE` alerts per outage
- **two** notification e-mails per alert

For the rest of the schedulers the doubling follows from the same two contexts; I did not measure each
one individually.

There is a second-order effect worth noting: because CI's Playwright stack
(`build.docker-compose.yml`) is also double-deployed, any E2E flakiness that depends on scheduled work
firing once has been racing two schedulers all along.

### Fix

Bring the two Docker-facing copies in line with the installer template that already works —
`autoDeploy="false" deployOnStartup="false"` on the `<Host>`, keeping the explicit `<Context>`
elements as the single source of deployment:

- `tomcat/oe_server.xml:139`
- `volume/tomcat/oe_server.xml`

Verifying the fix is one command — `ls /usr/local/tomcat/webapps/` should show
`api#OpenELIS-Global` and the `.war`, but **no** unpacked `OpenELIS-Global` directory.

Note for whoever picks this up: a container that has already run with the current config leaves stale
unpacked `OpenELIS-Global/` and `ROOT/` directories in `webapps/`, so recreate the container rather
than restarting it, or the old expanded context can be picked back up.

### Provenance

Pre-existing; **not** introduced by #3828. Found while verifying #3743 and #3904 at runtime — split out
of #4261 (item 1) because it is a platform/packaging bug rather than a cold-storage one, and its blast
radius is every scheduled job in the application.

mherman22:
Fixes #4262

## What the ticket asked

Docker deployments were deploying the WAR twice, so every `@Scheduled` method ran
twice: doubled Modbus traffic, two `FREEZER_OFFLINE` alerts per outage, two
notification e-mails per alert.

## Root cause

A Tomcat `<Host>` deploys a web application from two independent sources when both
are active: the Host's own deployer, which scans `appBase` when `deployOnStartup`
is true, and any explicitly configured `<Context>` children, which deploy
regardless. `docBase="OpenELIS-Global"` names a resource inside that same Host's
`appBase`, so one WAR was the target of both.

`deployOnStartup` was **absent** from the two Docker-facing copies, and the
[Tomcat 10.1 Host documentation](https://tomcat.apache.org/tomcat-10.1-doc/config/host.html)
gives its default as `true`. So the absent attribute, not the visible
`autoDeploy="true"` (which governs runtime rescanning), is what double-deployed at
startup.

That page names the remedy this PR applies:

> Note that if you are defining contexts explicitly in server.xml, you should
> probably turn off automatic application deployment or specify `deployIgnore`
> carefully. Otherwise, the web applications will each be deployed twice, and that
> may cause problems for the applications.

`install/installerTemplate/linux/templates/oe_server.xml:109-110` has carried the
identical explicit `<Context>` block with both attributes `false` since
2023-01-03 (`57e517f00`), so the installer path was never affected and the two
Docker copies are now brought into line with it.

Alternatives the docs offer, and why not: moving `docBase` outside `appBase` would
relocate the WAR out of `webapps/`, touching the Dockerfile, every compose mount
and the installer templates; `deployIgnore` keeps scanning on and needs a regex
kept in sync with WAR filenames; renaming to `api#OpenELIS-Global.war` would
require moving the `<Resources>` and the `jdbc/LimsDS` `<Resource>` into a
`conf/Catalina/localhost/` descriptor.

## Runtime verification of the merging head

Driven on the `dev.docker-compose.yml` stack against commit `3bc20286`, with the
container's `OpenELIS-Global.war` sha256 confirmed identical to the host build
(`51518a2a2b05ad1c…`) and the container's `conf/server.xml` confirmed to be the
bind-mounted `volume/tomcat/oe_server.xml` from this branch.

**One deployment, not two.** `webapps/` holds `api`, `api#OpenELIS-Global`,
`OpenELIS-Global.war` and `ROOT.war`, with no unpacked `OpenELIS-Global` and no
unpacked `ROOT` directory. `work/Catalina/localhost/`, which gets one entry per
context Tomcat actually started, holds exactly `api` and `api#OpenELIS-Global`.
Over the full container log: `Initializing Spring root WebApplicationContext` = 1,
`Root WebApplicationContext: initialization started` = 1, `Quartz Scheduler v.2.3.2
created` = 1, `Server startup in` = 1, and `deployWAR` = `Deploying web application
archive` = `Deploying web application directory` = 0 (the string `HostConfig` never
appears). Cross-checked inside the JVM via `/proc/<pid>/task/*/comm`: a single
10-thread `pool-2-thread` group (SchedulerConfig's one `@Scheduled` pool), one
`MyScheduler_QuartzSchedulerThread`, three `MyScheduler_Worker`. One set, not two.

The paths the appBase scan would have created are gone: `:8443/OpenELIS-Global/`
and `:8443/` both 404, while `:8443/api/` answers 302 and
`:8443/api/OpenELIS-Global/` answers 200.

**Driven through the public nginx proxy, with 0 browser console errors** (25
console messages, 0 errors, 0 warnings): the React UI at `/` (renders the
dashboard and its ten backend-served metric tiles) and `/login` (locale combobox
populated from its own REST call); the Cold Storage Dashboard at
`/FreezerMonitoring`; the legacy UI under `/api/OpenELIS-Global/`; and
`/api/OpenELIS-Global/rest/supportedlocales/active` returning its real JSON body,
not merely a 200. Login as admin succeeded and the webapp logged exactly one
`Successful login attempt for admin`. The container healthcheck wiring
(`build.docker-compose.yml:84` to `install/openelis_healthcheck.sh:12`) returns 200
with `{"status":"UP"}`.

No `OverlappingFileLockException` anywhere in the log, no `SEVERE`, no
`Context … startup failed`.

No user-facing URL changes: nginx sends `/` to the frontend and only `/api/` to
Tomcat, so the React routes never reached the context this PR removes. The CI
analyzer repoint was checked against a live mapping too: `POST
/api/OpenELIS-Global/analyzer/fhir` returns 400 (controller reached, bad body)
where the old prefix returns 404.

Two limits on the above, stated rather than glossed. The scheduler cadence
reading from the original issue could not be reproduced, because this stack has
`freezer`, `freezer_reading` and `storage_device` all empty, so
`ModbusPollingService.pollDevices()` has nothing to poll and writes no rows; no
fixture was manufactured, and the thread and context counts are the substitute
evidence. And the "before the change it deployed twice" half rests on Tomcat's
documented `appBase` behaviour and on this PR's own guard test, not on a
measurement, since reverting the config to compare was out of scope for the
verification run.

## Upgrade note (measured, and stronger than it looks)

A container that has already run with the old config leaves an unpacked
`webapps/OpenELIS-Global/` behind. With `deployOnStartup="false"` nothing
re-expands the WAR, and Tomcat prefers an existing directory over `<docBase>.war`.
This was reproduced deliberately: with a stub directory in place, an in-place
`docker compose restart` made the explicit `<Context>` resolve to that directory,
`logEffectiveWebXml` printed a bare welcome-file-list `web.xml`, a file planted in
the stub was served at `/api/OpenELIS-Global/`, and
`/api/OpenELIS-Global/health` returned **404**. The application does not come up
at all, silently, rather than merely serving stale code.

**So recreate the container rather than restarting it.** The published image is
unaffected: its `webapps` is image-local and `Dockerfile` clears it (confirmed by
`docker run --rm itechuw/openelis-global-2-dev:develop ls /usr/local/tomcat/webapps/`,
which ships only the two `.war` files).

## Why the CI bridge URI and the dev-setup URL moved to `/api/`

Removing the filename-derived deployment removes the `/OpenELIS-Global` context,
measured returning 404 afterwards. Two tracked files targeted it:

- `.github/ci/ci.analyzer-harness.yml` pointed the analyzer bridge there, so the
  harness E2E lane would 404 on every forward. It is composed with
  `build.docker-compose.yml`, which mounts the `volume/tomcat/oe_server.xml` this
  PR fixes, so it always rides on the fixed config. `/api/OpenELIS-Global` works
  against unfixed containers too, so the repoint removes a dependency on the defect
  rather than adding a version dependency.
- `docs/dev_setup.md:71` sent developers to `https://localhost:8443/OpenELIS-Global`
  after `docker compose up -d --build`. Root `docker-compose.yml` runs the published
  image, whose `server.xml` comes from `Dockerfile:128`, so that URL 404s once this
  merges. It now points at the React UI at `https://localhost/`, with the legacy UI
  and REST base named under `/api/OpenELIS-Global/`. Both new URLs were driven and
  answer; the removed one is now genuinely 404.

The bridge URI was deliberately set to the prefix-less path in e1e38bd4f, on the
stated grounds that "the `/api/` path prefix is only used by the proxy's location
block rewrite rules". That is not so: `/api/OpenELIS-Global` is Tomcat's own context
path, and `install/openelis_healthcheck.sh:12` curls
`https://oe.openelis.org:8443/api/OpenELIS-Global/health` directly on the
container's connector with no proxy involved, wired as the container healthcheck at
`build.docker-compose.yml:84`. The prefix-less path worked only because the WAR was
deployed twice.

## Not in this PR

- `projects/analyzer-harness/docker-compose.analyzer-test.yml:94` still sets the
  bridge's `ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI` to
  `https://oe.openelis.org:8443/OpenELIS-Global/analyzer`, which 404s after this
  change. `projects/analyzer-harness/docker-compose.dev.yml:39` mounts the
  `volume/tomcat/oe_server.xml` this PR fixes, so that overlay does ride on the
  fixed config. Left out deliberately to keep this change to the double-deployment
  fix; it is the one real follow-up here.
- `volume/openelis-analyzer-bridge/configuration.yml:7` carries the same old URL
  under an `openelis.url` key, but nothing binds that key: the bridge reads
  `org.itech.ahb.forward-http-server.*`
  (`HTTPForwardServerConfigurationProperties.java:12`), and CI mounts this same
  file while overriding only the env var, which is why the E2E lane is green. It
  404s nothing, and is worth deleting as dead config rather than repointing.
- `docs/dbprofiles.md:30`, `:45`, `docs/add_dr.md:190`, `docs/deployomrs.md:52` and
  `docs/install.md:291` name the removed path for installer deployments, where the
  installer template has set both attributes `false` since 2023-01-03, so they were
  already wrong before this PR.
- `dev/tomcat/oe_server.xml` declares no `<Context>`, so it deploys once and is
  untouched. `docs/dev_setup.md:140` has developers copy it into an Eclipse Tomcat,
  where the filename-derived path is correct.
- Hibernate Search stays disabled. Only the comment above it changes, to describe
  the single-deployment state rather than assert the double deployment as live fact.
  Nothing reads that index: `DBSearchResultsServiceImpl` is `@Primary` and no
  injection site qualifies the Lucene implementation.
- `install/tomcat-resources/ROOT.war`'s `index.jsp` permanently redirects to
  `/OpenELIS-Global/`, so a direct hit on `https://host:8443/api/` now 301s to a
  404. Nothing on a live path reaches it, since nginx sends `/` to the frontend and
  only `/api/<something>` to Tomcat, and the frontend and Playwright suites all use
  `/api/OpenELIS-Global/...`. The repo's own source, `ROOT/WebContent/index.jsp`,
  already redirects to `/api/OpenELIS-Global/`; only the checked-in binary WAR is
  stale.

## Test

`ServerXmlSingleDeploymentTest` reads every git-tracked `*server.xml`, walking
git's index rather than the filesystem so an untracked stray config in a
developer's checkout cannot decide the result, and asserts two things about each
`<Host>`:

1. A Host declaring a `<Context>` whose `docBase` resolves inside that Host's
   `appBase`, under a name other than the one Tomcat would derive for it, sets both
   `autoDeploy` and `deployOnStartup` to a present, literal false. Presence matters,
   because an absent `deployOnStartup` is the defect. An absent or empty `appBase`
   is read as `webapps`, its documented default, and an `appBase` that resolves to
   no segments at all (`.`, `./`, `/`) is checked rather than skipped.
2. No two `<Context>` children of one Host resolve to the same `docBase` or the
   same context path, so a second context over one WAR is caught however it is
   spelled. The spellings that collapse to one key include `OpenELIS-Global`,
   `OpenELIS-Global/`, `./OpenELIS-Global`, the absolute container path,
   `webapps/../webapps/OpenELIS-Global`, and `OpenELIS-Global.war`.

A `<Context>` is only counted against the appBase scan when its declared name
differs from its docBase's base name, since Tomcat's `HostConfig.deploymentExists`
otherwise skips its own copy and the app deploys once.

To see what the guard catches, mutate an attribute or a `docBase` in either
`oe_server.xml` and read the failures rather than trusting a count. Restoring the
base version of both files reddens it with four violations naming both files and
both attributes.

What the guard does not reach, stated rather than claimed away: the invariant is
per-Host, so one WAR served from two `<Host>` blocks is not flagged; under a `${}`
placeholder `appBase` docBase keys are not comparable, leaving only the attribute
check on that Host; keys are segment text, so symlinks, case-insensitive filesystem
aliasing and `##` version markers are unresolved; a `<Context>` declared in
`conf/Catalina/localhost/*.xml` or an untracked mounted file is outside it, as is
`deployIgnore` as Tomcat's alternative remedy. It is a static config check and does
not prove at runtime that one scheduler runs, hence the runtime verification above.

`dev/tomcat/oe_server.xml` and the four `hapi_server.xml` copies declare no
`<Context>`, so they are skipped as inapplicable. A `<Context>` whose `docBase` is
genuinely outside `appBase`, the Tomcat-recommended layout, is allowed to keep
appBase scanning on.

`.github/workflows/backend.yml`'s paths filter gains the `*server.xml`-bearing
directories, without which a PR editing only a `server.xml` would skip the job that
runs this guard.

Root `mvn clean install` under JDK 21: 5678 tests, 0 failures, 0 errors, 8 skipped.

## Unrelated, noted while verifying

`TransformServiceDelegationParityTest.serviceRequest_orchestratorMatchesServiceRequestService`
compares two independently built FHIR resources by full JSON serialization
(`assertSameResource`, line 151), so it is sensitive to any field that differs
between two calls made at different instants. It was seen red once on an earlier
round of this branch and has been green since, including on the full build above.

https://claude.ai/code/session_019cUavqo8rQ4QhhEHt3pRPj

### DIFF
--- a/.github/ci/ci.analyzer-harness.yml
+++ b/.github/ci/ci.analyzer-harness.yml
@@ -65,9 +65,9 @@ services:
       # Production keeps FileConfig.java defaults (5000ms poll, 3000ms stability).
       - BRIDGE_FILE_POLL_INTERVAL_MS=1000
       - BRIDGE_FILE_FILE_STABILITY_TIMEOUT_MS=500
-      # Bridge connects directly to app (not through proxy), so no /api/ prefix
+      # Bypasses the proxy; /api/OpenELIS-Global is Tomcat's own context path, not a proxy rewrite
       # HttpForwardingRouter appends /astm, /hl7, /csv per message protocol
-      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/OpenELIS-Global/analyzer
+      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/api/OpenELIS-Global/analyzer
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_USERNAME=admin
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_PASSWORD=adminADMIN!
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_INSECURE_TLS=true
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -48,7 +48,7 @@ jobs:
               - '.github/scripts/publish-checkpoints*.cjs'
               - 'tomcat/**'
               - 'volume/tomcat/**'
-              - 'demo/**'
+              - 'dev/tomcat/**'
               - 'install/installerTemplate/**'
 
   backend-verify:
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -46,6 +46,10 @@ jobs:
               - '.github/ci/ci.analyzer-harness.yml'
               - '.github/workflows/backend.yml'
               - '.github/scripts/publish-checkpoints*.cjs'
+              - 'tomcat/**'
+              - 'volume/tomcat/**'
+              - 'demo/**'
+              - 'install/installerTemplate/**'
 
   backend-verify:
     name: Build + Test
--- a/docs/dev_setup.md
+++ b/docs/dev_setup.md
@@ -67,8 +67,9 @@ Detailed instructions found
 
 1. Open terminal to the OpenELIS-Global-2 directory
 2. `docker compose up -d --build`
-3. Access the application at
-   [https://localhost:8443/OpenELIS-Global](https://localhost:8443/OpenELIS-Global)
+3. Access the application at [https://localhost/](https://localhost/). The
+   legacy UI and the REST base sit under
+   [https://localhost/api/OpenELIS-Global/](https://localhost/api/OpenELIS-Global/)
 4. Might need to dismiss a security warning
 
 ## Developing in Eclipse
--- a/docs/dev_setup.md
+++ b/docs/dev_setup.md
@@ -68,7 +68,7 @@ Detailed instructions found
 1. Open terminal to the OpenELIS-Global-2 directory
 2. `docker compose up -d --build`
 3. Access the application at
-   [https://localhost:8443/api/OpenELIS-Global/](https://localhost:8443/api/OpenELIS-Global/)
+   [https://localhost:8443/OpenELIS-Global](https://localhost:8443/OpenELIS-Global)
 4. Might need to dismiss a security warning
 
 ## Developing in Eclipse
--- a/docs/dev_setup.md
+++ b/docs/dev_setup.md
@@ -68,7 +68,7 @@ Detailed instructions found
 1. Open terminal to the OpenELIS-Global-2 directory
 2. `docker compose up -d --build`
 3. Access the application at
-   [https://localhost:8443/OpenELIS-Global](https://localhost:8443/OpenELIS-Global)
+   [https://localhost:8443/api/OpenELIS-Global/](https://localhost:8443/api/OpenELIS-Global/)
 4. Might need to dismiss a security warning
 
 ## Developing in Eclipse
--- a/projects/analyzer-harness/docker-compose.analyzer-test.yml
+++ b/projects/analyzer-harness/docker-compose.analyzer-test.yml
@@ -90,8 +90,8 @@ services:
       - BRIDGE_SECURITY_PASSWORD=${ANALYZER_BRIDGE_PASSWORD:-adminADMIN!}
       - BRIDGE_FILE_ENABLED=true
       - LOGGING_LEVEL_ORG_ITECH=INFO
-      # Bridge connects directly to OE (not through proxy), so no /api/ prefix
-      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/OpenELIS-Global/analyzer
+      # Bypasses the proxy; /api/OpenELIS-Global is Tomcat's own context path, not a proxy rewrite
+      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/api/OpenELIS-Global/analyzer
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_USERNAME=admin
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_PASSWORD=adminADMIN!
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_INSECURE_TLS=true
--- a/projects/analyzer-harness/docker-compose.analyzer-test.yml
+++ b/projects/analyzer-harness/docker-compose.analyzer-test.yml
@@ -90,8 +90,8 @@ services:
       - BRIDGE_SECURITY_PASSWORD=${ANALYZER_BRIDGE_PASSWORD:-adminADMIN!}
       - BRIDGE_FILE_ENABLED=true
       - LOGGING_LEVEL_ORG_ITECH=INFO
-      # Bypasses the proxy; /api/OpenELIS-Global is Tomcat's own context path, not a proxy rewrite
-      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/api/OpenELIS-Global/analyzer
+      # Bridge connects directly to OE (not through proxy), so no /api/ prefix
+      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/OpenELIS-Global/analyzer
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_USERNAME=admin
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_PASSWORD=adminADMIN!
       - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_INSECURE_TLS=true
--- a/src/main/resources/persistence/persistence.xml
+++ b/src/main/resources/persistence/persistence.xml
@@ -176,8 +176,8 @@
             <property name="hibernate.cfg_xml_file"
                 value="classpath:hibernate/hibernate.cfg.xml" />
             <!--
-                Tomcat deploys this WAR twice into one JVM (/OpenELIS-Global and
-                /api/OpenELIS-Global, see conf/server.xml), and each context gets
+                Tomcat used to deploy this WAR twice into one JVM (/OpenELIS-Global
+                and /api/OpenELIS-Global, see conf/server.xml), and each context got
                 its own classloader, EntityManagerFactory and Hibernate Search
                 backend over this one directory. A Lucene index allows a single
                 writer, so whichever context did not take the lock fails every
@@ -185,15 +185,16 @@
                 committed, so the caller sees a 500 for a patient that was
                 actually created. Patient search is served from the database, so
                 nothing reads this index; it is left unmaintained rather than
-                maintained unsafely. Rebuild it on demand with GET /rest/reindex,
-                and solve the double deployment before relying on it again.
+                maintained unsafely. Rebuild it on demand with GET /rest/reindex.
+                The double deployment is fixed, so the lock contention is gone, but
+                switching indexing back on is a separate untested change.
             -->
             <property name="hibernate.search.automatic_indexing.enabled" value="false" />
             <!--
-                And do not create or validate that index at start-up either: both
-                contexts try it, one loses the writer lock and logs an
-                OverlappingFileLockException on every boot. Nothing reads the
-                index, so there is nothing to manage.
+                And do not create or validate that index at start-up either: while
+                the WAR deployed twice, both contexts tried it and one logged an
+                OverlappingFileLockException on every boot. Nothing reads the index,
+                so there is nothing to manage.
             -->
             <property name="hibernate.search.schema_management.strategy" value="none" />
             <property name="hibernate.search.backend.directory.type" value="local-filesystem" />
--- a/src/main/resources/persistence/persistence.xml
+++ b/src/main/resources/persistence/persistence.xml
@@ -176,25 +176,15 @@
             <property name="hibernate.cfg_xml_file"
                 value="classpath:hibernate/hibernate.cfg.xml" />
             <!--
-                Tomcat used to deploy this WAR twice into one JVM (/OpenELIS-Global
-                and /api/OpenELIS-Global, see conf/server.xml), and each context got
-                its own classloader, EntityManagerFactory and Hibernate Search
-                backend over this one directory. A Lucene index allows a single
-                writer, so whichever context did not take the lock fails every
-                save with OverlappingFileLockException - after the row is
-                committed, so the caller sees a 500 for a patient that was
-                actually created. Patient search is served from the database, so
-                nothing reads this index; it is left unmaintained rather than
-                maintained unsafely. Rebuild it on demand with GET /rest/reindex.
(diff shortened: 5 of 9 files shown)


.github/ci/ci.analyzer-harness.yml
```diff
@@ -65,9 +65,9 @@ services:
       # Production keeps FileConfig.java defaults (5000ms poll, 3000ms stability).
       - BRIDGE_FILE_POLL_INTERVAL_MS=1000
       - BRIDGE_FILE_FILE_STABILITY_TIMEOUT_MS=500
-      # Bridge connects directly to app (not through proxy), so no /api/ prefix
+      # Bypasses the proxy; /api/OpenELIS-Global is Tomcat's own context path, not a proxy rewrite
       # HttpForwardingRouter appends /astm, /hl7, /csv per message protocol
-      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/OpenELIS-Global/analyzer
+      - ORG_ITECH_AHB_FORWARD_HTTP_SERVER_URI=https://oe.openelis.org:8443/api/OpenELIS-Global/analyzer
```
mherman22, line 70: The same repoint is still missing in `projects/analyzer-harness/docker-compose.analyzer-test.yml:94`, and I think it has to land before this merges. `projects/analyzer-harness/docker-compose.dev.yml:39` bind-mounts the `volume/tomcat/oe_server.xml` this PR fixes, so once `deployOnStartup="false"` is in, that stack's Tomcat no longer serves `/OpenELIS-Global` (your own before/after table measures it as a 404), while the bridge in the same stack is still configured with `https://oe.openelis.org:8443/OpenELIS-Global/analyzer`. Every ASTM/HL7/CSV message the bridge forwards then hits a path with no context behind it, so analyzer results stop reaching OpenELIS while both containers stay up and healthy. That is the stack `projects/analyzer-harness/README.md` documents and the one `scripts/deploy-to-testing.sh` drives against testing.openelis-global.org.

The "keep this to the double-deployment fix" reasoning reads differently for this file than for the doc paths: it is the identical one-line change you already made right here, and unlike `docs/dbprofiles.md` and friends it is not already-stale, it breaks *because of* this PR. The comment above it on line 93 also still says the thing this hunk just corrected.

On `volume/openelis-analyzer-bridge/configuration.yml:7`, I would leave it and drop it from the follow-up: its `openelis.url` key is not bound by anything. The bridge reads `org.itech.ahb.forward-http-server.*` (`tools/openelis-analyzer-bridge/src/main/java/org/itech/ahb/config/properties/HTTPForwardServerConfigurationProperties.java`), and CI mounts that same file while overriding only the env var, which is why the E2E lane is green here. So it 404s nothing; it is just dead config.

PR #4282, opened 2026-09-11
fix(OGC-1201): order entry — consolidated fix list

mozzy11:
Fixes OGC-1201 — the consolidated order-entry list, worked end to end.

**What changed**

*Release blockers*
- **AR** Editing an order and then re-entering Enter Order from the side nav no longer carries the previous order's lab number and patient onto the new form. The reset is decided by the URL, not by sticky edit-mode state. A scanned order is no longer wiped on arrival either.
- **AH/AI/AJ** A new order generates its lab number as the form opens instead of loading unsaveable. Generate is a real button: keyboard-reachable (it was an anchor with no `href` — WCAG 2.1.1 Level A on the only required field) and disabled in flight, so repeat clicks stop spending numbers.
- **J** QA Review shows one checklist. The second, unguarded one is gone; intake acceptance keeps the submit gate.

*Everything else on the list*
- **AL/AB/W** Ordering without a patient, and continuing past a failed acceptance check, are now recorded decisions — one `sample_order_override` row each, with reason, user and timestamp. EQA becomes a control on the shared order form (not a fourth lane) that declares no-patient honestly, replacing the sentinel `NULL NULL` patient every EQA order used to share. `PatientRequired` governs the manual override; EQA is not gated by it.
- **K** The lanes honour `PatientRequired` and `REQUESTER_REQUIRED`, which existed but were read only by the legacy JSP. `SampleEntryReferralSiteNameRequired` marks the field — that is all it does in the legacy screen it was written for, and nothing reads it server side — so it drives the asterisk that regressed, not a save gate.
- **R** A save is verified to have persisted; server messages are translated instead of printing `errors.no.sample`; the failure notice is mounted on the steps that discarded it.
- **A** Tests and panels can be added at Collect. **I** Rejection is reachable from Collect. **T** Order attachments are back.
- **V/AF** Order date, next-visit date, sampling location + "if other", fax/email validation, GPS capture, remember-site-and-requester, collection method, temperature, specimen origin.
- **Y** Programmes match by code, not a bare `vl` substring, and a configured questionnaire is no longer suppressed. **AM** Lab receipt is captured, not silently stamped. **AP** Holding time is shown, anchored on receipt. **O** Dates stamp in site-local time, not UTC. **AE** An unsaved order no longer claims to be saved. **AK** The blocked-save message names what is actually missing. **C** The seeded sample can be removed. **D** Collect-only copy no longer shows on collect-less lanes. **S** Blocked label prints fall back to a link. **Q** Consent gating is configurable, advisory by default. **U** Referred-out has a dashboard filter.

**Why**

The three lanes had copied the same entry-step code between themselves, so one defect in it was three. Most of the "missing" legacy fields turned out to be frontend-only gaps: `SampleItem` and `SampleOrderItem` already stored them and the save path already parsed them.

**How it was tested**

Frontend suite green: 282 files, 1954 tests, ~90 new. Each release-blocking fix was inversion-verified — fix reverted, test watched to fail, fix restored. Backend suite green: 6194 tests.

**Driven live against a running stack**, which is how four of the defects below were found at all. Two Playwright specs (19 tests) now cover the workflow permanently, and the API and database were checked directly:

- An EQA order saves with **no patient attached** — `sample_human.patient_id` is NULL and no sentinel patient is created — and the decision is recorded: `NO_PATIENT / EQA / user 1 / 2026-09-11 09:19:36`.
- Both Liquibase changesets present in `databasechangelog`, with the objects they create.
- `/rest/order-override` returns 404 for an unknown order and 400 for a manual override with no reason.

Four defects the green unit suite could not see, each found by driving the real application:

1. `SampleOrderItem` rejects undeclared properties, so the override fields returned **400 on every clinical save** the moment EQA or no-patient was ticked.
2. The consent gate could never fire: the property is published as `CONSENT_REQUIRED_FOR_COLLECTION` (the `Property` enum name) while the screen read the `site_information` spelling. The unit test passed because it mocked the configuration and agreed with the mistake.
3. The server still demanded a patient, so the no-patient feature was non-functional end to end.
4. The override entity mapped `lastupdated` where `BaseObject` maps `last_updated`, so reading an override threw.

**Notes — what is deliberately not here**

- **AN** dropped per the decision on the ticket: not functional today, goes with the legacy screen.
- **T, notification half** — the decision calls it a design session, not a port.
- **H** — the decision sequences retirement behind that notification capability landing, and says explicitly that switching the screen off before it does would lose result notification. The EQA deep link is redirected (W unblocked it); the other three open an existing order for modification, which the new lanes do not yet cover.
- **AQ, stepper half** — `VectorOrderComplete` is vector-specific (animal/pool summary, handoff to the V-03 identification worklist). Environmental has no equivalent downstream stage, so making the steppers identical would mean inventing an empty screen or deleting a working one. The copy half of AQ/D is fixed.
- **F** — confirmed as intended lane vocabulary rather than a defect: an environmental order searches for an organization, a clinical one for a site and provider.
- **AE** interacts with OGC-1051, in progress with herman — worth a look together.
- **EQA migration**: existing EQA orders still point at the sentinel patient row, and their historical normal/abnormal flags were evaluated against it. Both are data migrations needing a retention decision, and the ticket lists them as open.
- `D-057` is cited for lab-number generation but lives in FRS v4, which I could not read. Generation on open uses the reserving call, so an abandoned form consumes a number; a non-reserving preview exists but would hand the same number to two concurrent users. Say so if you would rather have only the affordance fix.


mozzy11 (commented):
Decision: Approve

Summary:
- Consolidates three duplicate order-entry reset implementations into a single `useNewOrderReset` hook that decides by URL, not stale context state
- Extracts `LabNumberField` component with keyboard-accessible button (was inaccessible anchor), auto-generates on new-order mount, and disables during flight
- Removes the duplicate unguarded "QA Checklist" tile from OrderQA, keeping only the per-specimen SampleAcceptanceReview as the single acceptance gate
- Fixes vector/environmental date stamping: `todayLocalIso()` and `currentLocalTime()` use local time instead of UTC's `toISOString()`, which was off by a day for non-UTC labs
- `SaveFailureNotice` now translates server-reported message keys (e.g. `errors.no.sample`) instead of displaying raw bundle keys

Verified:
- Frontend test suite green: 278 files, 1908 tests pass (ran in isolated worktree)
- 20 new tests cover each finding with inversion verification per PR description
- Prettier formatting clean on all changed files
- No backend changes; CI pending but this is frontend-only

Note: The console errors in the test run (`render error`, `Cannot read properties`) come from existing `RouteErrorBoundary.test.jsx` and `LoadDepartmentsGuard.test.jsx` - these are intentional error boundary tests that throw to verify error handling, not failures introduced by this PR.

mozzy11 (commented):
Decision: Approve

Summary:
- EQA and no-patient orders no longer fabricate a sentinel patient; decisions are recorded in sample_order_override with user/reason/timestamp
- Lab number generation moved to form-open (was broken), generate button is now WCAG-accessible (was untrappable anchor) and disabled during flight
- New-order reset keyed on URL params instead of stale edit-mode state; scanned orders no longer regenerate
- UTC date stamping replaced with site-local todayLocalIso() / currentLocalTime()
- Duplicate unguarded QA checklist removed; the per-specimen SampleAcceptanceReview remains the single acceptance gate

Verified:
- Frontend suite: 282 files, 1954 tests passed (isolated worktree)
- Backend compiles clean; CI: all checkpoints green including E2E (19 new Playwright tests)
- Liquibase 099/100 guarded (preConditions onFail=MARK_RAN), with rollback, no ID collisions against base branch
- Prettier clean on all changed frontend files

PR #4285, opened 2026-09-11
Fix published freezer migration checksum compatibility

pmanko:
The testing server stopped booting after #4136 added `MICROBIOLOGY_CRITICAL` to two freezer migrations that had already run. Liquibase correctly rejected their changed checksums. Fresh installs could pass while existing databases could not start.

Accept only the two known published freezer checksums and add migration 099 to apply the complete alert-type constraint to databases where 089/092 already ran. The newer published versions remain valid. Existing alerts are retained; accepting a checksum alone would not apply the missing constraint change.

The backend workflow now supplies the previous revision's real changelog to a generic compatibility test. It rejects removed or incompatible one-time changesets instead of reconstructing migration history from the candidate's SQL. Intentional `runAlways`/`runOnChange` changesets retain their semantics. This check runs in this repository's backend CI.

Validation:

- Java 21 WAR build and required Spotless/frontend formatting completed. Frontend formatter changes are excluded from this PR.
- Three PostgreSQL upgrade cases pass: fresh alert schema, the published freezer history, and the published microbiology history. They cover repeat execution, preserved records/checksums, supported and unsupported alert types, and rollback/reapply with microbiology alerts present.
- Full application changelog integration tests pass for both existing analyzer-data scenarios; three existing microbiology rollback tests pass.
- Historical compatibility passes against develop `5c1c33dc42743db3ae290ecd47fca568fba36938`. A negative control using the unfixed generated resources fails on the original checksum rewrite; restoring the fix passes.

Testing's operational reset to published develop `672c92a` is separate from this repair. A reset does not remediate upgrades for existing installations. No checksum clearing or live database-history edits are part of this change.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -71,6 +71,20 @@ jobs:
         with:
           repository: ${{github.repository}}
           submodules: recursive
+          fetch-depth: 2
+
+      - name: Preserve the prior revision's migration history for compatibility checks
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+        run: |
+          baseline="${BASE_SHA:-$(git rev-parse HEAD^)}"
+          if [ "$baseline" = "0000000000000000000000000000000000000000" ]; then
+            baseline="$(git rev-parse HEAD^)"
+          fi
+          git fetch --depth=1 origin "$baseline"
+          mkdir -p "$RUNNER_TEMP/published-migrations"
+          git archive FETCH_HEAD src/main/resources | tar -x -C "$RUNNER_TEMP/published-migrations"
+          echo "LIQUIBASE_BASELINE_DIR=$RUNNER_TEMP/published-migrations/src/main/resources" >> "$GITHUB_ENV"
 
       - name: check formatting
         run: mvn spotless:check
--- a/src/main/resources/liquibase/3.5.x.x/092-freezer-humidity-alert-type.xml
+++ b/src/main/resources/liquibase/3.5.x.x/092-freezer-humidity-alert-type.xml
@@ -5,6 +5,8 @@
     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.8.xsd">
 
   <changeSet id="3.5.0-092-add-freezer-humidity-alert-type" author="mherman22">
+    <!-- Accept the published freezer-only revision; 099 upgrades its constraint. -->
+    <validCheckSum>8:9874f4ac18e40bc491bab25d06051072</validCheckSum>
     <preConditions onFail="MARK_RAN">
       <tableExists schemaName="clinlims" tableName="alert"/>
     </preConditions>
--- a/src/main/resources/liquibase/3.5.x.x/base.xml
+++ b/src/main/resources/liquibase/3.5.x.x/base.xml
@@ -505,7 +505,7 @@
 
   <!-- Cold-storage monitoring hardening (issues #3743/#3904). Its first changeset redeclares
        chk_alert_type wholesale (086's union plus FREEZER_OFFLINE). 092 below redeclares it once more to
-       add FREEZER_HUMIDITY and is now the last word on that constraint - see the note there. -->
+       add FREEZER_HUMIDITY; 099 reconciles both published versions after these have run. -->
   <include relativeToChangelogFile="true" file="089-cold-storage-monitoring-hardening.xml"/>
 
   <!-- Acknowledgment notes on alert, so the detail modal's notes field is no longer dropped on acknowledge. -->
@@ -515,8 +515,11 @@
        existence makes 089's guarded insert MARK_RAN. -->
   <include relativeToChangelogFile="true" file="091-freezer-monitoring-enabled-value-type.xml"/>
 
-  <!-- FREEZER_HUMIDITY on the alert_type constraint. This is now the last changeset anywhere in
-       base-changelog.xml that may redeclare chk_alert_type - it inherits 089's ordering invariant. -->
+  <!-- FREEZER_HUMIDITY on the alert_type constraint. Keep this after earlier constraint declarations. -->
   <include relativeToChangelogFile="true" file="092-freezer-humidity-alert-type.xml"/>
+
+  <!-- Final union: reconcile databases that ran the published freezer-only versions of 089/092.
+       Any future constraint migration must preserve every type in this union. -->
+  <include relativeToChangelogFile="true" file="099-reconcile-freezer-microbiology-alert-types.xml"/>
 
 </databaseChangeLog>
--- a/src/test/java/org/openelisglobal/liquibase/PublishedMigrationCompatibilityTest.java
+++ b/src/test/java/org/openelisglobal/liquibase/PublishedMigrationCompatibilityTest.java
@@ -1,0 +1,55 @@
+package org.openelisglobal.liquibase;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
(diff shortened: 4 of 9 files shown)


src/main/resources/liquibase/3.5.x.x/base.xml
```diff
@@ -515,8 +515,11 @@
...
   <include relativeToChangelogFile="true" file="091-freezer-monitoring-enabled-value-type.xml"/>
 
-  <!-- FREEZER_HUMIDITY on the alert_type constraint. This is now the last changeset anywhere in
-       base-changelog.xml that may redeclare chk_alert_type - it inherits 089's ordering invariant. -->
+  <!-- FREEZER_HUMIDITY on the alert_type constraint. Keep this after earlier constraint declarations. -->
   <include relativeToChangelogFile="true" file="092-freezer-humidity-alert-type.xml"/>
 
+  <!-- Final union: reconcile databases that ran the published freezer-only versions of 089/092.
+       Any future constraint migration must preserve every type in this union. -->
+  <include relativeToChangelogFile="true" file="099-reconcile-freezer-microbiology-alert-types.xml"/>
```
mherman22, line 523: This lands after `057-alert-entity-ref` and `058-alert-type-union` (included at lines 412 and 415 of this file), and both of those redeclare `chk_alert_type` without `FREEZER_OFFLINE` or `FREEZER_HUMIDITY`. On an existing install that already ran the published 089/092 and holds an alert row of either type, the upgrade now gets past the checksum rejection and dies one changeset later instead: 057's `ALTER TABLE clinlims.alert ADD CONSTRAINT chk_alert_type` fails with SQLSTATE 23514 because that row violates the narrowed list, 099 never runs, and the server still doesn't boot.

I ran it against postgres:14.4. Seeded a schema with the published 089/092 history (the fixtures this PR adds), inserted one `FREEZER_OFFLINE` row, then applied 057, 058, 089, 092 and 099 in this file's include order. The run aborts with `check constraint "chk_alert_type" of relation "alert" is violated by some row`, and `databasechangelog` stops at orderexecuted 2, the two seeded freezer changesets. Drop the `FREEZER_OFFLINE` row and the identical run succeeds, which is why `FreezerAlertUpgradeTest` doesn't see this: its changelog holds only 089, 092 and 099, and the one alert it pre-seeds is `FREEZER_TEMPERATURE`, which 057 and 058 both keep.

The fix that matches what you are already doing here is to add `FREEZER_OFFLINE` and `FREEZER_HUMIDITY` to the constraint lists in 057 and 058 and give each changeset a `<validCheckSum>` for its published revision, the same treatment 089 and 092 get. I applied exactly that to a copy of the resources and the `FREEZER_OFFLINE` case then runs clean through to 099. It would also be worth extending `FreezerAlertUpgradeTest` to include 057 and 058 in its changelog and to seed a freezer-specific alert, so the path that actually broke is covered.

src/main/resources/liquibase/3.5.x.x/092-freezer-humidity-alert-type.xml
```diff
@@ -5,6 +5,8 @@
     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.8.xsd">
 
   <changeSet id="3.5.0-092-add-freezer-humidity-alert-type" author="mherman22">
+    <!-- Accept the published freezer-only revision; 099 upgrades its constraint. -->
```
mherman22: The `<comment>` a few lines below, at line 17, still says that no changeset anywhere in base-changelog.xml may redeclare `chk_alert_type` after this one, and 089's comment says the same about 092. 099 now does exactly that, so both read as an instruction to put future constraint work ahead of a changeset that no longer has the last word, which is the same ordering mistake that produced this PR. You have already moved that note correctly in base.xml; the two changeset comments should follow it.

Editing them is safe, in case that was the worry: in liquibase-core 4.8.0 `ChangeSet.generateCheckSum()` only walks the changes and the sql visitors, so `<comment>` sits outside it. I rewrote the comment text in a copy of this file and the checksum stayed at `8:bbd6b84cec065edab324893d98b543dd`, the same value develop's copy produces today.

src/test/java/org/openelisglobal/liquibase/PublishedMigrationCompatibilityTest.java
```diff
@@ -0,0 +1,55 @@
...
+            }
+            for (ChangeSet old : previous.getChangeSets()) {
+                ChangeSet change = current.get(identity(old));
+                assertNotNull("Published changeset was removed: " + identity(old), change);
+                if (change.shouldAlwaysRun() || change.shouldRunOnChange()) {
+                    continue;
+                }
+                assertTrue(
+                        "Published changeset changed without accepting its prior checksum: " + identity(old) + " ("
+                                + old.generateCheckSum() + "). Use a new migration for the schema change.",
```
mherman22, line 46: `old.generateCheckSum()` here does not read the baseline for the one changeset in the changelog that loads external SQL. In liquibase-core 4.8.0, `SQLFileChange.openSqlStream()` resolves the file through `Scope.getCurrentScope().getResourceAccessor()` rather than the accessor its changelog was parsed with, so under Surefire it reads `target/classes`, which is the candidate build. The affected changeset is `3.5.x.x/041-result-components.xml::OGC-937-component-backfill`.

I ran this comparison with the baseline resources in a directory and the candidate resources on the classpath, the way the job arranges them. Appending one line to the candidate's `041-component-backfill.sql` moved the *baseline* checksum from `8:434fbbf06bed32131fa54f8cf11dc1c2` to `8:167ae03989894ca5f639a6ed1f3e7793` in step with it, and `isCheckSumValid` still returned true. So a PR that edits that published .sql file, precisely the regression this test exists to stop, goes green.

Computing the published checksum inside a scope that carries the baseline accessor closes it:

```java
CheckSum published = Scope.child(
        Map.of(Scope.Attr.resourceAccessor.name(), (Object) previousResources),
        (Scope.ScopedRunnerWithReturn<CheckSum>) old::generateCheckSum);
```

With that in place the same edit is reported as a checksum change, and a run against `672c92a`, the baseline your CI actually used, stays clean. Worth noting the message string is built eagerly, so `generateCheckSum()` runs on every changeset regardless of the assertion, which is why this is not limited to the failing path.

.github/workflows/backend.yml
```diff
@@ -71,6 +71,20 @@ jobs:
         with:
           repository: ${{github.repository}}
           submodules: recursive
+          fetch-depth: 2
+
+      - name: Preserve the prior revision's migration history for compatibility checks
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
```
mherman22: `base.sha` is not the commit the branch started from; it follows the base branch. PRs 4269 and 4267 in this repo currently have a `base.sha` ahead of their merge-base, so once develop moves and a PR gets another push, the baseline holds changesets the branch has never carried and the test reports them as deletions.

That is already the situation here. Running the comparison with develop's current tip as baseline and this branch as candidate gives:

```
REMOVED: liquibase/3.5.x.x/099-order-consent-required.xml::099-order-consent-required-1::oe
REMOVED: liquibase/3.5.x.x/100-order-override-record.xml::100-order-override-record-1::oe
```

Both arrived with #4282, which merged about two hours after this PR's last backend run, so the next push here is likely to go red on `Published changeset was removed`, naming an author who removed nothing. With `672c92a`, the merge-base, the same run is clean.

The baseline that matches what the assertion means is the merge-base: fetch the base branch and use `git merge-base origin/$BASE_REF HEAD`. If you would rather keep comparing checksums against the newer base tip, which is the stronger half of this check, then scope only the removal assertion to the merge-base and let the checksum comparison run over whatever both revisions share.

PR #4294, opened 2026-09-14
fix(OGC-1186): file a multi-select result on the component it was entered on

mozzy11:
Fixes OGC-1186.

**What changed**
- A multi-select (or dictionary) value saved on a multi-component test is now filed on the component it was entered on, even when a sibling component offers the same dictionary entry.
- The sibling keeps its own type and value; no phantom result row appears on a re-save.
- A permanent Playwright spec orders a numeric + dictionary + multi-select test, ticks one option on the multi-select row, saves, and checks where the value landed; an integration test covers the binding for both directions and the legacy no-component case.

**Why**
- The save bound each selected dictionary value to the first option row matching test and value, with no component filter. On a test whose dictionary and multi-select components share entries, the first row belonged to the dictionary component, so the value landed there.

**How it was tested**
- New `MultiComponentDictionaryRoutingIntegrationTest` (3 tests, red without the fix); full backend suite green in an isolated copy; new e2e spec green on dev compose.
- Live on dev compose: reproduced on develop (value on the dictionary row, type coerced to M), then with the fix the `result` row binds to the multi-select component's option row and the worklist reads the dictionary row back empty and typed D.

**Notes**
- Legacy option rows without a component id keep the historic first-match behaviour, so single-component tests and older data are unaffected.


mozzy11 (commented):
Independent review by Claude

Decision: Approve

Summary:
- Adds componentId parameter to getTestResultsByTestAndDictonaryResult lookup so multi-select/dictionary values bind to the component they were entered on, not the first row matching test+value
- Backward-compatible: blank/null componentId falls back to the existing test-wide match for legacy data and single-component tests
- New integration test (4 cases) covers both directions, the component-aware lookup, and the legacy fallback
- New Playwright E2E spec creates a multi-component test, saves a selection on the multi-select row, and asserts the value lands on the correct component

Verified by execution: new MultiComponentDictionaryRoutingIntegrationTest (4 tests) green in isolated copy; CI green (backend, frontend, E2E including the new Playwright spec). Local dev compose had an unrelated npm dependency issue preventing manual live walk, but CI E2E run on the same stack confirms the fix.

PR #4296, opened 2026-09-14
fix(OGC-1187): stop batch test reassignment from asking for a null id

mozzy11:
Fixes OGC-1187.

**What changed**
- Batch test reassignment no longer asks the server for a sample type's tests or a test's pending analyses until an id has been chosen, so opening the page fires no request with a null id.
- The two provider endpoints answer a missing, blank or non-numeric id as a client error with a plain message, keep answering an unknown numeric id with an empty success, and log any remaining failure at error level instead of debug.
- A Playwright spec opens the page and checks that no provider call carries a null id and nothing answers 5xx, then picks a sample type and checks the tests arrive; an integration test covers the endpoints' id handling and a component test covers the mount-time guard.

**Why**
- Both effects ran on mount with their initial null state interpolated into the URL, and the endpoints only guarded against null and empty, so the literal string "null" reached id parsing and blew up as a 500 on every visit. The failure was logged at debug level, which is why it went unnoticed.

**How it was tested**
- Integration test 4/4, component test 3/3, full backend and frontend suites green in an isolated copy; e2e spec green on dev compose.
- Live on dev compose: reproduced the two 500s on mount, then with the fix the page opens with no provider call and no server error, choosing a sample type populates the test list, and the endpoints answer `null`, `abc` and blank ids with 400.

**Notes**
- The blank-id message used to read "Internal error, please contact Admin and file bug report"; it now names the parameter. Nothing in the UI shows that text.


mozzy11 (commented):
Independent review by Claude

Decision: Approve

Summary:
- Frontend guards prevent provider calls until sample type/test ids are selected (lines 130-131, 165-166 in JSX)
- Backend validates ids as numeric, returning 400 for "null"/non-numeric instead of 500 (AllTestsForSampleTypeProviderRestController:44-48, PendingAnalysisForTestProviderRestController:68-72)
- Error logging upgraded from debug to error level for actual failures
- Test coverage: 4/4 integration tests, 3/3 component tests, 2 Playwright E2E specs

Verification:
- Backend integration test: 4/4 passed in isolated copy
- Frontend component test: 3/3 passed
- Live test on dev compose: page loads with no provider calls until sample type chosen; bad ids return 400 with descriptive messages; valid numeric ids work correctly
- CI: all checks green

PR #4297, opened 2026-09-14
fix(OGC-1053): say what an incomplete report link is missing

mozzy11:
Fixes OGC-1053.

**What changed**
- A `/Report` link that is missing its `type` or `report` parameter now stays on the page and shows a warning naming what is missing, with links to Routine Reports and Study Reports, instead of silently landing on the Dashboard.
- A complete link opens the report exactly as before.
- A Playwright spec walks both links; the report guard unit test now asserts the message for the incomplete link and the untouched redirects of the other report screens.

**Why**
- Deep links to reports work; only a link missing one of the two parameters bounced to the Dashboard with nothing said, so a stale bookmark looked like a session or permission problem (the ticket's rescope of 2026-09-11).

**How it was tested**
- Report guard unit tests 6/6 (the new ones red without the change); full frontend suite green; e2e spec green on dev compose and red without the change.
- Live on dev compose: `/Report?type=patient` stays put with the warning and both links (Routine Reports lands on `/RoutineReports`); `/Report?type=patient&report=TBPatientReport` still opens under Routine Reports.

**Notes**
- The message is inline rather than a toast on the destination page, because the report list pages do not render the shared alert and an actionable toast would trap focus. Two new keys in `en.json` only.


mozzy11 (commented):
Independent review by Claude

Decision: Approve

Summary:
- Replaces silent Dashboard redirect with an actionable inline warning when /Report URL is missing type or report
- Uses Carbon InlineNotification and React Intl correctly; links to report lists are properly styled
- Unit tests (6/6) and Playwright E2E spec cover incomplete and complete link cases
- Removes unused useHistory and context destructuring

Verified by execution: frontend unit tests in isolated worktree (1970 pass), CI E2E checkpoint green. Live test blocked by pre-existing dev compose issue; CI E2E substitutes.

Issue #4191, opened 2026-09-04
Pathology Case View v2 — bench stage visibility, and blocks and slides as identified objects


caseyi:
## What this delivers

Today the Pathology Case View shows a case, but it cannot answer the two questions a histopathology lab asks all day:

1. **Where is this case right now?** The shipped `PathologyStatus` enum has stage values that do not match the bench (`CUTTING`, `SLICING`, `ADDITIONAL_REQUEST`), so the dashboard filter cannot tell grossing from microtomy.
2. **Which block is missing?** Blocks and slides are counts, not rows. "3 of 4 embedded" cannot name the fourth. `pathology_slide` has no `block_id` at all, so a slide cannot be traced to the block it was cut from — which `ISO 15189:2022` 7.2.6.1(g) requires.

This issue tracks the first two versions of the v2 design, which fix exactly those two things. After them, the lab can see where every case is, and every block and slide is individually identified, parented and traceable.

**Tracking surface.** This issue is the tracker. The Jira epic `OGC-264` mirrors it for the internal team, but progress, questions and review happen here.

---

## Read these first — in this order

Do not start from the FRS. It assumes you already know what physically happens to a piece of tissue, and none of it makes sense without that.

1. **[The Specimen Journey](https://digi-uw.github.io/openelis-work/designs/pathology/specimen-journey.html)** — one patient followed through all three anatomic-pathology benches, illustrated. ~15 minutes. Start here.
2. **[Contributor handoff](https://github.com/DIGI-UW/openelis-work/blob/main/designs/pathology/volunteer-handoff.md)** — what already exists in the codebase, the conventions, and which seams are deliberately not yours to build.
3. **[Case View Shell](https://github.com/DIGI-UW/openelis-work/blob/main/designs/pathology/case-view-shell.md)** — the ten structural elements all three case views share (`S-1` … `S-10`).
4. **[Pathology Case View v2 FRS](https://github.com/DIGI-UW/openelis-work/blob/main/designs/pathology/pathology-case-view-v2.md)** — 18 FRs, 32 ACs. The contract.
5. **[Slicing guide](https://github.com/DIGI-UW/openelis-work/blob/main/designs/pathology/pathology-case-view-breakdown.md)** — the slices below, in detail, with the dependency order.

**Reference implementations** — read the header comment block of the mockup before writing any UI:

[...]

PR #4298, opened 2026-09-14
feat(pathology): bench stages replace the legacy PathologyStatus values (OGC-264)

WodPachua:
# Pull Requests Requirements

      Issue number if applicable.
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide).
      project.

## Summary

Slice S1 of Pathology Case View v2 (#4191, epic OGC-264).

The old enum was not the bench: `CUTTING` duplicated grossing, `SLICING` used
non-standard wording, and `ADDITIONAL_REQUEST` parked a case in a fake stage so
it stopped reporting where its tissue was. The tiles and the dashboard filter
also disagreed about which statuses counted as in progress.

**Stages.** `PathologyStatus` becomes the eleven bench stages of FR-2.1; the
three names above are retired (AC-5). Changeset
`OGC-264-pathology-status-bench-stages` (new `liquibase/2.9.x.x/`, the folder
the FRS names) maps stored rows onto the stages that absorb them. Its rollback
is deliberately empty: the merge is lossy, and writing a retired name back
would recreate rows Hibernate cannot load.

**Per-deployment switches (FR-2.3, AC-7 storage and exposure).** Seven
`site_information` rows named `pathology.stage.<VALUE>.enabled`, on by default,
in the `resultConfiguration` domain so they show on Result Entry Configuration.
The four mandatory stages have no row and no constant. Each has a
`ConfigurationProperties` constant, a hardcoded default of `true` so a missing
row still visits the stage, and a line in the authenticated
`configuration-properties` endpoint, the only path from a setting to the
browser. As directed on the issue, `3.3.x.x/035` widens `site_information.name`
to `VARCHAR(60)` so the spec names fit rather than inventing shortened keys;
the inserts follow in the same folder because include order is execution order.
Nothing reads the switches yet.

**Tiles.** "Additional requests" counts cases holding a `pathology_request` at
`OPENED` (AC-6) instead of a status, so a case at `READY_PATHOLOGIST` with an
open request is counted on both tiles. Grouping as approved on the issue: in
progress is every stage but `READY_PATHOLOGIST` and `COMPLETED`. The dashboard
filter reads the same list, so list and tiles agree (the filter clause of
AC-4).

**A request raised from the case view is now stored as open.** The case view
never writes its `OPENED` fallback into the posted form, so the service stored
a null and the request read as neither open nor closed. That went unnoticed
while the dashboard counted the `ADDITIONAL_REQUEST` stage; with the tile
reading request rows, raising a request left it at zero. Changeset
`OGC-264-pathology-request-status-backfill` brings existing rows to the value
the screen has always shown for them.

**Localization.** Eleven `pathology.stage.*` keys in `en.json` through one
derived id, so the Stage column, filter and case view select read "Microtomy"
rather than `MICROTOMY`. Only `en.json` changes, plus one backend instruction
message.

## Tests

103 backend and 14 frontend tests, all green, plus three Playwright specs:
migrations, switches and the endpoint (17); tiles and the request flag (17);
stage rules and message keys (14); frontend units (14); Playwright (5). Both
controller suites refuse an unauthenticated call first. The request fix is
inversion-verified: reverting it fails its test with `expected:<OPENED> but
was:<null>`.

## Screenshots

https://github.com/user-attachments/assets/3ef425b7-ce37-4ec0-a19b-96996e771c89

A case opens at `Accessioned`, the filter offers the eleven stages by name, the
case moves across the tiles as its stage changes, a request raised against it is
counted while it keeps its real stage, and the seven switches are shown. Walking
this through by hand is what found the request defect.

## Related Issue

Part of #4191, slice S1. Epic OGC-264.

## Other

- **New cases now start at `ACCESSIONED`** instead of `GROSSING`. The FRS is
  silent on the initial value; FR-2.1 defines `ACCESSIONED` as "received and
  triaged, awaiting grossing", and the later "Begin grossing" transition needs
  an origin. A test pins it; one word reverts it.
- The backfill will raise the additional-requests tile on a deployment already
  holding requests saved without a status. They were always outstanding and
  always visible in the case view, only invisible to a count.
- The dashboard's stage filter has never had an accessible name (Carbon
  `Select` with `noLabel`). This PR only changes its option texts; an
  `aria-label` is a follow-up in this epic.
- For the FRS author, nothing to do here: six `pathology.stage.*` keys and
  `pathology.label.case` are marked "existing" but did not exist; FR-2.3 points
  at `common.properties`, which never reaches the browser; and FR-10.8 never
  states the status a request carries at creation, which is how that defect
  survived.


src/main/java/org/openelisglobal/program/util/PathologyStages.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import org.openelisglobal.common.util.ConfigurationProperties.Property;
+import org.openelisglobal.program.valueholder.pathology.PathologySample.PathologyStatus;
+
+/**
+ * The rules about the histopathology bench sequence that the dashboard, the
```
Agaba-derrick, line 14: Nit: consider dropping the (OGC-264) ticket reference from the javadoc. The rationale itself ("why the predicate is passed in rather than read from config") is valuable and should stay  but hardcoding ticket IDs into source comments means they go stale/dead once the tracker changes or the ticket is archived. Worth putting ticket refs in the PR description instead.

Agaba-derrick (dismissed):

src/test/java/org/openelisglobal/common/rest/DisplayListControllerStageFlagsTest.java
```diff
@@ -0,0 +1,195 @@
...
+         * for the refusal as a status rather than the redirect to /LoginPage the
+         * production chain's form login answers a browser with.
+         */
+        @Bean
+        SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
+            http.authorizeHttpRequests(auth -> auth.anyRequest().authenticated()).httpBasic(Customizer.withDefaults());
+            return http.build();
+        }
+    }
+
```
Agaba-derrick, line 93: authenticateAs("admin") here duplicates what BaseWebContextSensitiveTest's own @Before already sets up (same principal, same authorities). Harmless, but could be dropped for clarity, or kept as an explicit statement of intent if that's the preference

Agaba-derrick (dismissed):

src/main/java/org/openelisglobal/program/util/PathologyStages.java
WodPachua: Thanks @Agaba-derrick, happy drop them... had noticed this inlining ticket ids in existing comments, and thought its practice here... will drop them accross all files in this PR, except maybe the Liquibase changeSet ids like `OGC-264-pathology-status-bench-stages`, those would stay, since that is the id convention...

src/test/java/org/openelisglobal/common/rest/DisplayListControllerStageFlagsTest.java
WodPachua: yes! yes! you're right... let me remove it... thanks!

frontend/src/components/pathology/PathologyDashboard.jsx
```diff
@@ -70,10 +71,12 @@ function PathologyDashboard() {
...
-      // Filter out COMPLETED statuses and update the in-progress statuses state
-      const filteredStatuses = statusList
-        .filter((status) => status.id !== "COMPLETED")
-        .map((status) => status.id);
+      // Mirrors the backend dashboard tile grouping (everything except
+      // awaiting-review and complete counts as in progress) so the filter
+      // and the tiles never disagree about which cases are in progress.
+      const filteredStatuses = inProgressStageIds(
+        statusList.map((status) => status.id),
+      );
```
mherman22, line 79: Reusing the tile grouping for the initial filter also narrows the landing worklist, and I think that has to change before merge. It used to be every status except `COMPLETED`, so `READY_PATHOLOGIST` cases were listed; now they are not. `renderCell` only draws the pathologist's "Start" button on a `READY_PATHOLOGIST` row, and the tiles are not clickable, so after this a pathologist opens the dashboard, sees none of the cases waiting for them, and has no way to claim one until they change the Status dropdown by hand.

The FRS is explicit that this screen's filter *values* change but the screen is not redesigned, and the grouping approved on #4191 was for the count tiles. I would keep `inProgressStageIds` for the "In Progress" option and `inProgressStatusObjects`, and leave the `setFilters` call a few lines down on every stage but `COMPLETED`.

src/test/java/org/openelisglobal/pathology/PathologyStatusMigrationTest.java
```diff
@@ -0,0 +1,181 @@
...
+ * a migrated row loads again through the service, which a row on a retired name
+ * did not.
+ */
+public class PathologyStatusMigrationTest extends BaseWebContextSensitiveTest {
+
+    /**
+     * The SQL equivalent of the changeset's three update elements
+     * (001-pathology-status-bench-stages).
+     */
+    private static final String MAP_CUTTING_TO_GROSSING = "UPDATE clinlims.pathology_sample"
```
mherman22, line 35: Because the SQL is re-typed here, nothing ever executes `001` itself against rows: drop the `2.9.x.x` line from `base-changelog.xml`, or mistype a `<where>` in the changeset, and this suite still passes while upgrading deployments keep rows on a retired name. `src/test/resources/database-upgrades/README.md` asks for a fixture at the boundary rather than a per-migration runner, and `DatabaseUpgradeIntegrationTest` would then run the real changeset over data you insert (`before.id=OGC-264-pathology-status-bench-stages`, `before.file=liquibase/2.9.x.x/001-pathology-status-bench-stages.xml`, plus a data fixture); the same applies to the backfill copy in `PathologySampleOpenRequestsTest`. Not blocking, and in fairness the referral backfills that merged since that README skipped it too, but this is the migration #4191 singled out as the risky one.

src/main/java/org/openelisglobal/program/service/PathologyDisplayServiceImpl.java
```diff
@@ -45,6 +48,7 @@ public PathologyDisplayItem convertToDisplayItem(Integer pathologySampleId) {
         PathologySample pathologySample = pathologySampleService.get(pathologySampleId);
         PathologyDisplayItem displayItem = new PathologyDisplayItem();
         displayItem.setStatus(pathologySample.getStatus());
+        displayItem.setHasOpenRequests(hasOpenRequests(pathologySample.getRequests()));
```
mherman22, line 51: `requests` is a plain lazy `@OneToMany` with no batch size and no fetch join in `getWithStatus`, so this adds one SELECT per case on `/rest/pathology/dashboard`, which returns the whole filtered worklist unpaginated. I cannot find anything in the frontend that reads `hasOpenRequests`, yet AC-6 says a case with an open request "shows the open-request flag at its actual stage". Is rendering it S3's, or was a badge on the dashboard row meant to land here? If it is S3's, I would drop the call from `convertToDisplayItem` and keep it on the case view item until there is a consumer.

frontend/src/components/pathology/pathologyStages.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return intl.formatMessage({
+    id: stageDisplayKey(stageId),
+    defaultMessage: fallback ?? stageId,
+  });
+}
+
+export function inProgressStageIds(servedIds) {
+  return IN_PROGRESS_STAGES.filter((stage) => servedIds.includes(stage));
+}
```
mherman22, line 61: `IN_PROGRESS_STAGES` is derived from the hardcoded `PATHOLOGY_STAGES` copy, so a stage added to `PathologyStatus` and served by `DisplayListService`, but not added here, silently drops out of the in-progress grouping while the backend tile keeps counting it. That is the disagreement this module exists to prevent. The served list already arrives in the enum's declaration order, so filtering it directly needs no copy:

```suggestion
export function inProgressStageIds(servedIds) {
  return servedIds.filter(
    (stage) => stage !== "READY_PATHOLOGIST" && stage !== "COMPLETED",
  );
}
```

That leaves `IN_PROGRESS_STAGES` with no caller outside its own test, so it can go with it. Take or leave, but it removes the one place the two sides can drift.

src/main/java/org/openelisglobal/program/util/PathologyStages.java
```diff
@@ -0,0 +1,154 @@
...
+     * the same, so a case could sit in the filter's results while no tile counted
+     * it. Both now read this one rule.
+     *
+     * <p>
+     * {@code READY_PATHOLOGIST} is excluded because a case there is queued for a
+     * pathologist and is counted on the "awaiting review" tile instead, and
+     * {@code COMPLETED} because the case is finished. {@code UNDER_REVIEW} is
+     * included: a pathologist has taken the case out of that queue and is working
+     * it.
+     */
```
mherman22, line 60: Twelve lines of prose carrying three separate points here (the rule, the bug that prompted it, why each constant is in or out), and most of the new javadoc in this PR is the same shape. The bar we hold is two lines of description and one idea, with the history in the PR body where it will not go stale against the code. The part worth keeping is the bit that stops someone moving a constant:

```suggestion
    /**
     * The stages at which a case counts as work in progress (AC-4): every stage
     * but READY_PATHOLOGIST, which has its own tile, and COMPLETED.
     */
```

src/main/java/org/openelisglobal/program/dao/PathologySampleDAOImpl.java
```diff
@@ -37,6 +38,17 @@ public Long getCountWithStatus(List<PathologyStatus> statuses) {
         return count;
     }
 
+    @Override
+    @Transactional(readOnly = true)
+    public Long getCountWithOpenRequests() {
+        String sql = "select count(distinct ps.id) from PathologySample ps join ps.requests r where r.status = :status";
```
mozzy11, line 44: There is no predicate on the case's own stage here, so a finished case still counts. I set a case to Completed holding one OPENED request on a deployed build and `/rest/pathology/dashboard/count` returned `additionalRequests: 1` next to `complete: 1`. Worth either excluding COMPLETED here and in the backfill, or confirming on #4191 that the tile is meant to include closed-out cases.

src/main/resources/liquibase/2.9.x.x/002-pathology-request-status-backfill.xml
```diff
@@ -0,0 +1,37 @@
...
+    the rows already in the table to the same value the screen has been showing
+    for them all along, so that "this case has an outstanding request" can be
+    read from the request rows (FR-2.1, AC-6) rather than from a stage
+    of its own.
+    -->
+    <changeSet id="OGC-264-pathology-request-status-backfill" author="OGC">
+        <comment>Store the requests that were saved without a status as OPENED, the state the case view has always displayed for them</comment>
+        <update tableName="pathology_request" schemaName="clinlims">
+            <column name="status" value="OPENED"/>
+            <where>status IS NULL</where>
```
mozzy11, line 25: On a develop build every request the case view raises is stored with a null status, which I reproduced before applying this branch, so on a live deployment this `where` matches essentially every request row ever written, not just a stray few. Combined with the unfiltered count query, the new tile will open at roughly the number of cases that have ever recorded a request, including long-finished ones, and each has to be opened and closed by hand to clear it. Scoping the backfill to cases that are not COMPLETED would keep the tile meaningful on day one.

src/main/resources/languages/message_en.properties
```diff
@@ -8466,5 +8466,10 @@ instructions.validation.bulk.release = When enabled, the Validation page offers
                                            released individually after review.
 instructions.validation.retest.note = When enabled, a validator sending a result back \
                                            for retest must write a note for the bench.
+instructions.pathology.stage.enabled = When enabled, pathology cases pass through this \
```
mozzy11, line 8469: This sentence promises behaviour the slice does not ship yet. I set `pathology.stage.MICROTOMY.enabled` to false on Result Entry Configuration: it saved, `/rest/configuration-properties` served `false`, and the case view still offered Microtomy, accepted it and stored the case at that stage, with nothing shown as not applicable. Either hold the seven rows back until S3 reads them, or reword this to describe what the switch does today.

mozzy11 (commented):
Decision: Request changes

Summary:
- Cloned the branch, built it and deployed it on an isolated dev stack over a copy of a seeded database, then walked the dashboard, the case view and Result Entry Configuration in the browser and through the API.
- The spine of the slice is sound: eleven stages served in bench order, labels localised on the filter, the Stage column and the case-view select, and all ten changesets apply cleanly on both an upgrade and a fresh install.
- The request fix is real and the inversion holds: a develop build stores a newly raised request with a null status, this build stores OPENED.
- Two things block merge for me, both left inline: the backfill leaves the new tile permanently inflated on any existing deployment, and the seven stage switches do nothing while the admin page tells the administrator they do.
- I also reproduced the landing-worklist narrowing @mherman22 raised, so treat that one as confirmed rather than a second opinion.

Required changes:
- Scope the additional-requests tile, or say it is meant to include finished cases. A Completed case holding an OPENED request is counted, and the backfill turns every historically null request row into OPENED. Detail on `PathologySampleDAOImpl` and on the backfill changeset.
- Make the stage switches either do something or stop claiming to. Detail on `message_en.properties`.

Optional improvements:
- `pathology_request` carries only its primary key. The new count query and the per-row `getRequests()` both scan it, so an index on `(pathology_sample_id, status)` would pay for itself well before the N+1 does.
- `PathologyStageEnablementFlagsTest` seeds the seven rows in its own `@Before`, so it stays green if changeset `036` is dropped from the changelog, which is the thing it exists to protect.
- `docs/User Manual.md` still tells the reader to set a case to "slicing for slides".

Verified by execution: the stage list and its labels, the four tiles, the request lifecycle end to end through the UI and the database, the seven switches on Result Entry Configuration including a save, both Liquibase paths, and the column widening. Verified by reading only: the Playwright specs, which I did not run. Nothing here was measured on CI alone.

mozzy11 (commented):
Correction to my review above: I had the weighting wrong, so treat this as replacing it.

Decision: Comment (was: Request changes)

Neither of the two things I listed as required is actually a blocker, and I should not have filed them that way:

- **The additional-requests tile.** You disclose the backfill's effect in the PR body, and counting every case with an OPENED request is the rule agreed on #4191, which says nothing about excluding finished cases. It is also recoverable in the product: I closed the request on a Completed case and the tile went straight back to 0. So this is a question for @caseyi and @Agaba-derrick about what the tile should mean on an upgraded deployment, not a defect in your implementation. My inline note stands as that question.
- **The stage switches.** "Nothing reads the switches yet" is in your PR body, and landing a configuration row ahead of the screen that reads it is normal. What I actually found is narrower: the help text an administrator reads promises the stage is skipped and shown as not applicable, and today it is not. That is a sentence, not a slice.

What I would still hold the merge for is the landing worklist @mherman22 described, and only because I reproduced it: with one case at Ready for Pathologist the tile reads 1 and the table below it is empty, so a pathologist has nothing to claim until they change the Status dropdown themselves. That is his call to close, not mine.

Everything else in my review was verification rather than objection, and it came out well: the eleven stages, the localised labels, both Liquibase paths on a real upgrade and a fresh install, the seven switches saving from the admin page, and the request fix with its inversion confirmed against a develop build. Sorry for the noise on the verdict.

mozzy11 (approved):
Approving for merge, with two things recorded rather than blocked on.

Verified by running: the eleven stages served in bench order, the labels on the dashboard filter, the Stage column and the case-view select, all ten changesets applying cleanly on both an upgrade over seeded data and a fresh install, the column widening, the seven switches saving from Result Entry Configuration, and the request fix confirmed by inversion — a develop build stores a newly raised request as null, this build stores OPENED.

I pushed @mherman22's two suggestions (deriving the in-progress grouping from the served stages, and trimming the `inProgress()` javadoc), updated the two tests that named the removed constant, corrected the user manual, and brought the branch up to date with develop.

Carried forward instead of blocked:
- The landing worklist not listing cases awaiting a pathologist is now #4353, with the reproduction and the two options, because the obvious revert changes what the Status dropdown reads and that is a product call.
- Whether the additional-requests tile should count finished cases is a question for @caseyi and @Agaba-derrick on #4191; the count matches the rule agreed there, and the backfill's effect is disclosed in the PR description.

@Agaba-derrick's two earlier reviews were dismissed as stale, not as disagreed — both points were fixed in d209aa0a7 after they were written.

WodPachua:
thanks @mozzy11 @mherman22 🙏 ... my next tasks will handle the switch help text (the rail renders n/a for a disabled stage, so it becomes true rather than reworded), the hasOpenRequests badge or dropping the dashboard call, and the javadoc trimming across the module.

 And @mherman22 ... you're right that re-typing the SQL means the suite passes even if the changeset never runs, so taking the migration point for S4, will use the database-upgrades fixture there... since its where it touches most...

PR #4299, opened 2026-09-14
feat(OGC-1134): wait for two characters before searching the test catalog

mozzy11:
Fixes OGC-1134.

**What changed**
- The Test Catalog list search no longer asks the server for a single character; it waits until the term has two, then behaves as before (one request after the pause, stale requests aborted).
- A unit test covers the one-character, two-character and cleared cases; a Playwright spec counts the real requests while typing.

**Why**
- The debounce and the in-flight abort the ticket asks for have been in place since #3714, but a lone character still queried, which returns most of the catalog and is not a search yet.

**How it was tested**
- List unit suite 13/13 (the new test red without the change); full frontend suite green; e2e spec green on dev compose and red without the change.
- Live on dev compose: typing "a" sends nothing and leaves the URL alone; "am" sends one request; "amylase" typed in one go sends one more.

**Notes**
- The minimum is two characters, matching the ticket's suggested range. Clearing the box still refreshes the unfiltered list.


mozzy11 (commented):
Independent review by Claude

Decision: Approve

Summary:
- Adds a 2-character minimum before the Test Catalog search queries the server, preventing broad single-character matches
- Implementation is minimal and correct: early return from the debounce effect for length 1, normal flow otherwise
- Existing abort test updated from "x" to "xy" to accommodate the new minimum; new unit test covers the single-character, two-character, and cleared cases
- Playwright e2e test follows best practices (waitForRequest for sync, URL/UI assertions for state, no anti-patterns)

Verified by execution: frontend unit suite (13/13 pass in isolated copy), CI green (backend, frontend, E2E checkpoints). Verified by reading: effective diff, test coverage, Playwright test quality.

PR #4302, opened 2026-09-14
fix(home-dashboard): keep pagination next enabled when page size is 100

HerbertYiga:
The Home dashboard tile list stacked two paginations: a server-side one (fixed at 99 items per API page) and a Carbon Pagination whose totalItems was the size of the currently loaded server page. Choosing a page size of 100 left Carbon computing 1 page of 99 items with next disabled, so the user could not reach the rest of the data through that control.

Route Carbon next/previous through the server: infer the server page size from any non-last response, extend totalItems across all server pages, and translate the requested virtual page into a (server page, local page) pair, fetching a new server page when the virtual position moves past what is currently loaded. Fall back to the previous per-server-page behavior when a specific test section is narrowing the list, since the filtered total cannot be estimated across unfetched pages.


frontend/src/components/home/Dashboard.tsx
```diff
@@ -415,13 +432,95 @@ const HomeDashBoard: React.FC<DashBoardProps> = () => {
...
+  // page client-side, so we can't estimate a global total across unfetched
+  // server pages. Fall back to plain per-server-page Carbon pagination.
+  const isFilterNarrowing =
+    tilesWithTabs.includes(selectedTile?.type) &&
+    selectedTestSection !== "all" &&
+    selectedTestSection !== undefined &&
+    selectedTestSection !== "";
+
+  const hasMoreServerPages = totalApiPages != null && totalApiPages > 1;
+  const useVirtualPaging = hasMoreServerPages && !isFilterNarrowing;
```
mherman22: Page size 30 skips rows: `handlePageChange` resolves the server page from a contiguous item index while `virtualPage` maps back in fixed blocks, and the two agree only when a server page divides evenly into client pages (a full page here is 100 rows, since `createPages` adds one past the configured 99), so on a 400 row tile at 30 per page rows 190-199, 260-299 and 360-399 are never rendered and nothing throws.

```suggestion
  // The two page mappings agree only when a server page splits evenly.
  const useVirtualPaging =
    hasMoreServerPages && !isFilterNarrowing && serverPageSize % pageSize === 0;
```

frontend/src/components/home/Dashboard.tsx
```diff
@@ -133,6 +133,9 @@ const HomeDashBoard: React.FC<DashBoardProps> = () => {
   const [pagination, setPagination] = useState(false);
   const [currentApiPage, setCurrentApiPage] = useState(null);
   const [totalApiPages, setTotalApiPages] = useState(null);
+  // Inferred from any non-last server page (guaranteed full). Default matches
+  // Pagination can compute totalItems across unfetched server pages.
+  const [serverPageSize, setServerPageSize] = useState(99);
```
mherman22: This comment breaks off after "Default matches", and 99 is the wrong number anyway: `createPages` adds one item past the configured `results.pageSize`, so a full server page is 100 rows (the PR description carries the same slip).

```suggestion
  // Inferred from any non-last server page, which is always full. The default
  // is only a placeholder until the first response lands.
  const [serverPageSize, setServerPageSize] = useState(100);
```

frontend/src/components/home/Dashboard.tsx
```diff
@@ -415,13 +432,95 @@ const HomeDashBoard: React.FC<DashBoardProps> = () => {
...
+    : page;
+
   const handlePageChange = (pageInfo) => {
-    if (page != pageInfo.page) {
-      setPage(pageInfo.page);
+    const requestedVirtualPage = pageInfo.page;
+    const requestedPageSize = pageInfo.pageSize;
+
+    if (useVirtualPaging) {
+      const itemStart = (requestedVirtualPage - 1) * requestedPageSize;
```
mherman22: This mapping is pure arithmetic, so it is worth extracting and table-testing one case per offered page size in vitest; nothing in the repo exercises it today.

frontend/src/components/home/Dashboard.tsx
```diff
@@ -415,13 +432,95 @@ const HomeDashBoard: React.FC<DashBoardProps> = () => {
...
+      const requestedServerPage = Math.max(
+        1,
+        Math.min(
+          parseInt(totalApiPages),
+          Math.floor(itemStart / serverPageSize) + 1,
```
mozzy11: This maps a virtual page to a server page by flat item index, but `virtualPage` above maps back by block index (`(currentApiPage - 1) * clientPagesPerServerPage + page`). The two only agree when `pageSize` divides the server page size, and of the offered sizes [10, 20, 30, 50, 100] only 30 does not divide the server's 100: at 30 over a 310-row list I measured 50 rows unreachable (global rows 191-200 and 261-300), then blank pages with next still enabled. Deriving both directions from the block model fixes it, e.g. `requestedServerPage = Math.ceil(requestedVirtualPage / clientPagesPerServerPage)` with `totalItems` sized as `(totalApiPages - 1) * clientPagesPerServerPage * pageSize` plus the loaded last page.

frontend/src/components/home/Dashboard.tsx
mozzy11: The server's effective page is 100, not 99: `PatientDashBoardPaging.createPages` appends one more item after `resultCount >= getResultsPageSize()` flips, so page 1 comes back with 100 items (confirmed against the live endpoint). The default is harmless in practice because the first multi-page response overwrites it, but it is misleading, and the comment above reads as two sentences merged together.

frontend/src/components/home/Dashboard.tsx
```diff
@@ -415,13 +432,95 @@ const HomeDashBoard: React.FC<DashBoardProps> = () => {
     setSelectedTile(tile);
   };
 
+  // A specific test section is being applied. Filtering happens per server
+  // page client-side, so we can't estimate a global total across unfetched
+  // server pages. Fall back to plain per-server-page Carbon pagination.
+  const isFilterNarrowing =
```
mozzy11: For users without Global Administrator, `selectedTestSection` defaults to `sections[0]?.id`, so `isFilterNarrowing` is true on all eight `tilesWithTabs` tiles and those users keep the old behavior, including the page-size-100 symptom this PR is fixing. Worth calling that out in the description, since the fallback currently reads as an edge case rather than the default for most of the user base.

mozzy11 (commented):
Decision: Request changes

Summary:
- The headline fix works: at page size 100 all 310 rows across 4 server pages are now reachable, where before only the first page was.
- Page size 30 falls into the same class of bug this PR fixes: 50 of 310 rows are unreachable, and the control then emits blank pages with next still enabled.
- Cause: `handlePageChange` maps virtual page to server page by flat item index, while `virtualPage` maps back by block index; the two agree only when `pageSize` divides the server page size, and 30 is the one offered size that does not divide 100.
- Nothing tests the new arithmetic, which is exactly where the defect sits.
- Non-admin users keep the old per-server-page behavior on all eight test-section tiles, so the reported symptom is unfixed for them.

Required changes:
- Derive the virtual/server mapping from one model in both directions (inline at Dashboard.tsx:482).
- Add a unit test over the page-mapping math, covering each offered page size against a multi-page server list.

Optional improvements:
- `totalItems` reports "400 items" for a 310-row list until the last server page loads, so ranges read like "241-270 of 400". Clamping once the last page is known would keep the label honest.
- `serverPageSize` is not reset alongside `currentApiPage`, `totalApiPages` and `page` when the tile changes.

Verified by execution: ran the branch's Dashboard.tsx on the dev stack against a Ready For Validation list of 310 rows (4 server pages of 100/100/100/10) and walked every page size in the UI, identifying rows by React key against the API's ordering. Sizes 10, 20, 50 and 100 reach all 310; size 30 reaches 260. No console errors, and CI is green. Not verified: the narrowed-test-section fallback branch, since the rows I seeded all share one test section.

PR #4321, opened 2026-09-15
feat(caseView): shared case-view shell for the anatomic-pathology screens (OGC-1195)

WodPachua:
# Pull Requests Requirements

      Issue number if applicable.
      [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide).
      project.

## Summary

Slice S2 of Pathology Case View v2 (#4191, epic OGC-264, Jira OGC-1195).

**No screen changes here, and that is the point.** This adds the components the
three anatomic-pathology case views will be rebuilt on. Nothing renders them
yet; the pathology case view moves onto them in S3, then IHC and cytology.

Histopathology, IHC and cytology each carry their own copy of the
section-locking logic, the action bar and the summary panel, so an
accessibility fix is three changes in three places. Two of those designs also
got the state model wrong: one put a `caseReadyForReview` boolean beside the
case status enum and the two fell out of step, unlocking sections the sidebar
still showed as locked; another had no gating at all. The rules those failures
broke are enforced by the components rather than left to each screen.

| Component | What it enforces |
|---|---|
| `sectionState.js` | Exactly four states derived from case status, stage order, role and prerequisites. Nothing to store, and no `HIDDEN` state, since a section is never removed from the DOM for status or role. Takes an already-localized stage name, so a hint cannot print `READY_PATHOLOGIST` at a technician. |
| `CaseSection.jsx` | A disabled section always resolves to a sentence naming its unlock condition; "Locked" alone is unavailable. Children stay in the document in every state. |
| `CaseActionBar.jsx` | Discard, Save draft and **one** primary; a fourth cannot be passed. The primary is disabled **iff** given the reason it is disabled, as a `title` and via `aria-describedby`, since a disabled button is not focusable. |
| `CaseSummaryPanel.jsx` | Values arrive already derived; the panel never sums. `0` renders as `0`, never "not recorded". |
| `ProgressRail.jsx` | Carbon `ProgressIndicator`, as `MicrobiologyCaseView` already does. Pending work travels as text naming the objects it stands for. |
| `StatusBadge.jsx` | The seven-kind colour vocabulary in one place; a badge always renders its meaning as text. |
| `PatientHeader.jsx` | Optional `statusTag` and `assignedStaff`. With neither given the markup is byte-for-byte today's, down to the Carbon column span — ~30 screens render it. |

**Three places where Carbon's real behaviour decided the design**, each read in
its source rather than assumed. `AccordionItem` already renders a native
`<button>` with `aria-expanded`, drops a disabled header out of the tab order
and dims it with `cursor: not-allowed`, so none of that is reimplemented — and
it will not expand a disabled item, so a disabled section's reason lives in its
always-visible header rather than its body. `Tag` discards `title` on
non-filter tags, so the rail names pending work through Carbon's secondary
label. `ProgressIndicator` forces `complete` onto every step before its
`currentIndex`, which would have shown a completion tick for a stage the
deployment does not track — `COVERSLIPPING` is the stage most often switched
off and sits before the last three — so the rail passes `-1` and carries the
per-step flags itself.

**i18n.** Thirteen `caseView.*` keys, `en.json` only, in the shell FRS's group
vocabulary and reusing the mockup's key names where it had them. `Status`
reuses `common.status`.

## Test plan

- [x] `npx vitest run` — 293 files, 2078 passed, 8 skipped, 0 failed
- [x] 106 new tests across 9 files: `CaseActionBar` 19, `CaseSection` 18,
      `sectionState` 18, `StatusBadge` 15, `ProgressRail` 11, `PatientHeader` 8,
      `CaseSummaryPanel` 7, `CaseViewLayout` 6, `useUnsavedChangesGuard` 4
- [x] Each file pins its rule with a test that fails when the rule is reverted;
      the rail's suite was checked by breaking the component seven ways
- [x] `PatientHeader`'s layout guard verified by hand: forcing the narrowed
      span makes it fail, restoring it makes it pass
- [x] `prettier --check` clean; `eslint` clean on every file added or changed

## Screenshots

None, and no video: nothing here is reachable from the UI yet, so a recording
would show an unchanged screen. S3 puts the case view on these components and
carries the walkthrough.

## Related Issue

Part of #4191, slice S2. Epic OGC-264, Jira OGC-1195.

## Other

Worth a reviewer's opinion:

- **The rail uses Carbon's `secondaryLabel` where the shell FRS says "count
  badge".** The FRS names a widget; the constitution says Carbon components by
  name, and microbiology already builds this rail with `ProgressIndicator`. The
  naming text ends up visible as well as in the step's accessible name.
- **`CaseActionBar` requires a Router.** The unsaved-changes warning needs both
  halves: `beforeunload` for a reload or closed tab, react-router's `Prompt`
  for a move to another screen, which is how a half-written case is actually
  lost. All three case views are routed screens.
- **Two near-duplicate keys minted for sentence case**:
  `caseView.action.saveDraft` against `button.save.draft`, and
  `caseView.badge.readOnly` against `label.readonly`. Both existing keys belong
  to other screens. Happy to reuse them instead.
- **Deferred, not forgotten**: the rail calls `onNavigate(id)` and leaves
  scrolling to the screen, as microbiology does (each section takes an `id`);
  and `Save draft` has no disabled state until a terminal status exists at
  sign-out.
- **Not in scope**, per the shell FRS: prior-results panel (S-2), status
  transitions and reopen (S-8), critical-result hook (S-9), slide list and
  deactivation (S-10).


WodPachua:
Hlloo @mozzy11 @Agaba-derrick ... S2 of the pathology case view is ready for your review, #4321.

heads up that this one changes no screen at all deliberately... it's just the shared components the three anatomic-pathology case views get rebuilt on, pathology moves onto them in S3 and IHC and cytology follow... so there's no video, a recording would just show an unchanged page.

@reagan-meant @mherman22 I request and welcome your reviews too when you get a moment...

frontend/src/components/caseView/CaseSummaryPanel.jsx
```diff
@@ -0,0 +1,52 @@
...
+const CaseSummaryPanel = ({ rows = [] }) => {
+  const intl = useIntl();
+
+  return (
+    <Tile className="case-view__summary">
+      <h2>
+        <FormattedMessage id="caseView.label.caseSummary" />
+      </h2>
+      {rows.map(({ id, labelKey, value }) => {
+        const isEmpty = value === null || value === undefined || value === "";
```
mozzy11, line 31: `false` is not covered here, and React renders nothing for it, so a boolean row shows its label with an empty value: neither the value nor the "not recorded" hint. I rendered `{ value: false }` against this component and got just the label back, where `{ value: 0 }` correctly gives `0`. Adding `false` to the same guard the `0` test already protects would close it.

frontend/src/components/common/PatientHeader.test.jsx
```diff
@@ -0,0 +1,160 @@
...
+  it("survives an explicit null for the assigned staff", () => {
+    const { container } = renderHeader({ assignedStaff: null });
+
+    expect(container.querySelector(".patient-name")).toHaveTextContent(
+      "Lovelace Ada",
+    );
+    expect(screen.queryAllByTestId("case-assigned-staff")).toHaveLength(0);
+  });
+
+  // This component is rendered by roughly thirty screens, so the two new props
```
mozzy11: It is six, not thirty: the pathology, cytology and immunohistochemistry case views, Modify Order, the results viewer and the generic program case view. Worth correcting, because the next reviewer will size their regression pass off this number. The layout guard itself is the right test and it held up when I rendered the header on two of those screens.

mozzy11 (commented):
Decision: Comment

Summary:
- Cloned the branch and put its `frontend/src` on a running dev stack, because "no screen changes" is the one claim in a PR like this that has to be checked rather than accepted.
- It holds. I rendered the pathology case view and Modify Order against this build and diffed the patient header with the same screens on the base: identical text, identical Carbon spans down to the `col-span-1` / `col-span-15` pair, and no `case-view__*` element anywhere in the DOM.
- Full frontend suite on the branch is 312 files and 2304 tests green, the 106 new ones included, and the nine new files each pin their rule rather than restate it.
- `Prompt` is fine, in case anyone wonders on the way past: this repo is on react-router-dom 5.3.4, and the action bar's router tests are the strongest thing in the PR.
- Two small things left inline, neither blocking; I would take this as the foundation for S3.

Optional improvements:
- The summary panel drops a `false` value on the floor, which matters once a report-signed-out style row exists. Detail inline.
- The "roughly thirty screens" figure in the header test is six. Detail inline.
- Because every leaf component is unreferenced, `caseView.scss` has never been loaded by a browser, so the sticky summary, the action-bar border and the flex header row are the one part of this PR that no test and no screen has exercised. Worth a deliberate look when S3 first renders them rather than trusting them then.

Verified by execution: the full vitest suite, the nine new test files, the summary panel's empty-value handling driven directly, and the patient header rendered on two live screens against a base-branch baseline. Verified by reading only: the components nothing imports yet, which by their nature cannot be exercised in a browser in this PR.

mozzy11 (approved):
Approving. Verified by deploying this branch's frontend on a dev stack and rendering the patient header on the pathology case view and Modify Order against the same screens on the base: identical text and identical Carbon spans, and no shell component in the DOM. Full frontend suite green at 312 files and 2304 tests.

I pushed one commit correcting the consumer count in the header's layout guard comment, and brought the branch up to date with develop. The `false` value in the summary panel is left as an inline note for S3, since how a boolean should render there is a design call rather than a defect to patch blind.

WodPachua:
Thanks @mozzy11...

PR #4326, opened 2026-09-16
fix(OGC-1221): cancel every analysis on a sample item removed in Edit Order

mozzy11:
Fixes OGC-1221.

**What changed**
- Edit Order "Remove Sample" now cancels every test on the removed sample item, not just the one on the ticked row, so nothing is left live on the Results worklist for a withdrawn sample.
- Other sample items on the same order are untouched.

**Why**
- The cancel logic detected the start of each sample-item group by a non-null accession number, but the form sends an empty string on every row after the first, so the group ended at the ticked row. Grouping is now by sample item id, which every row carries.

**How it was tested**
- New integration test with two analyses on the removed item plus a third on another item (inversion-verified: fails on the old code); full backend and front-end suites green locally.
- New Playwright spec `ogc-1221-edit-order-remove-sample.spec.ts` walks the ticket's steps and reads the worklist back; it fails on the unfixed build and passes on the fix.
- Live on dev compose: ticket steps through the UI, `POST /rest/SampleEdit` and the analysis rows read back from the database.

**Notes**
- Front end left as is: the backend no longer relies on the accession-number sentinel, so the empty string it sends is harmless.
- Two smaller UI nits mentioned in the ticket (duplicate `id="labNumber"`, second "Next" control from table pagination) are not part of this change.


mozzy11 (commented):

Decision: Approve

Summary:
- Fix correctly groups rows by `sampleItemId` instead of relying on accession-number sentinel (which was empty on non-first rows)
- Two-pass algorithm: first pass collects removed sample item IDs, second pass cancels all analyses on those items
- Integration test covers the exact bug scenario (two analyses on same item, one on another) with inversion verification
- Playwright E2E test validates the full user flow through the UI

Verification:
- Backend integration tests: 7/7 pass (ran locally in isolated rsync copy)
- CI status: green (E2E checkpoint passed)
- Code review: logic is sound, HashSet prevents duplicate sample item cancellation, second-pass ensures all analyses on removed items are cancelled regardless of row position
- Live test limited by dev environment data (no editable samples available), but CI E2E validates the flow

Optional improvements:
- None identified

PR #4327, opened 2026-09-16
feat(OGC-1193): load one test per catalog row with all its specimens, on every restart

mozzy11:
Fixes OGC-1193.

**What changed**
- A `tests.csv` row with `sampleType=Serum|Plasma|Whole Blood` now loads as one test, described by its plain name and linked to all three specimens (the OGC-1145 model the Test Catalog editor uses), written through `TestCatalogCreationService`.
- Identity is `localCode`, then the plain name (exact, then normalized description); a translation never identifies a test, so re-loading a file with or without `localization:xx` columns updates the same records.
- Records an earlier loader wrote as `Name(SampleType)` are recognised by `LegacyTestVariantFinder` (shared by the tests and panels handlers): the name is compared in its normalized form, so the shipped `HIVVIRALLOAD(Serum)` is a variant of the row `HIV Viral Load`, and the parenthesised part must be a known sample type. Variants whose specimen the row lists are updated and linked in place, take the row's spelling (`HIV Viral Load(Serum)`), keep their result definitions, and no plain-named duplicate is created; a variant for a specimen the row does not list (`Albumin(Urines)` for an `Albumin` row on Serum) is a different test and is left alone.
- Every row runs in its own transaction; a rejected row is skipped with its line number and reason while the rest of the file loads. Sample-type rows are checked against the 40/10 column limits before the database sees them.
- New optional `tests.csv` columns: `isReportable` (default Y), `notifyResults`, `localCode` (10 characters, unique).
- Panels resolve members by plain name, including through legacy variants. Each of the three files ends with `SUMMARY file=... domain=... created= updated= skipped=`.

**Why**
- Reproduced on develop: a localized multi-specimen row became a single `Name(<last specimen>)` test, an unlocalized one fanned out into N tests, panel members by plain name were dropped, and one 13-character abbreviation lost every later row of its file.

**How it was tested**
- New `CatalogCsvLoaderIntegrationTest` against a real database (row isolation, one test per row with links and flags, identical and column-reduced re-loads, rename by local code, legacy variants by exact and by normalized name with their numeric result kept, panel resolution, SUMMARY lines); `LegacyTestVariantFinderTest`; handler unit tests updated; full backend suite green locally.
- Live on the dev compose stack through `POST /rest/configuration/domains/reload` with a repro catalog, read back from the database and the Test Catalog UI.
- The analyzer-harness E2E (`ogc-1054-analyzer-mvp.spec.ts`) exercised the normalized case: its `HIV Viral Load` row must land on the shipped `HIVVIRALLOAD(Serum)` record, whose two-decimal result definition the FILE results display.

**Notes**
- The FRS (`test-catalog-csv-loader-v2-frs.md`) and gap analysis referenced by the ticket are not in `openelis-work`; this follows the ticket text, the Confluence implementer guide (v5) and the code. Decisions a reviewer may overrule: a collapsed legacy record (one `Name(X)` for the row, X listed) is linked to every listed specimen; multiple legacy variants each keep their own specimen; variants for unlisted specimens are not touched; a row whose local code matches an existing test renames that test to the row's name.
- `TestSectionServiceTest.getUserLocalizedTesSectionName` asserted `""` only because `TestSectionServiceImpl` builds its name cache once at startup; the integration test drives `TestSectionConfigurationHandler`, which refreshes that cache as production does. The test now refreshes the cache after loading its dataset and asserts the dataset's value.
- No Playwright spec: the loader has no UI path and the CI stack cannot receive catalog files at test time; the DB-backed integration test is the permanent regression cover.
- The implementer guide's §6.1, §6.7 and §8 describe the superseded behaviour and will need a refresh once this lands.


mozzy11 (commented):

Decision: Approve

Summary:
- Catalog loader now produces one test per CSV row linked to all listed specimens, fixing the fan-out bug that created duplicate `Name(SampleType)` records.
- Identity is by `localCode` first, then exact/normalized description; translations never identify a test, preventing the unique-index violation on re-load.
- Legacy `Name(SampleType)` records are recognized via `LegacyTestVariantFinder` using normalized comparison, updated in place, and keep their result definitions.
- Row-level transactions (`RowTransactionRunner`) let a rejected row fail alone with a logged reason while the rest of the file loads.
- SUMMARY line and `CsvLoadSummary` provide greppable outcome counters; panels now resolve plain names through legacy variants.

Verified by execution:
- Build: `mvn clean install -DskipTests -Dmaven.test.skip=true` succeeded.
- Tests: Full backend suite (6326 tests) passed with 0 failures, 0 errors.
- Live: Test Catalog UI at `/MasterListsPage/TestCatalogList` loads 213 tests with no console errors.

Could not verify:
- The catalog-reload API (`POST /rest/configuration/domains/reload`) was not exercised live; the new `CatalogCsvLoaderIntegrationTest` covers the loader logic against a real database.

PR #4329, opened 2026-09-16
feat(OGC-1194): carry the whole test catalog in CSV, import it from the admin page, and queue the names it cannot resolve

mozzy11:
Fixes OGC-1194. Builds on OGC-1193 (#4327).

**What changed**
- **CSV contract.** `tests.csv` takes `reportingName`, `domain`, `amr`; five new domains load through the Test Catalog editor's own services, so file and screen validate the same way: `result-components` (210), `result-limits` (320), `terminology` (320), `sample-handling` (320), `reflex-rules` (330, conditions and actions in one flat file, a rule identified by its name).
- **Shared loader shape.** `AbstractCatalogCsvHandler` + `CsvRow`: columns addressed by name, one unit of work per record, per-record transactions, a skipped record named with its line and reason, and the `SUMMARY file=... created= updated= skipped=` line. Lab units (`test-sections`) now load a row at a time too.
- **Import page.** `Admin → Import Catalog (CSV)`. Preview evaluates every row against the current catalog in a transaction it rolls back (the plan is the outcome, not an estimate); Apply stores the files under the configuration tree and loads them, so start-up and the page remain one mechanism. Files name their own area (`result-limits-cphl.csv` → `result-limits`), and areas load in dependency order however they were dropped. `POST /rest/configuration/import/{preview,apply}` (multipart, ADMIN).
- **Needs your decision.** a lab unit, specimen, test or result component the files name but the catalog does not know becomes an item in `unresolved_reference` (folded by spelling, counted, pointing at the `configuration_import_run` that noticed it) instead of a log line. The panel on the import page points it at a record, or skips it; "Remember this name" writes a `reference_alias` that resolves it by itself next time. Start-up, the reload API and the page are all import runs.
- **Apply reads only what was uploaded.** It used to force every file of the touched areas, so adding four small files re-read the whole shipped catalog: 35s on the dev stack, past the UI's own timeout. `ConfigurationReloadOptions` now carries the file names to force, and applying the same four files takes 7s.
- **Two fixes found on the way**: a numeric reflex condition that is not `BETWEEN` parsed a second bound it never has (every such rule failed to save), and `AppTestConfig`'s `"org.openelisglobal.config.*"` exclusion is a regex, not a package prefix, so it was also excluding `org.openelisglobal.configuration.*` from the test context.

**Why**
- An implementer bringing up a site had to re-key everything the CSV could not carry (CPHL: 29 reflex rules, every range, every result component) and needed a server administrator for each load. A misspelled specimen silently dropped its rows.

**How it was tested**
- `CatalogAttributeCsvLoaderIntegrationTest` (real database): components created then merged by code; ranges, terminology and handling through the editor services with blank cells leaving stored values alone; a reflex rule built from its rows and re-loaded by name; a preview computing the same outcomes and keeping nothing; an unknown specimen queued, folded on a second sighting, then resolved as an alias that the next load uses by itself.
- Frontend `CatalogImport.test.jsx` (4): area inferred per file, Preview before Apply, a rejected file blocking Apply, and a decision remembered only when asked.
- Playwright `foundational/core/ogc-1194-catalog-import.spec.ts` (2): preview keeps nothing then apply puts the test in the catalog; an unknown specimen is decided on the page and the remembered name resolves the same file.
- Full backend suite green; live on the dev stack through the page and the reload API, read back from the database.

**Notes**
- The FRS (`test-catalog-csv-loader-v2-frs.md`) and the gap analysis are not in `openelis-work`; this follows the epic text, the implementer guide and the code. Decisions a reviewer may overrule: `result-limits` treats a test's rows as its complete numeric ranges (the editor's own semantics) while `result-components`, `terminology` and `sample-handling` merge; a reflex rule is identified by `ruleName`; Apply forces a reload so a re-upload of unchanged content still applies.
- Order-intake hold state (OGC-1145 P1b) is not in this PR: the queue and its panel are built so it can join them, but the intake side is a separate change.
- A Feature Doc for the epic is not written yet; the catalog domains, the SUMMARY line and the decision queue are documented in `volume/configuration/README.md`.


mozzy11 (commented):

Decision: Approve

Summary:
- CSV catalog import from Admin page with preview/apply workflow and unresolved name queue
- Well-architected backend with proper layered structure: 3 new entities, DAOs, services, and 5 new configuration handlers
- Frontend follows Carbon Design System and React Intl patterns correctly
- CI green including new E2E test (ogc-1194-catalog-import.spec.ts), integration test (6 cases), and frontend tests (4 cases)
- Liquibase changeset 103 is properly guarded with preConditions

Required changes:
- None blocking

Optional improvements:
- ResultComponentConfigurationHandler.java:150 - add null check before setSignificantDigits(row.integer(...))
- Consider documenting the DAO @Transactional placement as an existing codebase pattern (violates constitution text but matches all existing DAOs)

Verified by execution: isolated rsync copy build (SUCCESS), backend tests including CatalogAttributeCsvLoaderIntegrationTest (36 tests pass), frontend CatalogImport.test.jsx (4 tests pass). CI verified E2E on live stack. Structural checks: no submodule changes, no workflow changes, Liquibase 103 properly included in base.xml.

mozzy11:
Disposition of the two optional review items, for the record:

1. Null check before `setSignificantDigits(row.integer("significantDigits"))` in `ResultComponentConfigurationHandler`: not needed. The call sits inside `if (!row.isBlank("significantDigits"))`, and `CsvRow.integer` returns null only for an empty value, which that guard already excludes; a non-numeric value throws `IllegalArgumentException` with the column name, which the per-row transaction turns into a skipped row with that reason. The same shape is used for `displayOrder`.

2. DAO `@Transactional` placement: left as is, matching every existing DAO in the codebase. Changing it here would make this one file inconsistent with its neighbours; it belongs in a separate pass across all DAOs if the team wants the constitution text enforced.

PR #4332, opened 2026-09-17
fix(OGC-1220): consolidate analyzer tests and audit history

pmanko:
Analyzer integration tests previously replaced internal history and parsing services, and shared fixture helpers could commit outside test rollback or block on their own locks. This foundation runs those internal services for real, separates isolated controller coverage, and consolidates fixture transaction ownership.

- Preserve numeric and UUID audit references, including guarded migration rollback. Migration evidence remains recorded for the commits on which it ran.
- Make fixture load, cleanup and legacy helper calls share the active transaction. Keep explicit standalone cleanup atomic, and consolidate sequence advancement without reusing allocated IDs.
- Replace analyzer-result paging upper bounds with exact membership, order and boundary assertions. Give status, analyzer-result, QC-chart and QC-dashboard fixtures rollback ownership; restore dashboard names after rollback and remove QC reference-row repairs.
- Advance analyzer/profile/mapping/revision sequences in the fixture loader; remove per-test repairs and the unused analyzer cleanup helper.
- Allow first branding access to commit defaults, with a database assertion after the request; this production correction is a separate commit.
- Correct affected callers: independent-transaction referral checks use committed setup, sample tracking uses the inherited database wiring, and practitioner timestamp boundaries use the fixture's actual instant.
- Align active guides, root agent instructions and reusable templates with the actual runners and unit/component/integration/end-to-end boundaries. Preserve the critical-result browser check's exact color requirements.

Latest local validation on `c7df5bf49f`: all 133 explicitly selected classes pass in alphabetical and reverse alphabetical order, with 1,237 executed checks passing and one pre-existing ignored generic-sample test per run. This includes every current analyzer core class: 57 classes and 312 checks in each order. The meaningful sequence collision and dashboard/referral leak reproductions fail before correction; the nine-class correction run passes all 83 checks. Scoped formatting and the clean build pass. Published source blobs match the hashes captured during validation. The exact class selection and reproduction commands are committed with the roadmap. Earlier evidence remains historical rather than a rerun on this head.

**Foundation remediation remains incomplete.** Remaining result-review/acceptance/QC-result fixture owners, catalog seed repairs and retained service/cache dependencies still need correction. The older CI result-option failure's ambient-catalog dependency remains open even though it passed this selection. Three preserved T2 lifecycle/query/security classes will extend the selection when restored. New-head full CI, browser acceptance and required maintainer approval remain pending; no dependency pins changed.

The [single analyzer roadmap](https://github.com/DIGI-UW/OpenELIS-Global-2/blob/feat/ogc-1054-held-result-reprocessing/specs/roadmaps/ogc-1054-analyzer-feature-roadmap.md#analyzer-test-remediation-execution-plan) owns the findings, exact evidence and remaining work. This PR depends on #4241. Real mapping lifecycle and automatic held-result recovery remain separate milestones. No merge or deployment is included.


pmanko:
Heads-up from the OGC-479 reporting stack: **this PR is load-bearing for us too**, not just analyzer work. Sharing the evidence in case it helps prioritise #4241's review.

### The failure

Reporting stack PRs fail intermittently with 25 `analysis_status_fk` violations, all carrying the `-1` sentinel that `IStatusService.getStatusID()` returns when a name cannot be resolved:

```
insert into clinlims.ANALYSIS (... status_id ...) values (... ('-1'::int4) ...)
  type_of_sample_name = 'Microbiology integration specimen'
  violates foreign key constraint "analysis_status_fk"
```

Affected classes are always the same five and always unrelated to the change under test: `MicrobiologyOrderEligibilityIntegrationTest`, `SearchResultsServiceTest`, `TestCatalogEditorRangesIntegrationTest`, `CatalogAttributeCsvLoaderIntegrationTest`, `ColdStorageTimestampWireShapeTest`.

### Root cause

`executeDataSetWithStateManagement` TRUNCATEs every table a dataset names and reinserts only that dataset's rows. Of the 46 fixtures declaring `status_of_sample`, **45 declare an incomplete catalog** and 27 name no ANALYSIS status at all. Whichever suite runs after one of those sees a catalog missing the name it asks for.

It is latent on `develop` purely because ordering never lines up a clobberer with a victim. Adding test classes shifts Surefire ordering and exposes it. Proof it is ordering rather than code: the same SHA `2594ba24f7` both passed (run 35757598379) and failed (run 35757734310).

### Why your approach is the right one

We tried a per-load repair in our stack, an additive `ensureStatusCatalog()` modelled on the existing `ensureAuditSystemUser()`. It made things **worse**: full suite went from 4 errors / 2 failures to **34 errors / 14 failures**, newly breaking `StatusOfSampleServiceTest.getPageOfStatusOfSamples` plus several FHIR facade and validation suites. Reverted.

The lesson is that the approach cannot work, not that the implementation was poor. Tests legitimately assert on the catalog's contents, so any repair that **adds** rows changes what they observe; and adding the table to `PROTECTED_SEED_TABLES` **removes** fixture-declared rows, breaking the two fixtures that hardcode banded ids. Every per-load repair mutates shared state some test depends on.

Making the fixture load transactional so the TRUNCATE rolls back is the only approach that satisfies both constraints at once: during a test the fixture sees exactly what it declared, and afterwards the seeded baseline is intact for everyone else. Your diff comment names it exactly:

> A separate committed connection defeats `@Transactional` rollback and leaves unrelated suites with truncated seed tables.

No action needed from you. We are not adding anything to this file and will rebase onto a `develop` containing this once #4241 and this land, so there should be no conflict from our side.

https://claude.ai/code/session_01748DR4rHZgxgK3RAw7oKgw

PR #4334, opened 2026-09-17
fix(OGC-1209): file each panel under its own domain and say when its specimens fall outside it

mozzy11:
Fixes OGC-1209.

**What changed**
- A panel's domain is **set on the panel**, never derived from the specimens its member tests use. `panels` CSV takes an optional `domain` column (CLINICAL / ENVIRONMENTAL / VECTOR); an unknown value skips the row with its line and reason, and omitting the column leaves the stored domain alone.
- The shipped seed now carries it, each panel filed under the domain holding most of its specimens: 18 vector panels, 1 environmental, the rest clinical. The Environmental and Vector filters return rows instead of an empty table.
- The Panel list marks a panel that reaches outside its domain. `GET /rest/test-catalog/panels` returns `sampleTypesOutsideDomain`, and the row shows a `Mixed domain` tag naming those specimens on hover, the same passive shape the Tests list uses for a test with no LOINC.
- Environmental and Vector are selectable in the panel editor's Basic Info, and the screen note now describes the model instead of promising those domains "in a later phase".

**Why**
- `panel.domain` defaults to CLINICAL and nothing ever set it: not the CSV loader, not the OCL mapper, not either legacy create path. So all 23 panels read Clinical while 18 of them carry only vector specimens, and two of the three filter options could never match anything.

**How it was tested**
- `PanelConfigurationHandlerTest` (11): the domain column is applied, is case-insensitive, an unknown value skips the row, and no column leaves the panel untouched. Inversion-verified.
- `PanelsList.test.jsx` (11) and `PanelEditor.test.jsx` (8): the warning appears with the offending specimens on hover, is absent when the panel is consistent, and every domain is selectable.
- Playwright `foundational/core/ogc-1209-panel-domains.spec.ts` (2).
- Full backend and frontend suites green; live on the dev stack: after a panels reload the Vector filter returns 18 rows and Environmental 1 (both were 0), and the warning fires on a real case, `Bilan Biochimique`, whose member test reaches a Serum belonging to the vector domain.

**Notes**
- The ticket's open question is answered by the product owner: explicit, not derived. The seed values follow their rule (the domain where a panel has most of its sample types).
- The coverage the ticket cites (`TC-PANEL-08` / `TC-PANEL-09` in `test-catalog-admin-list-screens.spec.ts`, one marked `test.fail()`) is **not in this repository** and neither is any `test.fail()`; those specs were never committed. The new spec above is the permanent cover.
- Unrelated pre-existing gap, left alone: every vector panel has zero member tests because the tests its seed names are absent from the shipped catalog. The OGC-1194 decision queue records them. Their Sample Types column therefore stays empty on a stock database.
- `savePanelBasicInfo` still rejects a domain that disagrees with a **member test's** domain (422, pre-existing from OGC-224). That guard is about tests, not specimens, and is untouched here.


mozzy11 (commented):

Decision: Approve

Summary:
- Adds `domain` column to panels CSV loader with validation and case-insensitive parsing
- Exposes `sampleTypesOutsideDomain` from panel list endpoint, enabling the "Mixed domain" warning in UI
- Enables Environmental/Vector domain radio buttons in panel editor (previously disabled with later-phase note)
- Updates seed CSV with explicit domain values for 18 vector + 1 environmental panels

Required changes:
(none)

Optional improvements:
(none)

Verified by execution: CI green (E2E checkpoints), frontend tests (292 files, 2031 passed), `PanelConfigurationHandlerTest` (11 tests). Local live test blocked by pre-existing Vite import error in dev container (not from this PR). Code read: all changed files, effective diff.

PR #4335, opened 2026-09-17
fix(OGC-1222): keep order entry alive when a Program row is deleted

mozzy11:
Fixes OGC-1222.

**What changed**
- **The endpoint survives a dead cached id.** `UserServiceImpl.getUserPrograms` resolves each cached program with `getMatch` and drops an id that resolves to nothing. `get` raises `ObjectNotFoundException` for a missing row rather than returning null, which is what produced the 500.
- **A failed request is no longer data.** `getFromOpenElisServer` reports `undefined` when `response.ok` is false, instead of handing the caller Spring's error body. That is the contract the newer callers already assume.
- **Both program consumers ignore a non-list payload**, so a bad response leaves an empty picker rather than throwing in the next render.
- **The three order lanes are wrapped in the existing `RouteErrorBoundary`**, so one failing section degrades to an inline error instead of unmounting the whole page.
- **`ProgramSelect` picks its default by code, not by the literal name "Routine Testing"** (from the ticket's follow-up comment), so a site that renames or translates that program keeps its default.

**Why**
- Deleting any Program row at the database level blanked Clinical, Environmental and Vector order entry until a restart: the cached list kept the dead id, resolving it threw, the 500 body reached the page as data, and `programs.find` on a non-array unmounted the React tree.

**How it was tested**
- `UserProgramsTest` (3, inversion verified), `ProgramSelect.test.jsx` (3, inversion verified), `ProgramSection.test.jsx` (+2), `Utils.test.ts` (+1).
- Playwright `foundational/core/ogc-1222-order-entry-survives-program-failure.spec.ts` injects a 500 on the endpoint and asserts each of the three lanes still renders.
- Full backend and frontend suites green (frontend 2037 passed).
- Live on the dev stack, program deleted with the application running and no restart: `/rest/user-programs` went from 500 to 200 with the deleted program absent, and the Environmental order page went from an empty root to a rendered form with the Program field; the console error `programs.find is not a function` is gone.

**Notes**
- The ticket's analysis says `programService.get()` returns null for a deleted id. It throws instead, so a null check alone left the 500 in place; the live check is what caught this, and the unit test now pins the real contract.
- Deactivating a program from the admin screens (the "Preserve, Don't Delete" follow-on in the ticket, needing an `is_active` column and a UI action) is deliberately **not** in this PR.
- The new fetch contract is a global behaviour change. Verified against the full frontend suite, and the two attachment consumers that now receive `undefined` on a 404 already guard with `Array.isArray`.
- Programs are auto-created from JSON under `/var/lib/openelis-global/programs` at every start-up and matched by code, so a deleted row that has a file reappears on restart. That is unchanged here and worth knowing when reproducing.


mozzy11 (commented):

Decision: Approve

Summary:
- Backend fix correctly uses `getMatch` (returns `Optional`) instead of `get` (throws `ObjectNotFoundException`) for cached program IDs that may no longer exist in the database.
- Frontend `getFromOpenElisServer` now reports `undefined` on HTTP error responses instead of passing the error body as data, preventing downstream `find()` crashes.
- Both `ProgramSelect` and `ProgramSection` guard against non-array payloads with `Array.isArray()`.
- `RouteErrorBoundary` wraps the three order entry lanes so a failure degrades gracefully.
- Default program selection now matches on `code` rather than display name, supporting sites that rename/translate programs.

Verified by execution:
- `UserProgramsTest` (3 tests), `ProgramSelect.test.jsx` (3 tests), `ProgramSection.test.jsx` (12 tests), `Utils.test.ts` (6 tests) all green in isolated copy.
- Live test on dev stack: inserted program id=999, restarted to populate cache, deleted the row, called `/rest/user-programs` - returned HTTP 200 with the deleted program absent (the bug would have returned 500).
- CI status shows all Playwright, Cypress, and analyzer-harness suites passed.

Could not verify:
- Frontend live UI test skipped because the dev frontend container image is stale (missing `@tanstack/react-query`); backend API test confirmed the fix.

PR #4339, opened 2026-09-18
feat(OGC-1223): give a provider a title, chosen from a list an administrator keeps

mozzy11:
Fixes OGC-1223.

**What changed**
- A provider has a **Title** (Dr, Prof, Sr, HEO), chosen from a list rather than typed into a name box. It sits before First Name on the form, shows as its own column in the provider list, and is optional.
- **Admin → Provider Management → Provider Titles** maintains that list: title, abbreviation, sort order, in-use count, status and text filters, add/edit modal. **No delete** — a title is deactivated and the providers carrying it keep it, with the confirmation naming how many.
- **Search still matches names only.** Typing "Dr" returns nobody; the title is a separate filter, whose total follows the filter so paging cannot run past its own result.
- **One display helper** renders `Dr John Kila` everywhere a provider name is a single string (order entry dropdown, provider typeahead, printed report), replacing three backend and four frontend concatenations.
- FHIR carries it as `Practitioner.name.prefix` both ways, and the scheduled practitioner import now copies it.

**Why**
- A Provider record was only a first and last name, so the rank went into one of them: the same clinician appeared three ways, surname sorting interleaved people recorded as "Dr", and a report printed "Dr Dr John Kila". CPHL traced most of its duplicate providers to this one missing field.

**How it was tested**
- `ProviderDisplayNameTest` (10) and the frontend `providerDisplayName` (11) pin the rendering rules, including every missing-part case; `ProviderTitleMenu.test.jsx` (8) covers the page, its filters and that deactivation names the count and never deletes.
- Playwright `foundational/core/ogc-1223-provider-titles.spec.ts` (3) asserts the seeded list with no delete, the title on the provider row, that searching "Dr" returns nobody while filtering by it does not, and the titled typeahead name.
- Full backend and frontend suites green. Live on the dev stack: both changesets in `databasechangelog`, the column and category created, six titles seeded, a provider saved with `Dr` and read back from the database, the typeahead returning `Dr John Kila`, search for "Dr" returning 0, and the filter returning 1 of 4 with a total of 1.

**Assumptions a reviewer may overrule**
- **Dictionary category over a dedicated table**, as the FRS recommends: the category endpoint already exists, seeding is a CSV drop, and localisation and audit come free. A `provider_title` table would need an entity, DAO, service, DTO and controller first.
- **The column is on `person`, not `provider`.** The FRS puts storage on the shared record with only provider surfaces in scope, so a later patient or requestor surface needs no migration. `person` is Hibernate Search annotated, but an unindexed column does not affect the index mapping.
- Provider Management was a flat side-nav link; it becomes a group so Provider Titles can sit under it, as the FRS describes.

**Notes**
- Out of scope per the FRS: migrating titles already sitting in name fields, merging duplicate providers, a Patient-side surface, and any permission derived from a title.
- Unrelated pre-existing defect found while testing: `POST /rest/Provider/FhirUuid` answers **500** when a name contains a digit. The new spec works around it with a letters-only name; worth its own ticket.


mozzy11 (commented):

Decision: Comment

Summary:
- Feature implementation is solid: Liquibase migrations, dictionary category, FHIR support, and display helper all work correctly
- Backend tests (10) and frontend tests (19) pass; CI E2E suite green
- TypeScript compilation reveals 4 issues in ProviderMenu.tsx that need attention before merge

Required changes:
- ProviderMenu.tsx:558,656: Add `InlineNotification` to Carbon imports
- ProviderMenu.tsx:45,67: Add `titleCode?: string` to ProviderPerson and ProviderTableRow interfaces
- ProviderMenu.tsx:328-339: Initialize `setTitleCode(provider.titleCode || "")` in openUpdateModal
- ProviderMenu.tsx:314-322: Reset `setTitleCode("")` in openAddModal

Optional improvements:
- ProviderMenu.tsx:811: Duplicate `key` prop warning (low priority)

Verification: Backend build and unit tests run in isolated copy; frontend tests run via vitest; Liquibase changesets confirmed in fresh DB (3 rows in databasechangelog, 6 titles seeded). Live UI test blocked by auth on dev stack but CI E2E passed.

mozzy11:
All four required items are in `b7b7be0c0`, and walking the changed screens in a browser to verify them turned up three more faults that every suite had passed.

- `InlineNotification` imported; the FR-7 warning now renders with a clean console.
- `titleCode` added to `ProviderPerson` and `ProviderTableRow`; `providerDisplayName` converted to `.ts`.
- `openUpdateModal` initialises the title, `openAddModal` resets it. Both forms are mounted at once and shared `id="providerTitle"`, so the update control is now `updateProviderTitle`.
- Rejecting a duplicate showed "Failed to save": the reason arrives as a JSON string and the helper's parse failed and replaced it. Both save paths now read the response and unwrap it.
- The order-entry requester card is rebuilt from the order data, which was not carrying the title, so a selected provider lost it. The title now travels with the order data.

The duplicate `key` prop at `ProviderMenu.tsx:811` is the `key` + `getHeaderProps` idiom already on develop (line 702 there), so I have left it out of this diff.

Tests: 2074 frontend unit tests pass (11 new, each inversion-verified), and the e2e spec grew two cases covering the duplicate message and the edit round-trip; all 6 pass on the dev stack. The controlled/uncontrolled React warning in the update modal comes from `setTelephone`/`setFax` without a fallback, which also predates this PR.

Issue #4261, opened 2026-09-10
Cold storage: 10 findings from a runtime verification of #3743 / #3904 on develop


mherman22:
## Cold storage: findings from a runtime verification of #3743 / #3904 on `develop`

Both reported issues were verified **fixed** on `develop` at `5c1c33dc4` (the merge of #3828). The
verification then continued across the rest of the cold-storage surface, and this issue collects
everything it turned up. Nothing here reopens #3743 or #3904.

Every finding below was measured against a running stack, not read off the code. Where a claim is
only about code shape, it says so.

<details>
<summary><b>How this was verified</b> (environment, so the numbers can be reproduced)</summary>

- WAR built from `develop` @ `5c1c33dc4` under JDK 21, `sha256 c9c486d8…`, bind-mounted into the
  published `itechuw/openelis-global-2:develop` image and confirmed byte-identical inside the
  container. Frontend from `itechuw/openelis-global-2-frontend:develop` (built 06:49Z, after the
  09:43+03:00 merge; confirmed to contain the merged bundle by grepping it for i18n keys the merge
  added).
- Stack: [`openelis-distro-coldstoragedemo`](https://github.com/mherman22/openelis-distro-coldstoragedemo)
  — Modbus-TCP simulator, SMTP sink, 20 s poll, three seeded devices in "Demo Cold Room".
- **Run with `TZ=Africa/Kampala` (UTC+3) on purpose.** On a UTC stack a stripped timezone offset is
  indistinguishable from a correct one, so #3743's timestamp fix cannot be confirmed there.
- Evidence read from four channels per observation: HTTP wire, `volume/logs`/container logs scoped
  with `--since`, the `clinlims` database, and the browser console.
- **Caveat:** the database was carried over from a run the previous day. The cold-storage changesets
  had already applied under the same ids and filenames `develop` ships (`089`/`090`/`091`/`092`) and
  Liquibase raised no checksum error, so behaviour is valid — but nothing here demonstrates that
  `develop` installs *cleanly from scratch*.

**Verification result, for context:**

| Issue | Item | Result |
|---|---|---|
[...]

PR #4343, opened 2026-09-19
fix(coldstorage): resolve reports, NPE, and warning text (#4261)

dharmendra26-wiz:
# Pull Requests Requirements

- [x] The changes confirm to the OpenElis Global x3 [Styleguide and Design](https://uwdigi.atlassian.net/wiki/spaces/OG/pages/621346838/OpenELIS+Global+Style+Guide) documentation.
- [x] I have read and agree to the Contributing [Guidelines](https://openelis-global.org/community/get-involved/) of this project.

## Summary

Fixes multiple items identified during Cold Storage QA:
- **Reports:** Fixed `REPORT_TYPE_MAP` so Weekly and Monthly logs download correctly instead of defaulting to Daily.
- **Audit Trail:** Added null checks to date filters in `FreezerAuditTrailController` to prevent 500 errors on single-sided date searches. Updated excursion text to generic alert text.
- **Translations:** Updated `coldStorage.device.deleteWarning` in `en.json` to reflect current deletion behaviour.

## Screenshots

N/A (Backend logic, translation updates, and file generation mapping).

## Related Issue

Addresses #4261


frontend/src/components/coldStorage/Reports.jsx
```diff
@@ -48,8 +48,8 @@ const EXPORT_FORMATS = ["PDF"]; // CSV, XML, and Excel to be implemented in Phas
 
 const REPORT_TYPE_MAP = {
   "Daily Log": "freezerDailyLogReport",
-  "Weekly Log": "freezerDailyLogReport",
-  "Monthly Log": "freezerDailyLogReport",
+  "Weekly Log": "weeklylog",
+  "Monthly Log": "monthlylog",
```
mherman22: The mapping itself is right, `FreezerReportServiceImpl` takes `weeklylog`/`monthlylog` and the Jasper template switches bands on the `Weekly Log`/`Monthly Log` display name, so the three reports really are three reports now. Nothing exercises it though. `Reports.test.jsx` already mocks `downloadReportDirect`, so picking Weekly Log in the dropdown, clicking Generate and asserting `downloadReportDirect` got `reportName: "weeklylog"` is a handful of lines and fails on `develop`. Without it, the next person editing this map can silently collapse all three onto one value again, which is exactly how it got here.

Small thing while you're in this map: the service also accepts `dailylog`, so making the Daily entry `"dailylog"` would leave all three values in one convention rather than the current mix of a camelCase report id and two lowercase ones.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
```diff
@@ -151,8 +151,10 @@ public ResponseEntity<List<Map<String, Object>>> getAuditTrail(@RequestParam(req
                     if (alert.getAcknowledgedAt() != null) {
                         OffsetDateTime ackTime = alert.getAcknowledgedAt();
                         // Apply date filter only if dates are provided
-                        boolean includeAck = (startDateTime == null && endDateTime == null)
-                                || (!ackTime.isBefore(startDateTime) && !ackTime.isAfter(endDateTime));
+                        boolean includeAck = true;
+                        if (startDateTime != null && endDateTime != null) {
+                            includeAck = !ackTime.isBefore(startDateTime) && !ackTime.isAfter(endDateTime);
+                        }
```
mherman22: This fixes the NPE, but with only one bound supplied it now returns every acknowledgement ever recorded, at 200, with nothing to say the filter was dropped. So `?start=2026-09-01T00:00:00Z` looks like it worked and quietly isn't scoped, which for an audit trail is a worse failure than the 500 was: a 500 gets reported, a silently unfiltered result gets believed. Honoring each bound on its own is the same two lines:

```suggestion
                        boolean includeAck = (startDateTime == null || !ackTime.isBefore(startDateTime))
                                && (endDateTime == null || !ackTime.isAfter(endDateTime));
```

Same shape for the resolve branch below. The freezer-history and corrective-action branches have the identical gap (they were the model for this form), so it's worth carrying the per-bound check to all four while the file is open.

Worth a test either way, and there's a slice for it: `SecuritySliceMockMvcTest`, as used by `FreezerConfigControllerSecurityTest`, will let you pin `?start=` alone at 200 with an out-of-range event absent from the body. Nothing covers this path today. Note the blast radius is small for now, `fetchAuditTrail` in `api.js` never sends dates at all and filters client-side, so only a direct API consumer sees either behavior.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
```diff
@@ -162,7 +164,7 @@ public ResponseEntity<List<Map<String, Object>>> getAuditTrail(@RequestParam(req
                             ackEvent.put("actionType", "ALERT_ACKNOWLEDGED");
                             ackEvent.put("performedAt", ackTime.toString());
                             ackEvent.put("performedBy", getUserName(alert.getAcknowledgedBy()));
-                            ackEvent.put("comment", "Temperature excursion alert for " + freezerName + " acknowledged");
+                            ackEvent.put("comment", "Alert for " + freezerName + " acknowledged");
```
mherman22: Dropping "Temperature excursion" fixes the wrong half of the sentence and loses the right half. The Details column in the audit table renders `comment`, so the row now reads "Alert for Demo Freezer acknowledged", which the Action and Freezer columns on the same row already say. An auditor still can't tell from the row whether an over-temperature, a humidity breach or an offline device was acknowledged, which is what #4261 item 8 was asking for. `alertType` is on the entity, so `alert.getAlertType() + " alert for " + freezerName + " acknowledged"` keeps the row informative and makes it true for all three types.

Related, and in the same block: `actionType` a few lines down is still hardcoded `CRITICAL_ALERT_RESOLVED` for every resolution regardless of severity, and `formatActionLabel` in `Reports.jsx` renders that as "Critical alert resolved". A resolved WARNING excursion is therefore still labelled critical in the UI. Same defect as the comment strings, so probably the same fix round.

frontend/src/components/coldStorage/Reports.jsx
mherman22, line 52: The mapping is right, but nothing tests it. `Reports.test.jsx` already mocks `downloadReportDirect`, so picking Weekly Log and asserting `reportName: "weeklylog"` is a few lines and fails on develop. A missing test is how all three ended up on one value in the first place.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
mherman22: This kills the NPE, but `?start=` on its own now returns the whole audit trail with nothing to say the filter was dropped. A 500 gets reported; a silently unfiltered result gets believed. Honoring each bound separately is the same two lines:

```suggestion
                        boolean includeAck = (startDateTime == null || !ackTime.isBefore(startDateTime))
                                && (endDateTime == null || !ackTime.isAfter(endDateTime));
```

Same for the resolve branch below, and the history and corrective-action branches have the gap too. `SecuritySliceMockMvcTest` (see `FreezerConfigControllerSecurityTest`) would let you pin it.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
mherman22: You removed "Temperature excursion" because it was wrong, but what's left says nothing: the Details column now just repeats the Action and Freezer columns, so the row doesn't tell you which kind of alert was acknowledged. `alert.getAlertType()` gives you that.

Also in this block: `actionType` is hardcoded `CRITICAL_ALERT_RESOLVED` for every resolution, so a resolved WARNING excursion still shows as "Critical alert resolved".

mherman22 (changes requested):

dharmendra26-wiz:
@mherman22
 Updated! Added tests for the report mapping, switched to independent date bounds across all four filter branches, and included alertType in the audit comments and resolution action type. Thanks for the thorough review!

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
```diff
@@ -151,8 +151,8 @@ public ResponseEntity<List<Map<String, Object>>> getAuditTrail(@RequestParam(req
                     if (alert.getAcknowledgedAt() != null) {
                         OffsetDateTime ackTime = alert.getAcknowledgedAt();
                         // Apply date filter only if dates are provided
-                        boolean includeAck = (startDateTime == null && endDateTime == null)
-                                || (!ackTime.isBefore(startDateTime) && !ackTime.isAfter(endDateTime));
+                        boolean includeAck = (startDateTime == null || !ackTime.isBefore(startDateTime))
+                                && (endDateTime == null || !ackTime.isAfter(endDateTime));
```
mherman22, line 155: No test covers this, so `?start=` alone can regress to unfiltered and nothing would fail. A slice test through `SecuritySliceMockMvcTest` (see `FreezerConfigControllerSecurityTest`) pins it in a few lines.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
```diff
@@ -172,18 +173,19 @@ public ResponseEntity<List<Map<String, Object>>> getAuditTrail(@RequestParam(req
...
+                        boolean includeResolve = (startDateTime == null || !resolveTime.isBefore(startDateTime))
+                                && (endDateTime == null || !resolveTime.isAfter(endDateTime));
 
                         if (includeResolve) {
                             Map<String, Object> resolveEvent = new HashMap<>();
                             resolveEvent.put("id", "RESOLVE-" + alert.getId());
                             resolveEvent.put("freezerId", String.valueOf(fId));
                             resolveEvent.put("freezerName", freezerName);
-                            resolveEvent.put("actionType", "CRITICAL_ALERT_RESOLVED");
+                            resolveEvent.put("actionType", alert.getAlertType() + "_RESOLVED");
```
mherman22: This leaves `CRITICAL_ALERT_RESOLVED` in `formatActionLabel` dead and sends every derived value through the generic prettifier, so the Action column reads "Freezer Temperature Resolved" next to "Alert acknowledged" on the ack row, and the value set now grows with every new `AlertType`. The ack branch keeps a fixed constant and carries the type in the comment, so I would emit `ALERT_RESOLVED` here and let the comment below do that work.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
dharmendra26-wiz: @mherman22,
Good catch. I've updated the date filtering logic so that single-sided date parameters (just ?start= or just ?end=) correctly filter the results instead of bypassing the filter or throwing an NPE. I've also added FreezerAuditTrailControllerSecurityTest extending SecuritySliceMockMvcTest to thoroughly pin this partial-range behavior going forward.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
dharmendra26-wiz: Done. I've updated this to emit the fixed "ALERT_RESOLVED" constant for the actionType so it correctly aligns with the UI prettifier. The specific alert type context is now carried entirely in the comment field, exactly like the acknowledgment branch.

src/test/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailControllerSecurityTest.java
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * Slice test for {@link FreezerAuditTrailController}.
+ *
+ * <p>
+ * Pins the partial-date-range behaviour the reviewer flagged: {@code ?start=}
+ * alone and {@code ?end=} alone must still filter. Pre-fix, history and
+ * corrective-action rows regressed to unfiltered and alert rows NPE'd (→500)
+ * when only one bound was provided. Mirrors
+ * {@code FreezerConfigControllerSecurityTest}'s shape.
+ */
```
mherman22, line 47: The javadoc describes the review rather than the code, and "the behaviour the reviewer flagged" stops meaning anything the moment this merges. Two lines saying what the test pins covers it, and the five `// ---------` banner blocks below can go too, since the method names already say what each group does.

```suggestion
/**
 * Slice test for {@link FreezerAuditTrailController}: a request carrying only
 * {@code start} or only {@code end} must still filter, and must not 500.
 */
```

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
```diff
@@ -172,18 +172,20 @@ public ResponseEntity<List<Map<String, Object>>> getAuditTrail(@RequestParam(req
...
+                        boolean includeResolve = (startDateTime == null || !resolveTime.isBefore(startDateTime))
+                                && (endDateTime == null || !resolveTime.isAfter(endDateTime));
 
                         if (includeResolve) {
                             Map<String, Object> resolveEvent = new HashMap<>();
                             resolveEvent.put("id", "RESOLVE-" + alert.getId());
                             resolveEvent.put("freezerId", String.valueOf(fId));
                             resolveEvent.put("freezerName", freezerName);
-                            resolveEvent.put("actionType", "CRITICAL_ALERT_RESOLVED");
+                            resolveEvent.put("actionType", "ALERT_RESOLVED");
```
mherman22, line 183: `formatActionLabel` in `Reports.jsx` still keys on `CRITICAL_ALERT_RESOLVED`, so that entry is dead now and `ALERT_RESOLVED` falls through to the generic prettifier: the Action column shows "Alert Resolved" next to "Alert acknowledged" on the ack row. Renaming the key at `Reports.jsx:217` to `ALERT_RESOLVED: "Alert resolved"` finishes it.

src/test/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailControllerSecurityTest.java
dharmendra26-wiz: Done. I've updated the Javadoc to focus on the test behavior rather than the review history, and removed all the ASCII banners.

src/main/java/org/openelisglobal/coldstorage/controller/rest/FreezerAuditTrailController.java
dharmendra26-wiz: Done. I updated the key to ALERT_RESOLVED: "Alert resolved" so it properly catches the constant sent by the backend.

mherman22 (approved):

Issue #4340, opened 2026-09-18
Strengthen integration test assertions and eliminate weak guards across test suite


Agaba-derrick:
Current integration and service tests in OpenELIS Global 2 contain weak assertions (e.g., `assertNotNull`, `assertTrue(list.size() > 0)`, `assertNotEquals(0, size)`) and redundant comments/guards. While these pass when objects are instantiated, they fail to verify actual property correctness, count accuracy, or payload integrity, leaving the codebase vulnerable to silent regressions.

## Proposed Improvements
Refactor integration and service test suites to:
1. Replace bare `assertNotNull` guards with concrete property or field assertions (unless guarding fixture setup before invocation).
2. Replace loose size checks (`assertTrue(list.size() > 0)`) with exact count assertions matching test dataset fixtures.
3. Remove redundant comments, unused imports, and unused local variables.
4. Deliver changes incrementally across 4 modular, reviewable PR milestones (m1 to m4).

## Milestone Breakdown
- [x] **Milestone 1 (m1)**: Label Preset & Barcode REST Controller tests
- [x] **Milestone 2 (m2)**: Label Preset & Barcode Service tests
- [x] **Milestone 3 (m3)**: Dictionary & Catalog Service/Controller integration tests
- [x] **Milestone 4 (m4)**: Note & Type of Test Result Core Service tests

## Acceptance Criteria
- [ ] All refactored tests pass with 0 errors and 0 failures.
- [ ] Code formatting complies with Spotless (`mvn spotless:apply`).
- [ ] Weak assertions are replaced with explicit property/size checks.

PR #4349, opened 2026-09-21
test(m2): strengthen weak assertions in label preset service tests

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary

Strengthened integration test assertions across Label Preset service implementation tests:
- Refactored `LabelPresetServiceImplTest`, `OrderEntryLabelRequestServiceAggregationTest`, and `TestLabelConfigServiceImplTest`.
- Replaced redundant `assertNotNull` checks with explicit property assertions and exact expected bean values.
- Cleaned up redundant comments and unneeded helper imports.

## Related Issue


src/test/java/org/openelisglobal/labelpreset/service/LabelPresetServiceImplTest.java
```diff
@@ -102,16 +102,14 @@ public void create_sameNameWithWhitespaceVariant_isRejected() {
...
-            assertNotNull(e.getMessage());
+            assertTrue("Error should mention the colliding name", e.getMessage().contains(TEST_PREFIX + "beta"));
         }
     }
 
     @Test
     public void create_differentName_succeeds() {
         LabelPreset p1 = createPreset(TEST_PREFIX + "gamma");
         LabelPreset p2 = createPreset(TEST_PREFIX + "delta");
-        assertNotNull(p1.getId());
```
mherman22, line 113: why are we removing this?

src/test/java/org/openelisglobal/labelpreset/service/LabelPresetServiceImplTest.java
```diff
@@ -159,7 +157,6 @@ public void update_systemPreset_sameName_succeeds() {
         form.setDefaultPerSample(systemPreset.getDefaultPerSample());
         form.setMaxPerSample(systemPreset.getMaxPerSample());
         LabelPreset updated = labelPresetService.update(systemPreset.getId(), form, SYS_USER);
-        assertNotNull(updated);
```
mherman22, line 162: why are we removing this?

src/test/java/org/openelisglobal/labelpreset/service/LabelPresetServiceImplTest.java
```diff
@@ -188,10 +185,8 @@ public void toggleActive_systemPreset_deactivate_isRejected() {
     @Test
     public void createAndGet_roundtrip() {
         LabelPreset created = createPreset(TEST_PREFIX + "roundtrip");
-        assertNotNull("Created preset should have an id", created.getId());
```
mherman22, line 191: same here

src/test/java/org/openelisglobal/labelpreset/service/LabelPresetServiceImplTest.java
mozzy11: Fine to drop this one: the next line calls `p1.getId().equals(...)`, so a null id still fails the test. The guard only changed the message, not the outcome.

src/test/java/org/openelisglobal/labelpreset/service/LabelPresetServiceImplTest.java
mozzy11: Same here — `assertEquals(systemPreset.getId(), updated.getId())` dereferences `updated` on the next line, so a null return still fails.

src/test/java/org/openelisglobal/labelpreset/service/LabelPresetServiceImplTest.java
mozzy11: This one you were right about, and I've restored it in 6a784e11b. `created.getId()` is passed *into* `labelPresetService.get(...)`, so it guards fixture setup before the call under test — the case #4340 explicitly carves out. Did the same for the `id > 0` unboxing in `saveLabelPreset`.

mozzy11 (approved):
Decision: Approve

Summary:
- The collision assertions are the real win here: `e.getMessage().contains(TEST_PREFIX + "beta")` now pins the message to the offending preset, and I checked it lands on the service's actual text, `"A preset with name '" + normalizedName + "' already exists"` — including the trimmed form for the whitespace-variant case.
- Ran the full backend suite on a tree with all three milestone PRs merged together: `LabelPresetServiceImplTest` 15/15 and `TestLabelConfigServiceImplTest` 7/7 green.
- Pushed two one-line fixes in 6a784e11b rather than send this back for them — see below.
- Answered @mherman22's three questions inline: two of the removals are fine (the next line dereferences the same value), the third was a fair catch.

Optional improvements:
- `assertPerSamplePreset_perSamplePreset_doesNotThrow` now has no assertion at all. That's defensible — the test fails if the call throws — but a reader scanning for an assert will stop on it.

What I restored and why: `createAndGet_roundtrip` passes `created.getId()` into `labelPresetService.get(...)`, and `saveLabelPreset` unboxes the DAO's `Integer` for `id > 0`. In both, a null arrives as a bare NPE instead of a named failure, and #4340 keeps guards on fixture setup that precedes the invocation. The other removals don't have that property, so I left them.

Verified by execution: full backend suite on the combined merge tree (6563 tests); all 7 classes the three PRs touch are green. Remaining local failures are an uninitialized `tools/openelis-analyzer-bridge` submodule in my worktree and the known `DisplayListControllerStageFlagsTest` ordering artifact — neither touches this PR. No live UI/API test: the diff is `src/test/java` only, so there is no user-visible behaviour to walk.

PR #4351, opened 2026-09-21
feat(test): m4 strengthen core service test assertions

Agaba-derrick:
# Pull Requests Requirements

      Issue number if applicable.
      documentation.
      project.

## Summary
Strengthened integration test assertions across core services:
- Refactored `NoteServiceTest` and `TypeOfTestResultServiceTest`.
- Replaced loose list checks (`assertTrue(list.size() > 0)`) in `TypeOfTestResultServiceTest` with exact size assertions (`assertEquals(7, list.size())`).
- Replaced loose list checks in `NoteServiceTest` with direct index access and exact size checks.
- Removed unused local variables (`originalId`, `originalHl7Value`) and unneeded imports.
## Related Issue

## Other


src/test/java/org/openelisglobal/note/NoteServiceTest.java
```diff
@@ -41,8 +41,6 @@ public void setup() throws Exception {
     @Test
     public void deleteNote_shouldDeleteNote() throws Exception {
         Note savedNote = noteService.get("1");
-        assertNotNull("Note should exist in dataset", savedNote);
```
mherman22, line 44: why are we removing this?

mozzy11: Good catch — restored in d5043eea0. `savedNote` goes straight into `noteService.delete(...)`, so a missing fixture row would have surfaced as an NPE inside the service rather than as a named failure here.

mozzy11 (approved):
Decision: Approve

Summary:
- Counts check out against `notes.xml`, which seeds exactly one INTERNAL note (subject "Subject 1") and one EXTERNAL: all three `assertEquals(1, notes.size())` are correct, and so is `assertEquals(7, ...)` against the 7 rows in `type-of-testresult.xml`.
- Collapsing `getReferenceTableIdForNoteBinding` and `getTableReferenceId` from null-tolerant branches to a direct index read is the best change in the set — those two could previously pass on an empty list.
- `NoteServiceTest` 22/22 and `TypeOfTestResultServiceTest` 9/9 green on a tree with all three milestone PRs merged together.
- Pushed two small fixes in d5043eea0 rather than send this back: restored the guard before `noteService.delete(savedNote)`, and dropped `testDataInDataBase`.
- Answered @mherman22's question inline — they were right on that one.

On the guard: `savedNote` goes straight into the call under test, so a missing fixture row would have surfaced as an NPE inside the service instead of a named failure, which is the case #4340 carves out. On the duplicate: once `testDataInDataBase` asserted the exact count it was byte-identical to `getAll_shouldReturnAllTypeOfTestResults`, so I kept the better-named one.

Verified by execution: full backend suite on the combined merge tree (6563 tests); all 7 classes the three PRs touch are green. Remaining local failures are an uninitialized `tools/openelis-analyzer-bridge` submodule in my worktree and the known `DisplayListControllerStageFlagsTest` ordering artifact — neither related to this PR. No live UI/API test: the diff is `src/test/java` only, so there is no user-visible behaviour to walk.

PR #4355, opened 2026-09-21
feat(OGC-1226): let ordinary results clear for bulk release, and say why when they cannot

mozzy11:
Implements OGC-1226 (epic OGC-817; FRS `designs/results-validation/validation-clearance-rule.md`, D-056/D-057).

**What changed**
- Ordinary in-range results now land in the Clear lane and "Release all clear (N)" works. A row is held only for a *recorded* QC failure; a missing QC evaluation (every patient result) no longer blocks it. All other clearance terms are unchanged.
- The server decides the lane for every row it serves (`clear` on each row); the page reads it and no longer computes a lane of its own. The bulk release still re-derives the lane from its own load.
- Automated validation obeys the same rule: with "validate all results" off, a result is finalized at entry only when the Clear lane would have held it. Abnormal, unranged and re-saved results now wait for a validator instead of being released unattended.
- When the bulk button is disabled the queue says why next to it, with counts: rows carrying signals (most common named), tests with no reference value in the catalogue, an empty queue, or bulk release switched off. The button stays visible.
- The confirm dialog carries one line on what clearance checked and did not; the review panel shows a QC verdict only when one exists.

**Why**
- Clearance read `Result.qcEvaluation`, which is only written for QC sample items, so the Clear lane was unreachable by construction and the greyed button gave no reason.

**How it was tested**
- Backend: `ValidationSignalsTest` (39, incl. the entry-time rule), `AccessionValidationBulkReleaseTest` (7, new QC-failed accession), new `ResultUtilAutoValidationTest` (4); full `mvn test` green in an isolated copy; with the old rule restored, 5 of these fail.
- Frontend: validation component suites 83 green (lane from the server verdict, explanation reasons, scope line, QC slot).
- Live on dev compose through the real UI (Playwright spec `ogc-1226-validation-clearance.spec.ts`): in-range result in the Clear lane, released in one action with the confirmation and the empty-queue explanation; abnormal result held with the signals explanation, then released per row. Queue API read back with `clear` per row. FR-7 exercised with "validate all results" off: in-range finalized at entry, abnormal held (setting restored).

**Notes**
- FR-14(a) copy reads "{count} results in this queue carry…" instead of the FRS's "All {count}…", so it stays true when the no-reference reason applies as well.
- Not exercised live: the no-reference reason and the bulk-disabled reason (component tests), and a QC-failed row (integration test; it needs a QC-profile sample).
- The FRS asks for a Feature Doc under OGC-817 and a re-capture of manual page 1661763585; both are follow-ups, not in this PR.
- Full local `mvn test`: 6564 run, 4 errors, all in `DisplayListControllerStageFlagsTest` (new on develop from #4298). It reads the banner-text localization through `Optional.get()`, and 77 fixtures truncate `localization`; the pre-existing `AccessionValidationBulkReleaseTest` alone reproduces it when ordered first. Pre-existing order dependency, passes in CI's order; worth its own small fix.


mozzy11 (commented):

Decision: Approve

Summary:
- Clearance logic centralized in `ValidationSignals.isClear()` and `isClearAtEntry()`, correctly treating QC_UNKNOWN as "no input" rather than blocking
- Server-authoritative lane assignment via new `clear` field on `AnalysisItem`, frontend reads it instead of deriving its own
- Bulk unavailable reasons (FR-14a-d) display correctly: signals with dominant naming, no-reference tests, empty queue, bulk disabled
- QC verdict hidden in review panel for patient results (FR-10)
- Backend 50 tests pass (4 + 39 + 7), frontend 91 tests pass, no console errors in live test

Verification:
- Built and tested in isolated copy; backend `ValidationSignalsTest`, `AccessionValidationBulkReleaseTest`, `ResultUtilAutoValidationTest` all green (50 tests)
- Frontend validation suite 91 tests green
- Live tested on dev stack via Playwright MCP: abnormal result (25 vs 4-10 range) correctly held in Needs-review lane with explanation "1 results in this queue carry something to check before release. Most common: Abnormal."; QC section shows no status tag for patient sample; bulk button disabled with count (0); empty queue shows "Nothing is waiting for validation"
- Functions inventoried: 7 exercised, 7 passed; Clear-lane bulk release not exercised live (no in-range test data available, covered by automated tests)

Optional improvements:
- Consider memoizing `bulkUnavailableReasons()` in Validation.jsx if queue sizes exceed 500 rows
- The `DisplayListControllerStageFlagsTest` failures noted in PR description are pre-existing order-dependency issues, not introduced by this PR

mozzy11:
Thanks for the walk-through. On the two optional points:

- Memoizing `bulkUnavailableReasons()`: left as is. It is one linear pass over rows the page has already triaged in the same render, and the queue's other per-render derivations (`triageRows`, `countByFilter`, `clearRows`) take the same approach without memoization. Wrapping just this one in `useMemo` would buy nothing while `triaged` itself is rebuilt every render; if queue sizes ever make the render cost visible, the right change is to memoize the triage once and derive everything from it.
- `DisplayListControllerStageFlagsTest`: agreed, pre-existing order dependency on develop (its banner-text localization lookup versus fixtures that truncate `localization`). Tracked separately, not in this PR.

PR #4378, opened 2026-09-22
fix(OGC-1231): Modify User shows the right lab unit and refuses All Lab Units next to scoped grants

mozzy11:
Fixes OGC-1231.

**What changed**
- Modify User: each Lab Unit Roles row now shows the lab unit it is actually stored against (the dropdown was uncontrolled and stranded on "All Lab Units").
- "All Lab Units" is now explicit and exclusive: choosing it while scoped rows exist opens a confirm dialog naming the grants that will be replaced; Add New Permission never adds it by default and is disabled while it is present; the API refuses a submission that combines it with specific lab units (HTTP 400, nothing written) and the screen shows that error instead of a success toast.
- Login names such as `qa_recept` no longer render as invalid: the client rule now mirrors the seeded server charset, and an invalid name shows why.
- Saving an existing user says "User information updated successfully." (new user: "added"), spelled correctly.
- Found while live-testing: the save endpoint answers 200 for a refused save too (validation, or a duplicate first+last name), and the screen showed the success toast. It now shows the error toast unless the server answers with the redirect that means "written".

**Why**
- The write path never dropped anything: the GET normalizes a user with `AllLabUnits` to that row only, so the scoped grant disappeared from the screen and was deleted by the next save, with a green toast each time. Casey ruled the two are exclusive, so the combination is refused up front rather than reconciled.

**How it was tested**
- New vitest tests (11, each fails with the fix reverted; full frontend suite 2342 green) and a MockMvc test for the 400 plus two accepted shapes; live on dev compose: a Hematology-scoped user and an All-Lab-Units user driven through the dropdown, Add New Permission, the confirm dialog (confirm and cancel), Remove Permission, Save on both the modify and add screens, a refused duplicate, and the login-name field, with the combined map posted through the API and the DB rows read back after each call; new Playwright spec `admin-user-lab-unit-roles.spec.ts` passes against the stack.

**Notes**
- The Cypress admin spec `userManagement.cy.js` relied on Add New Permission defaulting to All Lab Units and built a user holding All Lab Units plus ten scoped units; it now selects All Lab Units explicitly and gives that user scoped units only, which is all its later lab-unit filter searches need.
- The GET still returns only the `AllLabUnits` row for legacy users that hold both shapes (authorization already treats `AllLabUnits` as superseding), so such users stay saveable and their dead scoped rows drop on the next save.
- The client login-name rule mirrors the default `userNameCharset`; a site that widens that charset would still see the stricter client rule. Exposing it through configuration properties is a follow-up.
- Observed but left alone: the POST does not return early on bean-validation errors, so an invalid login name reaches persist and 500s. Fixing that would also block saving users whose stored password-expiry date is already past, which needs its own ticket.


mozzy11 (commented):

Decision: Approve

Summary:
- Dropdown now correctly shows the stored lab unit value on Modify User (controlled component fix)
- All Lab Units exclusivity rule enforced: info text, disabled Add button, confirmation modal, and backend 400 rejection all work correctly
- Login name regex widened to accept underscores (test_lab_unit created and updated successfully)
- New Playwright spec and backend MockMvc tests provide solid coverage; all CI suites green

Verification: Built PR head in worktree, deployed to dev stack, tested 9 functions through UI (Playwright MCP) and API. All passed: create user with All Lab Units + Reception, modify user shows correct dropdown value, confirmation modal on switching to All Lab Units, API rejects combined All Lab Units + specific unit with 400 `labUnitRoles.allLabUnitsExclusive`, underscore in login name accepted. Backend suite (3 tests) and frontend suite (2342 tests) green.

PR #4380, opened 2026-09-22
feat(results): accept Unicode superscript exponents in numeric results

HerbertYiga:
Let users type values like "3²" or "3.5⁻³" in the result value field. A shared converter rewrites the superscript run as scientific 'e' notation (3² -> 3e2, 3.5⁻³ -> 3.5e-3) so both the frontend validators and the backend StringUtil.isNumeric / getActualNumericValue paths parse them as real numbers.

Frontend: adds superscriptNumeric.js (+ Jest tests) and wires it into SearchResultForm's validation, switching the result input from type="number" (which strips non-ASCII characters) to type="text" so the superscript characters survive the DOM. When conversion happens, the normalized value is written back via validation.newValue.

Backend: adds StringUtil.convertSuperscriptToScientific mirroring the JS logic, applies it inside isNumeric() and getActualNumericValue() so any downstream consumer that parses stored values keeps working. Plain numeric input is passed through unchanged.

Tests cover mantissa+integer/decimal, multi-digit exponents, negative exponents, and pass-through of plain and non-numeric input on both sides.


mozzy11 (commented):
Decision: Comment

Summary:
- Deployed the branch on the dev stack and walked it through both Results Entry screens, the Validation queue (display, Modify, Release), the patient report PDF and the per-analysis and logbook APIs. The `type="text"` switch and the idea of one converter shared by client and server both hold up.
- The convention does not: `3²` reads as nine, not three hundred, and storing it raw broke everything downstream. Live on the head commit: the Validation queue and the report printed `3².00`, the row lost its flag (`resultFlag` null, so a critical value would release unflagged), the FHIR transform logged `BigDecimal("NaN")` on release, the legacy screen showed the saved value red on reload, and the unified screen (still a number input) silently kept only `3`.
- I pushed 290ae0d77 on top: both screens and `ResultSaveService` now read scientific notation as typed (`1.5e5`, `1.5E+05`, `1.5 x 10^5`, `1.5×10⁵`, `10⁻³`) and store canonical `1.5e5`; a bare superscript is refused as not a number, on the client and by `ResultsValidation` (whose catch was unreachable before, so `abc` was being stored). Display pads the mantissa (`1.5e5` at two places prints `1.50e5`) instead of appending `.00` to the string.
- With that commit every row passes live: legacy and unified entry, 400 for `abc`/`3²`, canonical DB values, queue flags NORMAL/ABNORMAL, Modify (`1.6 x 10^1` → `1.6e1`), Release, report `2e5 (5.3)log E`, no FHIR errors.
- Tests: StringUtilTest 33, new ResultsValidationNumericFormatTest 4, ResultServiceTest +3 (110 backend tests green in an isolated copy); scientificNotation.test.js 10 and the unified cell tests updated; full frontend suite 316 files green; spotless clean in the CI container.

Optional improvements:
- Two calls I made on your behalf, worth a second look: a whole-number test (0 places) leaves a mantissa alone (`1.5e5` stays `1.5e5` rather than rounding to `2e5`), and the report prints the canonical `2e5` rather than a `× 10` form.
- Pre-existing, not introduced here: `POST /rest/LogbookResults` collects validation errors and still saves the row (the legacy Save now refuses a non-numeric field client-side, but the endpoint stays permissive), and the per-analysis 400 body prints `{0} is not a number.` with the placeholder unresolved. Both deserve their own ticket.
- The PR title and description still describe the superscript convention; please update them to the scientific-notation behaviour before merge.

Verified by execution: 14 functions walked on 75a8aef63 and 17 on 290ae0d77, UI via Playwright plus API, 17 of 17 passing after the fix. Not exercised: analyzer import, and the digit-stripping parsers in ExportTrendsByDate and ForCIDashboard (untouched).

mozzy11:
Pushed f22d93cb7 on top of 290ae0d77: formatting only. CI's spotless measures lines in characters, so the two lines I had wrapped by hand were re-joined and one Javadoc reflowed; no code change.

mozzy11:
One more small commit, 275c28470: the E2E shard failed on `ogc-1121-critical-result-flag.spec.ts` with the abnormal row's background read as rgb(255, 255, 161) instead of 160. Carbon animates the field's background-color over 70 ms and the spec read the computed style right after the value check, so the assertion now polls until the colour settles. Test-only change.

mozzy11:
develop moved under the branch (#4241 also hardened `ogc-1121-critical-result-flag.spec.ts`, the same colour flake), which left the PR in conflict and stopped GitHub from starting the pull_request checks. Merged develop into the branch (c1d52c52d), taking develop's version of that spec; 275c28470 and the empty re-trigger commit are superseded and will disappear on squash.

mozzy11:
Pushed fafd1c5e6, which changes the storage model to the one asked for on the thread: the value is kept in the notation the technologist typed, and every screen renders it that way.

`result.value` now holds the written string. `Result.getValue()` returns the canonical e-notation form so the 298 existing callers, and any future one, still hand a plain number to `Double.parseDouble`, to a SQL cast and to the FHIR interface; `getEnteredValue()` returns what was written and feeds the display formatters. The mapping gains `access="field"` so Hibernate persists the entered form rather than the normalized one.

Walked live on DEV01260000000000127 through both Results Entry screens, the Validation queue, Modify, Release and the patient report PDF:

| Typed | Stored | Rendered everywhere |
|---|---|---|
| `7.5 x 10^0` (2 dp) | `7.5 x 10^0` | `7.50 x 10^0` |
| `1.5×10¹` (1 dp) | `1.5×10¹` | `1.5×10¹` |
| `2.5E+02` (0 dp) | `2.5E+02` | `2.5E+02` |
| `3.567 x 10^3` (2 dp) | `3.57 x 10^3` | `3.57 x 10^3` |
| `1.5e5` | `1.5e5` | `1.5e5` |

The only rewriting is the reporting-precision rule plain decimals already follow. Flags still come out right against the ranges (the four in-range rows landed in the Clear lane, viral load Abnormal against 1-300), the report printed `1.5e5 (5.18)log` with the above-normal marker, and a bare `3²` is still refused by the screen and by the API.

One bug fixed along the way that is worth its own look: `ComplianceDashboardQueryServiceImpl` casts `r.value` directly in SQL behind a guard that matched plain decimals only, so a result in any scientific notation dropped silently out of compliance evaluation. The query now rewrites the written forms before the cast; I verified the expression in Postgres against 15 values, including that `abc` and `3²` are skipped rather than misread.

Full backend suite and the 316-file frontend suite are green; spotless clean in the CI container.

mozzy11:
Pushed 59158026d. Driving the rule engines and the range tiers end to end found a real regression in what I had pushed earlier, and it is now fixed.

**The bug.** Storing the value as written left seven places parsing the *displayed* string. E-notation happened to parse, the written forms did not, so a result entered as `5×10¹` came back from Results Entry **carrying no flag at all** — a critical result could have reached the Validation queue unmarked. Fixed inside the shared functions so every present and future caller is safe: the four-tier flag, the entry screen's valid/normal checks, the patient results interpretation, the trend chart, the timeline's plottable test, the unified interpretation matcher, the viral-load log suffix and the log column. The critical alert now names the result the way it was written, while the rules still run on the number.

**Walked live on the dev stack, every item observed in the browser:**

| Area | Result |
|---|---|
| Range tiers in written notation | `5×10¹` Normal, `8.5×10¹` Abnormal (yellow), `9.6×10¹` Critical (red + Critical chip + server flag), `1.5×10²` Invalid |
| Reflex engine | rule `> 90` fired on `9.6×10¹`, adding the reflex test to the order, twice |
| Alert engine | alert raised and rendered on the Alerts screen reading `result 9.6×10¹` |
| Compliance Dashboard | environmental order, `1.5×10³` Total Coliform scored **FAIL** against the 1000 maximum, 1 exceedance, result column shows `1.5×10³` |
| Patient trend chart | `3.57 x 10^3` plots as **3570**, not 3.57 |
| Patient timeline | every stored notation rendered as written |

Configured the reflex rule, the alert rule, the sampling site, the environmental order and the compliance standard through the UI to get there.

**Not verified:** calculated values. The engine never executed in this environment, for a plain `75` exactly as for `7.5×10¹`, including on an order created after the rule, so it is not notation-related. The code path reads the normalized value and evaluates it with Nashorn, which parses e-notation natively.

201 backend tests across every touched class and the full 316-file frontend suite are green; spotless clean in the CI container.

mozzy11:
Correction to my note above, and one more commit.

**Calculated values: verified. My "not verified" was wrong.** The engine had run every time; it writes into an analysis it creates itself rather than into the Hemoglobin ordered by hand, and I was reading the wrong row. Fresh orders, one value each, observed in the browser and in the database:

| Entered | Reflex `> 90` | Calculated `+ 1` | Critical `> 90` |
|---|---|---|---|
| `8.5×10¹` | not fired | `86.0` | no alert, row flagged Abnormal |
| `9.5×10¹` | fired, reflex analysis added | `96.0` | alert raised, message reads `9.5×10¹` |

Editing the first order to `87` recalculated to `88.0`; the stored value stays exactly as typed throughout.

**dc314dfd9, alert rules authored in notation.** The Specific Value trigger is a free-text field, so a rule typed as `8.7×10¹` never fired: the trigger side was not normalized and the parse failed silently. Both sides are now read as numbers. Live: rule `8.7×10¹` fired on a result of `87` and stayed quiet on `86`; the new unit test fails on the previous line and passes on this one. Reflex conditions and calculation operands are number-typed inputs, so they cannot be written this way and are unchanged.

Rule engines: 3 inventoried, 3 driven live through the UI, 3 passed. The only console errors were the pre-existing missing `esig.button.saveModification` label from #3946.

mozzy11 (approved):
Approving. Walked the full flow live on the dev stack: entry in every notation form, validation queue, patient report, patient results, compliance dashboard, and the reflex, calculation and alert engines with a firing and a non-firing value each. CI green on dc314dfd9.

PR #4381, opened 2026-09-22
fix(ci): isolate downstream E2E concurrency by build

pmanko:
PR-triggered `03 - E2E` builds all expose `develop` and the same base SHA to the downstream `workflow_run` event. The existing concurrency key therefore groups unrelated PR executors together, so each newly completed build cancels another PR's downstream run before it can publish `03 Checkpoint - E2E`.

Use the triggering build run ID as the concurrency key. Different PR builds can now reach the executor independently, while rerunning the same source build still replaces its earlier downstream attempt. A workflow-contract test prevents the key from returning to the shared base branch or SHA.

Validation:

- `/tmp/oe2-ci-fix-venv/bin/python .github/scripts/test_publication_workflow.py` (4 passed)
- `node --test .github/scripts/publish-checkpoints.test.cjs` (9 passed)
- `git diff --check`


pmanko:
Closing this diagnostic PR because its root-cause premise was not established. OE2's /fix-ci guidance explicitly notes that workflow_run API records inherit develop/base SHA while testing the PR head through e2e-build-context. The reporting stack will be re-evaluated through gh pr checks and the linked authoritative E2E artifacts.

